PR #7, opened 2020-11-18
669: Prep repo

NickPhura:
Ticket: https://quartech.atlassian.net/browse/BHBC-669

# /
  - Pulled a bunch of code over from InvasivesBC, partially to kick-start some of the development, but also just to get some of the folder structure and generic files in place. Will of course still require some additional updates/cleanup/configuration, etc.
  - I've replaced all variations of `invasivesbc` with `biohubbc`
## /api
  - Pulled in a bunch of files from InvasivesBC
  - Pared down the api doc and the routes, etc, so we have more of a clean slate to work from
## /app
  - Pulled in a bunch of files from InvasivesBC
  - Removed all of the Ionic elements (so it should be a pure React app)
  - Pared down some of the more Invasives specific pieces so we have more of a clean slate to work from
## /app-ionic
  - Pulled in a bunch of files from InvasivesBC
  - Did not pare this down much, as more parts of this are likely going to stay the same/similar (at least the underlying logic) to what BioHub wants.
    - Seemed easier to leave it and pare it down as we go, rather than pre-emptively.
## /database
  - Pulled in a bunch of files from InvasivesBC
  - Cleared out all migrations/seeding, but left the infrastructure there
## /app-common-library
  - Added a new folder for common react components/utils.ts
  - Added this folder to the `include` array in `/app/tsconfig.json` and `app-ionic/tsconfig.json`


NickPhura:
I don't really expect anyone to properly review this, but wanted you to be aware of what I had done.

rstens (commented):
I would keep the workflows out of this for now (maybe rename them to *.bkp).
Because they get kicked off. We also are in a different OS namespace, so careful with the config file if you copy that one.

NickPhura:
> 
> 
> I would keep the workflows out of this for now (maybe rename them to *.bkp).
> Because they get kicked off. We also are in a different OS namespace, so careful with the config file if you copy that one.

Good call, I went into the repo settings and disabled all actions for now . Did the same for the common-codes repo as well.

PR #14, opened 2020-12-21
BHBC-693, BHBC-694, BHBC-695

anissa-agahchen:


database/src/migrations/20201216104841_init.ts
```diff
@@ -0,0 +1,42 @@
+import Knex from 'knex';
+
+const DB_SCHEMA = process.env.DB_SCHEMA || 'biohubbc';
```
NickPhura, line 3: We probably don't need the DB_SCHEMA variable anymore. 

I know we added it to the Invasives migrations, but in the end it was never used (since the value never changes). And now that we have `.sql` files where the schema is static, having it be dynamic in the migrations is irrelevant.

We could clean this up later after this is merged, as some of the migrations I've already added (as part of that template work I did) also still use DB_SCHEMA, and could use updating. There are a few places in the code as well that reference it.  May as well do it all together.

NickPhura: I created a ticket for it: https://quartech.atlassian.net/browse/BHBC-733

NickPhura:
@anissa-agahchen @cgarrettjones Whats the state of this PR and the code changes? Can they be merged?

anissa-agahchen:
The lack of autonomous transactions is the blocker to this work ... I recommend we don't merge the code as it may be some time before we resolve this issue.  @cgarrettjones, your call.

cgarrettjones:
@NickPhura @anissa-agahchen i've documented what the new approach will be in the tickets which entails our going with as needed native database logging facilities. thus i think this work can be rolled back.

PR #39, opened 2021-01-29
BHBC-db-context: Add db user context pre-step and update db env vars

NickPhura:
# Overview

- Update the old `getDBConnection` function in `db.ts` to now return an object that wraps the raw connection client.
  - This object has handful of obvious functions for opening a connection, performing queries, and committing/rollingback/releasing the connection
  - When used, it will automatically set the user context for you, so you dont have to worry about it.
    - This will make the background functions/triggers set up by Charlie happy
- Updated the DB related env vars to be more obvious
  - This came up because the env var naming was based around 1 user, but now that we have 2 users, it was getting confusing to track which user was being used when (in the github workflows and docker compose, etc)
    - `DB_USER/DB_USER_PASS` come from the `database-user/database-user-password` secrets, and represent the `biohubbc-user` user
    - DB_USER_API/DB_USER_API_PASS come from the `database-user-api/database-user-api-password` secrets, and represent the the `biohub_api` user
    - The only exception to the above is in the db setup workflow
      - In this case, the `DB_USER/DB_USER_PASS` vars are set from the `database-user-root/database-user-root-password` secrets, as that is the only user that can currently run the migrations/seeding.  If the `biohubbc-user` gets those permissions later on, we can easily change this back to it, and we'd only have the 2 above users.
- Added a new seed to get our dev team IDIRs added to the system_user table.
  - As per the comment in that file, that seed will become unneeded once we have the capability in-app to add users.  This is just a stop-gap so that no one testing in dev is blocked in making api calls.
- Added a minor tweak to the database setup workflow, to toggle between 2 dockerfiles based on if the deployment is static or not.  If it is static it only runs the migrations, if it is non-static it runs the migrations+seeding.

- [ ] Version change

## Checklist

### General

### Code

### Documentation

### Tests

## How Has This Been Tested?

## Screenshots


NickPhura:
I notice that the seeding isnt running (because the command in the dockerfile is only running the migrations).
I made some updates in invasives to toggle between migrations+seeding (for local and pr builds) and just migration (for dev/test/prod) builds. So I can bring that in as well.

rstens (dismissed):

NickPhura:
Closing and re-opening to kick of fresh github workflows

PR #43, opened 2021-02-01
DNM: Use `database-user` as root system user

NickPhura:
Testing if I can still use the `biohubbc-user` as the base system user.  Would mean we dont need to do a clean wipe of the dev database or change any migrations


NickPhura:
Was only for testing purposes

PR #44, opened 2021-02-02
BHBC-744: Create Project Workflow

NickPhura:
# Overview

## Links to jira tickets

https://quartech.atlassian.net/browse/BHBC-744

- Add YesNoDialog
- Add ErrorDialog
- Create project stepper
  - Supports validation on `next` click
  - Remembers data if you move forward or back in the steppers
- Saving project details portion of create project record.
- Add support to populate code values in dropdowns
  - Currently used to populate the management action type dropdown 
- Add custom validation support
  - Currently used to validate the date ranges
- Tweaks to db.ts
- Update various doc
- Add some tests
  - Will need to add more, but want to get this out the door before it gets even more massive
- Update package dependency versions to resolve linter error (documented in readme) and to resolve npm version warnings
- Rename postman export files to be whatever the default export filename is.  So its easy to overwrite the old ones when exporting the new postman files.

- [ ] Version change

## Checklist

### General

- [x] My code follows the style guidelines of this project
- [x] New and existing unit tests pass locally with my changes

### Code

- [x] My code uses types/interfaces to help describe values/parameters/etc and inform future developers

### Documentation
- [x] I have added/updated README's and related documentation

### Tests
- [x] I have added/updated unit tests for any new code I've added/updated
- [x] I have added/updated Postman tests to account for any new/updated code

## How Has This Been Tested?

Locally + some unit tests.

## Screenshots


NickPhura:
(Stealing this from when I first opened this PR last week)

Hmm, the sonar cloud is failing because of duplicated lines, but they seem to be primarily false positives.

The complaint about the root-api-doc.ts is technically valid, but I don't know if extracting pieces into re-usable objects will really benefit us in any way. Possibly a candidate file to ignore? I can also make the changes to appease it of course if needed, it wouldn't be hard.

Some of the other complains are due to similar code existing in multiple projects (ie: interfaces in the api matching interfaces in the app). This can't really be helped as both places need them defined. Not sure how we could prevent cross project checking though...?

And a few others are just due to code in fairly empty pages matching each other (ie: in the view project page and edit project page) as those aren't flushed out yet.

NickPhura:
Looks like the migrations + seeding ran successfully, but because I pushed another commit (which was squashed down), it tried to run the seeding again and of course failed because of duplicate key constraints.  Might have to think of a better way to handle seeding in PRs, or make sure the seeding un-does itself on subsequent runs, which we did for invasives.  Anyways, should be easy enough to fix, but just a heads for when you notice the setup is marked as failed.

I opened a separate little PR to address this issue: https://github.com/bcgov/biohubbc/pull/45

rstens (approved):

PR #54, opened 2021-02-03
Fix useBioHubApi when running in docker

NickPhura:
# Overview

- Small fix to prevent an error when running the app locally in docker
  - Before, it would attempt to fetch the config from `/config/app` which would only exist if running in openshift (via `app/server/index.js`).
  - Now, it checks to see if the docker-compose env vars are set, and if so uses those instead.

- [ ] Version change

## Checklist

### General

### Code

### Documentation

### Tests

## How Has This Been Tested?

## Screenshots

### DIFF
--- a/app/src/hooks/useBioHubApi.ts
+++ b/app/src/hooks/useBioHubApi.ts
@@ -7,23 +7,30 @@ import { useMemo } from 'react';
 const API_HOST = process.env.REACT_APP_API_HOST;
 const API_PORT = process.env.REACT_APP_API_PORT;
 
-const API_URL =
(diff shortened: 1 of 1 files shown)


anissa-agahchen:
This PR fixes the issue useBioHubApi issue, but the create project workflow fails at the end.  It gives a 500 error.  Might be coming from the "scientific_collection_permit_number" column.

NickPhura:
> 
> 
> This PR fixes the issue useBioHubApi issue, but the create project workflow fails at the end. It gives a 500 error. Might be coming from the "scientific_collection_permit_number" column.

@anissa-agahchen Yeah most likely.  Its the one field that is not required by the app or api, but is required by the database.  So it throws an error if you dont populate it.  Its a known issue of some of the project work in general, so not a blocker for this PR.

anissa-agahchen (approved):

PR #58, opened 2021-02-05
Add config context

NickPhura:
# Overview

- Add ConfigContext, to provide environment variables to the app in a more centralized/streamlined way
  - Checks the NODE_ENV to determine if vars should be pulled from `process.env.*` or from `/config` (via `app/server/index.js`)
- Removed the config related changes from `useBioHubApi.ts`
  - Removed a no longer needed env var from the env.docker and docker-compose files
- Removed duplicate app/package.json script `build-prod` as the regular `build` already does a production build
- Remove `keycloak.json` file as it is no longer used

- [ ] Version change

## Checklist

### General

### Code

### Documentation

### Tests

## How Has This Been Tested?

## Screenshots


app/src/App.tsx
```diff
@@ -4,37 +4,50 @@ import type {} from '@material-ui/lab/themeAugmentation'; // this allows `@mater
...
-        </KeycloakProvider>
+        <ConfigContextProvider>
+          <ConfigContext.Consumer>
+            {(config) => {
+              if (!config) {
+                return <CircularProgress />;
+              }
+
+              //@ts-ignore
+              const keycloak: KeycloakInstance = new Keycloak(config.KEYCLOAK_CONFIG);
```
NickPhura, line 26: This block is essentially the only new part, everything below is just indented now that its being wrapped.

Basically, the app waits for the config to be loaded before rendering, as certain config items are mandatory.

app/src/hooks/useBioHubApi.ts
```diff
@@ -1,37 +1,10 @@
 import { useKeycloak } from '@react-keycloak/web';
```
NickPhura, line 1: Everything removed from this file was only recently added, to solve the same problem this PR is solving, but in a worse way.  All of the main logic here has more-or-less moved to the configContext.tsx.

NickPhura:
Already addressed the SonarCloud error, which was a false positive/not a real issue.

rstens (approved):

PR #61, opened 2021-02-10
Bhbc 528

anissa-agahchen:
# Overview

## Links to jira tickets

- BHBC-528 {https://quartech.atlassian.net/browse/BHBC-528}

- Get a list of files on s3, with projectId as the prefix
- Download a single file, and see its metadata
-  Upload a single file, or multiple files with a given projectId - the projectId becomes a folder in S3

- [ ] Version change

## Checklist

### General

- [x] My code follows the style guidelines of this project
- [x] New and existing unit tests pass locally with my changes

### Code

Caveat: some hard-coded projectIds that will be associated to a project in near future work
- [x] My code uses types/interfaces to help describe values/parameters/etc and inform future developers

### Documentation

### Tests
- [x] I have added/updated unit tests for any new code I've added/updated
- [x] I have added/updated Postman tests to account for any new/updated code

## How Has This Been Tested?

- unit tests in both app and api
- postman tests
- you can verify that the files uploaded through an s3 client

## Screenshots


api/src/paths/projects/{projectId}/artifacts/list.ts
```diff
@@ -0,0 +1,96 @@
...
+    },
+    default: {
+      $ref: '#/components/responses/default'
+    }
+  }
+};
+
+function getMediaList(): RequestHandler {
+  return async (req, res) => {
+    defaultLog.debug('started... req.params.projectId=' + req.params.projectId);
```
NickPhura: For the defaultLog calls, can you update this (and any others) to match the existing pattern:
```Typescript
defaultLog.debug({label: 'getMediaList', message: 'params', req.params);
```

Specifically, the `label` being set to the function name (or some other relevant label if there is no function).  The `message` being set to your log message.
You can keep the projectId as part of the message if you want, but you can also just throw the entire `req.params` object at the end, and it will get printed. Any number of objects can be added actually, and they will all get printed out.

api/src/utils/file-utils.ts
```diff
@@ -25,12 +28,31 @@ const S3 = new AWS.S3({
...
+  if (!prefix) {
+    return {};
+  }
+
+  const params = {
+    Bucket: OBJECT_STORE_BUCKET_NAME,
+    Prefix: prefix
+  };
+
+  defaultLog.debug('params:', params);
```
NickPhura: Extra log, can remove

api/src/paths/projects/{projectId}/artifacts/list.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (!req.params.projectId) {
+      throw {
+        status: 400,
+        message: 'Missing required path param `projectId`'
+      };
+    }
+
+    const { Contents } = await getFileListFromS3(req.params.projectId + '/');
+
+    defaultLog.debug('Contents:', Contents);
```
NickPhura: Extra log, can remove

api/src/utils/file-utils.ts
```diff
@@ -25,12 +28,31 @@ const S3 = new AWS.S3({
...
   return S3.getObject({ Bucket: OBJECT_STORE_BUCKET_NAME, Key: key }).promise();
 }
 
+/**
+ * Get file list from S3, based on name prefix
+ * @param prefix
+ */
+export async function getFileListFromS3(prefix: string): Promise<ListObjectsOutput> {
+  if (!prefix) {
+    return {};
```
NickPhura: I would return null here instead, and then just change the function response type to `Promise<ListObjectsOutput | null>`

api/src/paths/projects/{projectId}/artifacts/list.ts
```diff
@@ -0,0 +1,96 @@
...
+    defaultLog.debug('started... req.params.projectId=' + req.params.projectId);
+
+    if (!req.params.projectId) {
+      throw {
+        status: 400,
+        message: 'Missing required path param `projectId`'
+      };
+    }
+
+    const { Contents } = await getFileListFromS3(req.params.projectId + '/');
```
NickPhura: Related to this comment: https://github.com/bcgov/biohubbc/pull/61/files#r574888254

If it now returns null, we can then update this to be something like:

```Typescript
const ContentsList = await getFileListFromS3(req.params.projectId + '/');
if(!ContentsList) {
// do error stuff
}
const contents = ContentsList.contents;
```

And then further down, rather than casting the content to any
```Typescript
      Contents.forEach(function (content: any) {
```
we should just be able to leave it without casting it, as the code will know that it is an S3 Object type (since we already ruled out it being null).
```Typescript
      Contents.forEach(function (content) {
```

api/src/paths/projects/{projectId}/artifacts/{fileName}/download.ts
```diff
@@ -0,0 +1,108 @@
+'use strict';
```
NickPhura, line 1: Instead of creating a folder called `{fileName}` and having a file inside called `download.ts` you can actually just have a file called `{fileName}.ts`

So the structure would be: `paths/projects/{projectId}/artifacts/{fileName}.ts`

This aligns better with the REST naming conventions, as the {fileName} is the resource you are targeting, and then its the GET, PUT, POST, etc, operations that represent the download, update, create, etc, actions against that target.

NickPhura: The content of the file would be the same (other than adjusting the relative import paths)

anissa-agahchen: Take a look at HATEOAS.  I was thinking it would be nice to adopt this style, where the action is placed at the end of the path.
 https://en.wikipedia.org/wiki/HATEOAS#:~:text=Hypermedia%20as%20the%20Engine%20of,provide%20information%20dynamically%20through%20hypermedia.
 
Happy to change it back.

api/src/utils/file-utils.ts
```diff
@@ -1,14 +1,17 @@
...
+import { GetObjectOutput, ManagedUpload, Metadata, ListObjectsOutput } from 'aws-sdk/clients/s3';
 import { S3_ROLE } from '../constants/roles';
 import { MediaBase64 } from '../models/media';
+import { getLogger } from './logger';
 
-const OBJECT_STORE_BUCKET_NAME = process.env.OBJECT_STORE_BUCKET_NAME;
-// const OBJECT_STORE_URL = process.env.OBJECT_STORE_URL || 'nrs.objectstore.gov.bc.ca';
-// const AWS_ENDPOINT = new AWS.Endpoint(OBJECT_STORE_URL);
+const defaultLog = getLogger('file_utils.ts');
+const projectId = '3';
```
NickPhura: leftover, can remove

api/src/utils/file-utils.ts
NickPhura: Same goes for the getFileFromS3 function above

api/src/utils/file-utils.ts
```diff
@@ -43,10 +65,17 @@ export async function getFileFromS3(key: string): Promise<GetObjectOutput> {
...
+      ETag: '',
+      Key: '',
+      Bucket: ''
+    };
   }
+  
+  const key = `${projectId}/${media.mediaName}`;
 
-  const key = `${uuidv4()}-${media.mediaName}`;
+  console.log('key: ', key);
```
NickPhura: extra log, can remove

api/src/paths/projects/{projectId}/artifacts/upload.ts
```diff
@@ -0,0 +1,148 @@
...
+
+      const metadata = {
+        filename: media.mediaName || '',
+        description: media.mediaDescription || '',
+        date: media.mediaDate || '',
+        username: (req['auth_payload'] && req['auth_payload'].preferred_username) || '',
+        email: (req['auth_payload'] && req['auth_payload'].email) || ''
+      };
+
+      s3UploadPromises.push(uploadFileToS3(media, metadata));
```
NickPhura: This needs an `await` otherwise it won't wait and will call `res.status(200)` on line 145 before the upload has actually finished.

We should also check the response, to see if we succeeded or not, and return a different code/response depending on it.

app/src/features/projects/ProjectsPage.tsx
```diff
@@ -83,6 +83,10 @@ const ProjectsPage: React.FC = () => {
     const getProjects = async () => {
       const projectsResponse = await biohubApi.getProjects();
 
+      const mediaList = await biohubApi.getMediaList('3');
+
+      console.log('mediaList: ', mediaList);
```
NickPhura: extra log, can remove

app/src/hooks/useBioHubApi.ts
```diff
@@ -117,13 +123,39 @@ export const useBiohubApi = () => {
...
+  /**
+   * Fetch the media list.
+   *
+   * @return {*} {Promise<IMedia[]>}
+   */
+
+  const getMediaList = async (projectId: string): Promise<IMedia[]> => {
+    //const api = await apiPromise;
+
+    const { data } = await api.get('/api/projects/' + projectId + '/artifacts/list');
```
NickPhura: This is maybe a subjective thing, but I think its nicer if we use template strings:
```Typescript
const { data } = await api.get(`/api/projects/${projectId}/artifacts/list`);
```

They are less prone to human error as well, as there are less quotes and concatenations to worry about

app/src/hooks/useBioHubApi.ts
```diff
@@ -117,13 +123,39 @@ export const useBiohubApi = () => {
...
+
+    const mediaKeyList: IMedia[] = [];
+
+    if (data && data.length > 0) {
```
NickPhura: This is also somewhat subjective, but I find it makes for cleaner code if instead of wrapping the happy path in an if statement, we wrap the unhappy path instead, and fail fast.  So in this example:

```Typescript
const mediaKeyList: IMedia[] = [];

if(!data || !data.length) {
  return mediaKeyList;
}

data.forEach((file: any) => {...});

return mediaKeyList;
```
If data is null or has no length, we return immediately.  It also removes 1 level of nesting from the 'real' forEach logic below, which makes it a bit easier to read.

testing/integration/postman/BioHubBC-Environment-DEV.postman_environment.json
```diff
@@ -1,5 +1,5 @@
 {
```
NickPhura, line 1: We can just revert this file entirely, as we never use the object store env vars anywhere in postman.

testing/integration/postman/BioHubBC-Environment-DEV.postman_environment.json
```diff
@@ -1,5 +1,5 @@
...
-	"id": "65ce4931-9611-4df9-bca1-bc2340e0810a",
+	"id": "d9ec52f3-e88a-45ee-867d-116e25294e38",
```
NickPhura: We can just revert this file entirely, as we never use the object store env vars anywhere in postman.

api/src/paths/projects/{projectId}/artifacts/{fileName}/download.ts
NickPhura: Hmm, interesting.  I've never heard of that concept by name, though I've made apis with that sort of thing.  Usually it would come up because of some non-standard actions like "publish/unpublish".  

Okay, well lets leave it as you have it for now then.

api/src/paths/projects/{projectId}/artifacts/upload.ts
NickPhura: That being said, I wouldn't kill yourself trying to find a way to simulate an error or anything.  I would just go based on what the doc says as far as what it will return (as realistically S3 is never itself going to fail and It will be that we send it a bad key, or have a network issue or something).

anissa-agahchen: I refactored this file a bit ... I think it's overall in better shape now.

NickPhura (approved):
Looks good!  There were a couple nitpicky things I noticed, but nothing worth delaying this ticket over.

anissa-agahchen:
Awesome, we'll catch them on the next round.

PR #65, opened 2021-02-17
BHBC-794: Species section, multiselect widget + tests

NickPhura:
# Overview

## Links to jira tickets

https://quartech.atlassian.net/browse/BHBC-794

- Add multi-select autocomplete field
- Add species section
  - Update project endpoint to also save species info
  - Add migration to update db models to add species tables
- Add some hardcoded species values to temporarily satisfy the species dropdowns
  - These particular values are just a concatenation of some species names/latin names from the CDC public search for non-critical species, so definitely a complete or final version.
  - This will need to be flushed out further, before release, once we get a better handle on the whole taxon topic

- [ ] Version change

## Checklist

### General

### Code

### Style

- [x] My code uses native material-ui components/icons/conventions when possible

### Documentation

### Tests

- [x] I have added/updated unit tests for any code I've added/updated
- [x] I have added/updated the Postman requests/tests to account for any API endpoints I've added/updated

### Linting/Formatting

### SonarCloud

- [x] I have addressed all SonarCloud Bugs, Vulnerabilities, Security Hotspots, and Code Smells

## How Has This Been Tested?

## Screenshots


NickPhura:
@rstens Most of the duplication is coming from the unit tests.  Can we disable test files from being checked for duplicate code?  I also rejected some smells it identified for empty arrow functions, etc, which are normal in the tests to stub out functions, etc.

Edit: I took a whack at updating the `.sonarcloud.properties` file, which is included in this PR.  Looks like its working as desired.

anissa-agahchen (changes requested):
It looks good overall.  The main issue I found:
* when a species is selected in the focal species list, or the ancillary species list, it should not be available to select again

Less important, but worth noting:
* the postman file to be reverted completely, I think.... I couldn't find any intentional changes in that file.

rstens (commented):
Do we have a standard list of API HTTP return codes that we adhere to?

Also: Just to be clear: we are not using knex as a query builder?

rstens:
> @rstens Most of the duplication is coming from the unit tests. Can we disable test files from being checked for duplicate code? I also rejected some smells it identified for empty arrow functions, etc, which are normal in the tests to stub out functions, etc.
> 
> Edit: I took a whack at updating the `.sonarcloud.properties` file, which is included in this PR. Looks like its working as desired.

Should be fine.

NickPhura:
@anissa-agahchen 
> 
> 
> It looks good overall. The main issue I found:
> 
>     * when a species is selected in the focal species list, or the ancillary species list, it should not be available to select again
> 

In the interest of progress I think this will have to remain as it is for now.  There isn't a good solution to account for this that wont also take a bunch of work.  If it becomes an issue later we can see what can be done to update it.

> 
> Less important, but worth noting:
> 
>     * the postman file to be reverted completely, I think.... I couldn't find any intentional changes in that file.

There were no brand new endpoints, but I did update the `POST project` endpoint.  I added the new coordinator and species fields to the json object.

NickPhura:
@rstens 
> 
> 
> Do we have a standard list of API HTTP return codes that we adhere to?

Mostly yes.  

In the root openapi spec, there are a number of common responses listed, which are mostly for error cases (401, 403, 409, 500, 503, etc).  These are then referenced as possible responses in each of the endpoint openapi specs, alongside the 200 happy path response.  Right now, I just tend to list all of them, even if I only intentionally throw a few.  Something to refine later on.

That being said, there is no response validation currently, so there is nothing stopping us from throwing a code that isn't in the list of responses.  There is some response validation I believe that can be leveraged, but I haven't investigated how it works yet..

(One thing I want to do at some point, is make some util functions for throwing errors, so we have some pre-made documented `throw500('some message', error)` type calls that can be made without having to think about it.)

>
> 
> Also: Just to be clear: we are not using knex as a query builder?

We have not been using it, correct.  Charlie is leveraging the fancy psql modelling tool that Quartech bought a license for.  
And then we have either been running those sql files directly, by dropping them unchanged into the knex raw template string, or in the case of smaller one-off changes, just copying the sql contents directly into the knex raw template string.

anissa-agahchen (approved):
I'll throw a ticket in the backlog, about the species, and we can deal with it when/if there is a need

PR #80, opened 2021-02-24
767: Adding spatial columns

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-767

- 767: Adding spatial columns to project table
- geog
- geom

## How Has This Been Tested?

- Locally using DBeaver

### project table in `biohub`
Screenshot 2021-02-24 151631

### project view in `biohub_dapi_v1`
Screenshot 2021-02-24 151747

### DIFF
--- a/database/src/migrations/20210224144754_alter_project_table.ts
+++ b/database/src/migrations/20210224144754_alter_project_table.ts
@@ -1,0 +1,66 @@
+import * as fs from 'fs';
+import * as Knex from 'knex';
+import path from 'path';
+
+const DB_SCHEMA = process.env.DB_SCHEMA;
+
+/**
+ * Alter the project table:
+ * - add spatial columns
+ * - geom (geometry)
+ * - geog (geography)
+ *
+ * Regenerate the dapi views.
+ *
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set search_path = ${DB_SCHEMA};
+
+    ALTER TABLE project ADD COLUMN geom geometry(Geometry,3005) CHECK (st_isValid(geom));
+    COMMENT ON COLUMN project.geom IS 'Spatial geometry';
+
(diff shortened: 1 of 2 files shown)


database/src/migrations/20210224144754_alter_project_table.ts
```diff
@@ -0,0 +1,68 @@
...
+ * Regenerate the dapi views.
+ *
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set schema '${DB_SCHEMA}';
+    set search_path = ${DB_SCHEMA},public;
```
NickPhura, line 22: Does this still work if you don't include `public`?

Ideally, we don't reference the public schema anywhere, as its been locked down as much as possible (for security reasons).  But, I know the postgis stuff gets added under public, so not sure if that is a factor here.

sdevalapurkar: yup exactly why I had to add it in, to get access to the `geometry` type

rstens (commented):
Since all database definitions start in Charlie's model, we need to be careful adding new columns without him confirming that he updated the model.

database/src/migrations/20210224144754_alter_project_table.ts
rstens: That is one of the compromises we need to make to make that work.

rstens (changes requested):
I do not know if I agree with creating views out of a file instead of a proper migration. It seems to me that that mechanism is more fit for seeding.

NickPhura:
> 
> 
> I do not know if I agree with creating views out of a file instead of a proper migration. It seems to me that that mechanism is more fit for seeding.

@rstens Are you referring to the dapi_views?  Those need to get re-generated every time we modify a table, as they won't automatically pick up changes on their own.  The same thing is done when Charlie produces the SQL.  You can see it in the other migrations/sql files that he authored.

OK

rstens (approved):

PR #95, opened 2021-03-03
Project POST endpoint to accept geometry collection

sdevalapurkar:
# Overview

- Project POST endpoint to accept geometry collection

## How Has This Been Tested?

- Locally


NickPhura (dismissed):

NickPhura:
Dont merge this during the demo though (just in case)!

api/src/queries/project-queries.test.ts
```diff
@@ -14,38 +14,75 @@ describe('postProjectSQL', () => {
...
-
+      const postLocationData = new PostLocationData(locationDataWithGeo);
       const response = postProjectSQL({ ...postProjectData, ...postCoordinatorData, ...postLocationData });
 
       expect(response).to.not.be.null;
+      expect(
+        response?.values.includes(
+          '{"type":"Polygon","coordinates":[[[-128,55],[-128,55.5],[-128,56],[-126,58],[-128,55]]]}'
+        )
+      ).to.not.be.null;
```
NickPhura: You can probably leverage the `deep` equality matcher if you want to assert that the content matches exactly: https://www.chaijs.com/api/bdd/#method_deep

NickPhura: Otherwise I think this always passes? `expect(boolean | undefined).to.not.be.null` Unless false is considered a null type value.

sdevalapurkar: thank you! nice catch :)

database/src/migrations/20210107132219_setup_biohub_db.ts
```diff
@@ -38,7 +38,7 @@ export async function up(knex: Knex): Promise<void> {
   );
 
   await knex.raw(`
-    REVOKE ALL PRIVILEGES ON SCHEMA public FROM PUBLIC;
+    REVOKE CREATE ON SCHEMA public FROM PUBLIC;
```
NickPhura, line 41: keep in mind that this wont actually run in dev, unless we blow the dev pods away and remake them (which we can do for now, since we dont have any data we care about losing).

NickPhura (approved):

database/src/migrations/20210107132219_setup_biohub_db.ts
sdevalapurkar: how do we do this?

anissa-agahchen (approved):
Looks good.

PR #100, opened 2021-03-05
779/875: Location Shapefile Upload onto Map

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-779
- https://quartech.atlassian.net/browse/BHBC-736
- https://quartech.atlassian.net/browse/BHBC-875

- Location Shapefile Upload onto Map
- Map component with upload KML functionality in location step of project workflow
- Can edit uploaded kml geometry, or delete it if you wish
- Extent gets set to show the uploaded geo(s) effectively and reduce user scrolling

### General

## How Has This Been Tested?

- Locally

## Screenshots

https://user-images.githubusercontent.com/28017034/110386296-f7cc6180-8014-11eb-9133-d4bcf3ddadc1.mp4


app/src/components/map/MapContainer.tsx
```diff
@@ -1,31 +1,48 @@
...
+}
+
+const MapBounds: React.FC<IMapBoundsProps> = (props) => {
+  const map = useMap();
+  const { bounds } = props;
+
+  map.fitBounds(bounds);
+
+  return null;
+};
```
NickPhura: Whats this for?

Its a component, but it doesn't return any html. Could this just be a function instead?

app/src/features/projects/components/ProjectLocationForm.tsx
```diff
@@ -32,10 +44,31 @@ export interface IProjectLocationFormProps {
...
+  const [geometry, setGeometry] = useState<Feature[]>([]);
+
+  const handleSpatialUpload = async (e: any) => {
+    const file = e.target.files[0];
+    const fileAsString = await file.text().then((xmlString: string) => {
+      return xmlString;
+    });
+    const domKml = new DOMParser().parseFromString(fileAsString, 'application/xml');
+    const geojson = kml(domKml);
+    const bboxCoords = bbox(geojson);
```
NickPhura: If I pick a non spatial file, it throws an error in the console, but seems to be okay.  If I try picking something silly like a video file, it hangs my browser.  We might need to do some checking on the mine type of the file, to ensure it is some kind of acceptable type (whatever they are for spatial files), before we try to read/parse it.

We might also need to wrap this in a try/catch and display some kind of error message if it fails or if the user selects a bad file type.

NickPhura:
Can you make the initial map extent be all of BC (or there abouts)?

When I draw a shape, go to the next step, and then go back to the previous step, my shape is gone.

This doesn't send the data to the API yet right? Is that part of a different ticket? No worries if it is.  When you do go to do the actual saving it should be possible to leverage formik to handle the values still.  We should be able to have a formik value (with no rendered UI) that we read/write the geojson from/to, and then it will tie seamlessly into the rest of the form and can be accessed like all other fields.

app/src/components/map/MapContainer.tsx
sdevalapurkar: No I dont think it can be a function cause it uses a hook called useMap which is part of react leaflet and it expects it to be a child component of the leaflet map container

app/src/features/projects/components/ProjectLocationForm.tsx
sdevalapurkar: yeah for sure

sdevalapurkar:
> Can you make the initial map extent be all of BC (or there abouts)?
> 
> When I draw a shape, go to the next step, and then go back to the previous step, my shape is gone.
> 
> This doesn't send the data to the API yet right? Is that part of a different ticket? No worries if it is. When you do go to do the actual saving it should be possible to leverage formik to handle the values still. We should be able to have a formik value (with no rendered UI) that we read/write the geojson from/to, and then it will tie seamlessly into the rest of the form and can be accessed like all other fields.

might discuss this with you tomorrow as I have not used formik before

sdevalapurkar:
> Can you make the initial map extent be all of BC (or there abouts)?
> 
> When I draw a shape, go to the next step, and then go back to the previous step, my shape is gone.
> 
> This doesn't send the data to the API yet right? Is that part of a different ticket? No worries if it is. When you do go to do the actual saving it should be possible to leverage formik to handle the values still. We should be able to have a formik value (with no rendered UI) that we read/write the geojson from/to, and then it will tie seamlessly into the rest of the form and can be accessed like all other fields.

done - it is now showing all of bc on initial render

API work in separate ticket and PR-would like to get this one merged in soon 😄

sdevalapurkar:
just realized there is no need for api changes for my geom and map stuff - it is all working and set up already :)

app/src/features/projects/components/ProjectLocationForm.tsx
```diff
@@ -32,9 +47,54 @@ export interface IProjectLocationFormProps {
...
+  const { values, touched, errors, handleChange, handleSubmit, setFieldValue } = formikProps;
+
+  const [bounds, setBounds] = useState<any>([]);
+  const [uploadError, setUploadError] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    setIsLoading(false);
+
+    if (!values.geometry.length || (values.geometry.length === 1 && values.geometry[0].geometry.type === 'Point')) {
```
NickPhura, line 63: The length === 1 check seems pretty specific, whats the story there?

Any concern over having multiple geometries, but they are all points?

sdevalapurkar: if there is one geo only and it is a point, we cannot do the bound setting - because leaflet does not know how to handle that and tries to zoom in way too much
if there are multiple points or a polygon and a point, it is fine and knows what to do - thus, the === 1 check :)

NickPhura: gotcha, maybe just add that as a comment?

sdevalapurkar: yes sir 😺 good point, will do

NickPhura (approved):

PR #123, opened 2021-03-12
875: Map boundary and project structure

anissa-agahchen:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-875

- GET for project details
- Map and location info (boundary) on view project page

- [ ] Version change

## How Has This Been Tested?

- Locally

## Screenshots

sauce


api/src/paths/project/{projectId}.ts
```diff
@@ -82,11 +84,37 @@ function getProject(): RequestHandler {
...
+      ]);
+
+      defaultLog.debug({ label: 'getProjectWithDetails', message: 'test' });
 
       await connection.commit();
 
-      const result = (createResponse && createResponse.rows && createResponse.rows[0]) || null;
+      connection.release();
+
+      await connection.commit();
```
NickPhura: We only want to commit here once, and we don't need to release here as that gets called in the 'finally' block at the bottom already.

app/src/interfaces/project-interfaces.ts
```diff
@@ -38,6 +26,19 @@ export interface IProject {
...
+ * An interface representing the project table.
+ *
+ * @export
+ * @interface IProject
+ */
+export interface IProjectWithDetails {
+  id: number;
+  project: IProjectDetailsForm;
+  objectives: IProjectObjectivesForm;
+  location: IProjectLocationForm;
```
NickPhura, line 39: I think there are some tweaks we can make here. The general structure is spot on, but it will need to use different interfaces (not the Form ones). 

The Form interfaces only account for the fields the user interacts with, but in our response from the API we'll have extra fields like the primary key and revision_count for each row, as well as any additional details that are generated by the backend (if any). 

That being said, if its working now, then lets not worry about it in this PR so that we don't hold up progress on the rest of the dependent tickets.  We'll have to do some refinement inevitably anyways as we start putting more of the view pages together, etc.

anissa-agahchen: My intention with using the form interfaces was to be ready for the upcoming sprints where we need to edits each of these sections individually.  Not sure what the plan is at this point for those, but it would be good to re-use the form interfaces?

NickPhura: Gotcha, that makes sense.  In that case, we can still make new interfaces, but we can extend the form interfaces to ensure that those fields are present, but still allow us to include extra fields

app/src/features/projects/view/components/GeneralInformation.test.tsx
```diff
@@ -1,16 +1,32 @@
...
 import GeneralInformation from './GeneralInformation';
 
-const projectData = {
-  name: 'Test Project Name',
-  start_date: '1998-10-10',
-  end_date: '2021-02-26'
+const projectWithDetailsData: IProjectWithDetails = {
+  id: 1,
+  project: {
+    project_name: 'Test Project Name',
```
sdevalapurkar, line 9: need to add project_type_name here

sdevalapurkar (approved):

PR #133, opened 2021-03-16
Fix code smells and code cleanup

sdevalapurkar:
## Overview

Code cleanup (app-ionic and app-common-components) and code smells


anissa-agahchen (approved):
Looks good!

PR #142, opened 2021-03-17
Remove section from pr template cause jira lint does it for us

sdevalapurkar:
# Overview

Remove section from pr template cause jira lint does it for us

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,5 @@
 # Overview
 
-## Links to jira tickets
-
-- {Include links to all of the applicable Jira tickets}
-
 ## This PR contains the following changes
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,9 +1,5 @@
 # Overview
 
-## Links to jira tickets
-
-- {Include links to all of the applicable Jira tickets}
```
NickPhura, line 5: This does let us account for multiple tickets, which the jira action doesn't.

sdevalapurkar: Yeah was kinda thinking that but thought those prs it isn't super important to link the ticket to - but wanted to get thoughts :)

Happy to keep this, was more to start a discussion. Will close this PR for now 😇

PR #149, opened 2021-03-19
Code Cleanup

sdevalapurkar:
# Overview

Reduce code duplication
Add project type code table queries


anissa-agahchen (approved):
Looks good to me

PR #178, opened 2021-03-29
UI fixes and test additions

sdevalapurkar:
# Overview

- UI fixes and test additions

## How Has This Been Tested?

- Locally

### DIFF
--- a/app/src/components/dialog/EditDialog.tsx
+++ b/app/src/components/dialog/EditDialog.tsx
@@ -1,4 +1,6 @@
 import { Box, Button, Dialog, DialogActions, DialogContent, DialogTitle } from '@material-ui/core';
+import useMediaQuery from '@material-ui/core/useMediaQuery';
+import { useTheme } from '@material-ui/core/styles';
 import React from 'react';
 import { Formik } from 'formik';
 
@@ -63,6 +65,9 @@ export interface IEditDialogProps {
  * @return {*}
  */
 export const EditDialog: React.FC<IEditDialogProps> = (props) => {
+  const theme = useTheme();
+  const fullScreen = useMediaQuery(theme.breakpoints.down('sm'));
+
   return (
     <Box>
       <Formik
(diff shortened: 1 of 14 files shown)


app/src/themes/appTheme.ts
```diff
@@ -73,7 +73,10 @@ const appTheme = createMuiTheme({
     },
     MuiDialog: {
       paperWidthXl: {
-        maxWidth: '800px'
+        width: '600px',
+        '@media (min-width: 800px)': {
```
NickPhura: You should be able to use breakpoints here rather than a media query

NickPhura: And then the 800px isnt a hardcoded value, and matches whatever we set for the sm/md/lg/etc

app/src/components/dialog/EditDialog.tsx
```diff
@@ -76,6 +81,7 @@ export const EditDialog: React.FC<IEditDialogProps> = (props) => {
         }}>
         {(formikProps) => (
           <Dialog
+            fullScreen={fullScreen}
             maxWidth="xl"
```
NickPhura, line 85: I wonder if there is some way to leverage breakpoints to make the maxWidth dynamic.  In combination with making the width always 100%.  So for example, if the screen is XL, then the dialog is 100% of maxWidth="md" (or something like that).  And as the screen gets smaller, we decrease the maxWidth less and less, until its the same as the screen size (ie: when the screen is sm, maxWidth="sm" and since width is 100% it is fullscreen)

sdevalapurkar: hmm yeah possibly - might be something to discuss in further detail with Jeremy 
For now, I think this is a good fix for our dialog issues and more detailed fixes can be looked into later for smaller viewports :) 
(as per discussion with jer this morning)

sdevalapurkar:
Lesgoooo!!! We past 50% overall test coverage and 4 frontend components are now at 100%! ❗ 😺  @NickPhura @anissa-agahchen

anissa-agahchen (approved):

PR #189, opened 2021-03-31
Yup funding input

anissa-agahchen:
Screen Shot 2021-03-31 at 1 56 13 PM
# Overview

## Links to jira tickets
Set a limit on funding amount to match what the DB expects.


sdevalapurkar (approved):
I think you're adding a code smell here.... 😉

anissa-agahchen:
> I think you're adding a code smell here.... 😉

app/src/features/projects/components/ProjectFundingItemForm.tsx
```diff
@@ -53,6 +53,7 @@ export const ProjectFundingFormArrayItemYupSchema = yup.object().shape({
     .transform((value) => (isNaN(value) && null) || value)
     .typeError('Must be a number')
     .min(0, 'Must be a positive number')
+    .max(9999999999)
```
sdevalapurkar: this makes the error message that shows up say:
```
funding_amount must be less than.....
```

would it somehow be possible to change the message to say:
```
Funding amount must be less than.....
```

maybe by doing:
```
.max(9999999999, 'Funding amount must be less than...')
```

sdevalapurkar (changes requested):

app/src/features/projects/components/ProjectFundingItemForm.tsx
NickPhura: "Funding amount must be less than $9,999,999,999"

anissa-agahchen: updated

sdevalapurkar (approved):

PR #232, opened 2021-04-14
Build time improvements (maybe??)

NickPhura:
# Overview

Just want to see if this makes any noticeable improvement.  
Not exactly sure how, I guess see if we can compare this app build time with past app build times.
Can also do some local builds to compare this version to the old version and see how long they take.

## Changes

Replace all de-reference style imports (whatever these are called) with individual imports.
In theory will improve performance as only the items we use are imported, and not the all of material-ui.

IE:

Change
```Typescript
import { Box, TextField} from '@materail-ui/core';
```
to
```Typescript
import Box from '@materail-ui/core/Box';
import TextField from '@materail-ui/core/TextField';
```
Etc...

Also deleted the `rjsf` folder as the whole thing is unused.


rstens (dismissed):
Went from 64 seconds to 47 seconds on my machine - 27% improvement. I'd say that is worth it considering everything else i tried had literally no effect. 👍

sdevalapurkar (dismissed):

NickPhura:
> 
> 
> Went from 64 seconds to 47 seconds on my machine - 27% improvement. I'd say that is worth it considering everything else i tried had literally no effect. 👍

Yeah I was noticing about a 15 second improvement locally on my machine: ~20%

sdevalapurkar (approved):

anissa-agahchen (approved):

PR #237, opened 2021-04-16
Adding a new funding source test case

sdevalapurkar:
# Overview

Adding a new funding source test case

Finally got it @NickPhura @anissa-agahchen 😄 It is super cool!
Bit more than 15 mins 😉


anissa-agahchen (approved):
Good work man!  That test was tricky!

PR #238, opened 2021-04-16
Implement functionality to edit draft projects

sdevalapurkar:
# Overview

- Implement functionality to edit draft projects

## How Has This Been Tested?

- Locally


app/src/features/projects/CreateProjectPage.tsx
```diff
@@ -159,6 +163,38 @@ const CreateProjectPage: React.FC = () => {
...
+      }
+
+      setInitialProjectFieldData(response.data);
+    };
+
+    if (!queryParams.draftId) {
+      return;
+    }
+
+    getDraftProjectFields();
```
NickPhura, line 196: Small optimization: can we check for the existence of queryParams.draftId before we make this call? If its null then we can skip this altogether and it will save a bit of loading time.

NickPhura: We also probably want to check for hasLoadedDraftData === false too, so that we only ever load it once

NickPhura: Wait, ignore my first comment, I cant read apparently.  Second one is still valid I think...?

app/src/features/projects/CreateProjectPage.tsx
```diff
@@ -376,45 +414,81 @@ const CreateProjectPage: React.FC = () => {
...
+   */
+  const deleteDraft = async (draftId: number) => {
+    const response = await biohubApi.draft.deleteDraft(draftId);
+
+    if (!response) {
+      showCreateErrorDialog({
+        dialogError: 'Your project was created successfully, but your old draft was not deleted successfully.'
+      });
+    }
+  };
```
NickPhura: How does this impact the routing (to the view or list page) that normally happens right after a create project request?

app/src/features/projects/CreateProjectPage.tsx
sdevalapurkar: I already am making a check and this line will not be executed if draftId does not exist 😄

sdevalapurkar: hasloadeddraftdata is based on queryparams.draftId so we can just check it alone

app/src/features/projects/CreateProjectPage.tsx
sdevalapurkar: thanks

anissa-agahchen (dismissed):
Looks good

anissa-agahchen (approved):

PR #249, opened 2021-04-20
Active Users List

sdevalapurkar:
# Overview

- Active Users List

## How Has This Been Tested?

- Locally

## Screenshots

image


anissa-agahchen (dismissed):

api/src/security/auth-utils.ts
```diff
@@ -215,6 +215,10 @@ export const userHasValidSystemRoles = function (
     userSystemRoles = [userSystemRoles];
   }
 
+  if (!validSystemRoles.length) {
+    return true;
+  }
+
```
NickPhura: I ended up adding a check at the top of the function (206-208), so this shouldn't be needed anymore.

Having the check here caused a funny issue where some falsy values would still get added to the array, and then the length wouldn't be 0, even though it didn't really contain a role.

NickPhura: Noticed during unit testing :)

app/src/features/admin/users/AccessRequestList.tsx
```diff
@@ -97,7 +97,6 @@ const AccessRequestList: React.FC = () => {
               <TableCell>Regional Offices</TableCell>
               <TableCell>Request Date</TableCell>
               <TableCell>Status</TableCell>
-              <TableCell></TableCell>
```
NickPhura: I think this might still be needed?

There are 7 columns in the data row, the 7th being for the 'Review' button which doesn't get an actual header, but might still need the placeholder to block the space?

NickPhura: Ya it is (tested locally)

api/src/security/auth-utils.ts
sdevalapurkar: awesome

app/src/features/admin/users/AccessRequestList.tsx
sdevalapurkar: oops my bad, didn't mean to touch your file at all - this is definitely needed :) may have accidentally deleted something from this file when I was trying to delete from `ActiveUsersList.tsx` 😄 

good catch

NickPhura (approved):

PR #262, opened 2021-04-28
Get user.ts to 100% coverage

sdevalapurkar:
# Overview

Get user.ts to 100% coverage

image

### DIFF
--- a/api/src/paths/user/self.test.ts
+++ b/api/src/paths/user/self.test.ts
@@ -10,7 +10,7 @@ import SQL from 'sql-template-strings';
 
 chai.use(sinonChai);
 
-describe('getUser', function () {
+describe('getUser', () => {
   afterEach(() => {
     sinon.restore();
   });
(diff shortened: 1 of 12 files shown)


api/src/paths/user/self.test.ts
```diff
@@ -64,7 +64,7 @@ describe('getUser', function () {
       const result = self.getUser();
 
       // eslint-disable-next-line
-      //@ts-ignore
```
NickPhura: If we cast the value to an `any` does it solve this issue (or possibly the unknown/any thing)?  Might be nicer than ignores everywhere?

sdevalapurkar: the issue is with it missing like 76 fields that I dont need to mock out

NickPhura: Should be able to do something like:

```Typescript
  const sampleReq = {
    keycloak_token: {}
  } as any;
    
  await result(sampleReq, (null as unknown) as any, (null as unknown) as any);
```

And then won't need any of the ignores

sdevalapurkar: thanks!

NickPhura (dismissed):

NickPhura (approved):

PR #264, opened 2021-04-29
Aa more tests

anissa-agahchen:
Tests for authorize and much of authenticate

### DIFF
--- a/api/src/security/auth-utils.test.ts
+++ b/api/src/security/auth-utils.test.ts
@@ -273,25 +273,36 @@ describe('authorize', function () {
       expect(actualError.message).to.contain('Access Denied');
     }
   });
+
   it('should authorize a user with valid roles', function () {
-    const stubbedSystemUser: {
-      id: 1;
-      user_identifier: 'identifier';
-      role_ids: [1, 2];
-      role_names: ['role 1', 'role 2'];
-    };
-
-    sinon.stub(auth_utils, 'getSystemUser').returns(stubbedSystemUser);
-
-    // const userObject = {
-    //       id: 1,
-    //       user_identifier: 'identifier',
-    //       role_ids: [1, 2],
-    //       role_names: ['role 1', 'role 2']
+    // const sampleReq = {
+    //   keycloak_token: {}
     // };
 
-    // const hasValidSystemRole = auth_utils.authorize({ keycloak_token: 'some token' }, ['abc']);
+    // let expectedResult = {
+    //   id: null,
+    //   user_identifier: null,x
+    //   role_ids: null,
+    //   role_names: null
+    // };
 
-    // expect(result).to.be.true;
+    // const sampleRes = {
+    //   status: (status: number) => {
+    //     return {
+    //       json: (result: any) => {
+    //         expectedResult = result;
+    //       }
+    //     };
+    //   }
+    // };
+    const stubbedSystemUser= {
+      id: 1,
+      user_identifier: 'identifier',
+      role_ids: [1, 2],
+      role_names: ['role 1', 'role 2']
+    };
+
+    sinon.stub(auth_utils, 'getSystemUser').resolves(stubbedSystemUser);
+
   });
 });
--- a/api/src/security/auth-utils.test.ts
+++ b/api/src/security/auth-utils.test.ts
@@ -257,7 +257,9 @@ describe('authorize', function () {
     } catch (actualError) {
       expect(actualError).instanceOf(HTTP403);
     }
+  });
 
+  it('throws HTTP403 when the keycloak_token is undefined', async function () {
     try {
       await auth_utils.authorize(undefined, ['abc']);
       expect.fail();
@@ -283,21 +285,17 @@ describe('authorize', function () {
   });
 
   it('throws HTTP403 when stubbed getSystemUser throws error', async function () {
-    sinon.stub(auth_utils, 'getSystemUser').callsFake(async function (keycloak_token: object) {
-      if (!keycloak_token || keycloak_token) {
-        throw new Error();
-      }
-    });
-
+    sinon.stub(auth_utils, 'getSystemUser').rejects(new Error());
     try {
       await auth_utils.authorize({ keycloak_token: 'any token' }, ['abc']);
       expect.fail();
     } catch (actualError) {
       expect(actualError).instanceOf(HTTP403);
+      expect(actualError.message).to.equal('Access Denied');
     }
   });
 
-  it('throws HTTP403 when stubbed getSystemUser returns user with no roles', async function () {
+  it('throws HTTP403 when userHasValidSystemRoles returns falsie', async function () {
     sinon.stub(auth_utils, 'getSystemUser').resolves({
       id: 0,
       user_identifier: 'somebody',
@@ -322,12 +320,8 @@ describe('authorize', function () {
       role_names: []
     });
 
-    try {
-      const result = await auth_utils.authorize({ keycloak_token: 'any token' }, ['abc']);
-      expect(result).to.be.true;
-    } catch {
-      expect.fail();
-    }
+    const result = await auth_utils.authorize({ keycloak_token: 'any token' }, ['abc']);
+    expect(result).to.be.true;
   });
 });
 
@@ -335,12 +329,14 @@ describe('authenticate', function () {
   it('throws HTTP401 when authorization headers were null or missing', async function () {
     try {
       await auth_utils.authenticate(undefined);
+      expect.fail();
     } catch (actualError) {
       expect(actualError).instanceOf(HTTP401);
     }
 
     try {
       await auth_utils.authenticate({});
+      expect.fail();
     } catch (actualError) {
       expect(actualError).instanceOf(HTTP401);
     }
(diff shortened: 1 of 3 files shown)


api/src/security/auth-utils.test.ts
```diff
@@ -270,7 +278,161 @@ describe('authorize', function () {
...
-      expect(actualError.message).to.contain('Access Denied');
+      expect(actualError).instanceOf(HTTP403);
+    }
+  });
+
+  it('throws HTTP403 when stubbed getSystemUser throws error', async function () {
+    sinon.stub(auth_utils, 'getSystemUser').callsFake(async function (keycloak_token: object) {
+      if (!keycloak_token || keycloak_token) {
+        throw new Error();
+      }
```
NickPhura: nit: you don't need the if statement, you can just throw the error.

api/src/security/auth-utils.test.ts
```diff
@@ -254,7 +255,14 @@ describe('authorize', function () {
...
-      expect(actualError.message).to.contain('Access Denied');
+      expect(actualError).instanceOf(HTTP403);
+    }
+
+    try {
+      await auth_utils.authorize(undefined, ['abc']);
+      expect.fail();
+    } catch (actualError) {
+      expect(actualError).instanceOf(HTTP403);
     }
```
NickPhura, line 266: Can we move this into its own `it`: throws when keycloak_token is null

api/src/security/auth-utils.test.ts
```diff
@@ -254,7 +255,14 @@ describe('authorize', function () {
...
     } catch (actualError) {
-      expect(actualError.message).to.contain('Access Denied');
+      expect(actualError).instanceOf(HTTP403);
+    }
+
+    try {
+      await auth_utils.authorize(undefined, ['abc']);
+      expect.fail();
+    } catch (actualError) {
+      expect(actualError).instanceOf(HTTP403);
```
NickPhura, line 265: Can we move this into its own `it`: throws when keycloak_token is null

api/src/security/auth-utils.test.ts
```diff
@@ -270,7 +278,161 @@ describe('authorize', function () {
...
+
+    try {
+      await auth_utils.authorize({ keycloak_token: 'any token' }, ['abc']);
+      expect.fail();
+    } catch (actualError) {
+      expect(actualError).instanceOf(HTTP403);
+    }
+  });
+
+  it('throws HTTP403 when stubbed getSystemUser returns user with no roles', async function () {
```
NickPhura: nit: technically this test is checking that an error is thrown when `userHasValidSystemRoles` returns falsie.

api/src/security/auth-utils.test.ts
anissa-agahchen: Yeah, this is a wierd one.  If I don't check for the keycloak_token, I get:

  authenticate
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - authorization headers were null or missing  
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - unexpected error - Access Denied 
[object Object] 
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - authorization headers were null or missing  
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - unexpected error - Access Denied 
[object Object] 
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - authorization headers were null or missing  
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - unexpected error - Access Denied 
[object Object] 
    ✓ throws HTTP401 when authorization headers were null or missing
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - authorization header did not have a bearer  
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - unexpected error - Access Denied 
[object Object] 
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - token string was null  
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - unexpected error - Access Denied 
[object Object] 
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - decoded token was null  
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - unexpected error - Access Denied 
[object Object] 
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - decoded token header kid was null  
[2021-04-29 09:12:46] (warn) (security/auth-utils): authenticate - unexpected error - Access Denied 
[object Object] 
[2021-04-29 09:12:47] (warn) (security/auth-utils): authenticate - unexpected error - Unable to find a signing key that matches 'some id' 
SigningKeyNotFoundError: Unable to find a signing key that matches 'some id'

api/src/security/auth-utils.test.ts
```diff
@@ -270,7 +278,161 @@ describe('authorize', function () {
...
+      user_identifier: 'somebody',
+      role_ids: [],
+      role_names: []
+    });
+
+    try {
+      const result = await auth_utils.authorize({ keycloak_token: 'any token' }, ['abc']);
+      expect(result).to.be.true;
+    } catch {
+      expect.fail();
```
sdevalapurkar: shouldnt the expect fail be in the try and the catch needs to actually assert the error?

api/src/security/auth-utils.test.ts
sdevalapurkar: we can go over this

api/src/security/auth-utils.test.ts
```diff
@@ -270,7 +278,161 @@ describe('authorize', function () {
...
+  it('throws HTTP401 when authorization header containing valid bearer token', async function () {
+    try {
+      await auth_utils.authenticate({
+        headers: {
+          authorization: 'Not a bearer token'
+        }
+      });
+    } catch (actualError) {
+      expect(actualError).instanceOf(HTTP401);
+    }
```
sdevalapurkar: all of these kinds of sections need the expect.fail()

api/src/security/auth-utils.test.ts
```diff
@@ -270,7 +278,161 @@ describe('authorize', function () {
...
+  //       // callback && callback(null, undefined);
+  //     });
+
+  //   try {
+  //     await auth_utils.authenticate({});
+  //   } catch (actualError) {
+  //     expect(actualError).instanceOf(HTTP401);
+  //   }
+  // });
+});
```
sdevalapurkar: remove this? we can work on figuring it out later but for now we can remove

api/src/security/auth-utils.test.ts
```diff
@@ -270,7 +278,161 @@ describe('authorize', function () {
...
+    try {
+      await auth_utils.authenticate({
+        headers: {}
+      });
+    } catch (actualError) {
+      expect(actualError).instanceOf(HTTP401);
+    }
+  });
+
+  it('throws HTTP401 when authorization header containing valid bearer token', async function () {
```
NickPhura: should this say: "invalid" bearer token?

api/src/security/auth-utils.test.ts
anissa-agahchen: wonder if I should put a comment explaining why?  otherwise, can we solve it?

api/src/security/auth-utils.test.ts
```diff
@@ -270,7 +278,161 @@ describe('authorize', function () {
...
+      });
+    } catch (actualError) {
+      expect(actualError).instanceOf(HTTP401);
+    }
+  });
+
+  //TODO:  figure out how to stub the callback parameter of the function
+
+  // import * as jsonwebtoken from 'jsonwebtoken';
+  // it('returns a validated token through verifyToken', async function () {
```
NickPhura: If you just want to test how the return value of the `verifyToken` function is handled when called by the authenticate function, then you can just stub out the entire `verifyToken` function, and make it return w/e you want. 

And then we can test the verifyToken function directly, in its own set of unit tests.

api/src/security/auth-utils.test.ts
NickPhura: I think this test is asserting happy path, so we dont expect it to fail.

api/src/security/auth-utils.test.ts
```diff
@@ -270,7 +280,135 @@ describe('authorize', function () {
...
+      await auth_utils.authenticate({
+        headers: {
+          // sample encoded json web token from jwt.io (with kid header)
+          authorization:
+            'Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InNvbWUgaWQifQ.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.rXYCCTRmi7EAHVRy1QqHvysMxLZaH4EsI226hrJLtgM'
+        }
+      });
+      expect.fail();
+    } catch (actualError) {
+      expect(actualError).instanceOf(HTTP401);
```
sdevalapurkar: why do we expect this test case to fail??

sdevalapurkar (approved):

PR #265, opened 2021-04-29
Get codes.ts from 0-100 coverage and also permit-no-sampling coverage to 76%

sdevalapurkar:
# Overview

- Get codes.ts from 0-100 coverage and also permit-no-sampling coverage to 76%

image

### DIFF
--- a/api/src/paths/permit-no-sampling.test.ts
+++ b/api/src/paths/permit-no-sampling.test.ts
@@ -1,0 +1,135 @@
+import chai, { expect } from 'chai';
+import { describe } from 'mocha';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+import * as permit_no_sampling from './permit-no-sampling';
+import * as db from '../database/db';
+
+chai.use(sinonChai);
+
+describe('permit-no-sampling', () => {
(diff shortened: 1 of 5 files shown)


api/src/paths/permit-no-sampling.test.ts
```diff
@@ -0,0 +1,135 @@
...
+import { describe } from 'mocha';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+import * as permit_no_sampling from './permit-no-sampling';
+import * as db from '../database/db';
+
+chai.use(sinonChai);
+
+describe('permit-no-sampling', () => {
+  const dbConnectionObj = {
```
NickPhura, line 11: No necessarily in this PR, but we should probably move some of these base common mocks into a folder somewhere, so we dont have to re-create these common mocks everywhere. Similar to the test-helpers in the frontend.  And then add thtat folder to the sonar ignore, etc.

sdevalapurkar: for sure will do it in a later pr when touching api tests next

NickPhura (approved):

PR #270, opened 2021-04-30
Update Db code table content

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1012

- Update Db code table content

## How Has This Been Tested?

- Locally


rstens (commented):
Does this require a full refresh of the database? In other words, a drop of the current one and a complete rebuild? If so, this cannot be simply merged, some work needs to happen first (the dropping of the database pod).

sdevalapurkar:
> Does this require a full refresh of the database? In other words, a drop of the current one and a complete rebuild? If so, this cannot be simply merged, some work needs to happen first (the dropping of the database pod).

I believe so yep :)

cgarrettjones (approved):

PR #278, opened 2021-05-04
BHBC-1023: Front end work for - Create Survey API and connect to Save and exit button

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1023

- Front end work for - Create Survey API and connect to Save and exit button

## How Has This Been Tested?

- Yet to be fully tested end to end with backend changes (when ready)


NickPhura:
Are the Proprietary Information fields supposed to toggle based on the `Is the data captured in this survey proprietary?` radio field?

I'm guessing the idea is that if they say `No` then we don't show them anything else, for that section.

Looks good though!

sdevalapurkar:
> Are the Proprietary Information fields supposed to toggle based on the `Is the data captured in this survey proprietary?` radio field?
> 
> 
> 
> I'm guessing the idea is that if they say `No` then we don't show them anything else, for that section.
> 
> 
> 
> Looks good though!

Yeah I wasn't really aware of that requirement but makes sense. I can definitely add it in 😃

anissa-agahchen:
> Are the Proprietary Information fields supposed to toggle based on the `Is the data captured in this survey proprietary?` radio field?
> 
> I'm guessing the idea is that if they say `No` then we don't show them anything else, for that section.
> 
> Looks good though!

There is also the requirement that if the category selection is a first_nations_category, then the proprietor field is populated from the first_nations table.  Not sure what the expected behaviour is if the category selected is not a first nation. This could definitely be in a future PR though.

anissa-agahchen (approved):

PR #282, opened 2021-05-05
Updates to proprietary data section of create survey

sdevalapurkar:
# Overview

- Make proprietary data section conditionally visible
- Make data category used proprietor_type dropdown code table values
- Conditionally base the proprietor_name based on the type (if first nations, then make dropdown)

### DIFF
--- a/api/src/paths/codes.ts
+++ b/api/src/paths/codes.ts
@@ -141,6 +141,20 @@ GET.apiDoc = {
                   }
                 }
               },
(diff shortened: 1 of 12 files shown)


anissa-agahchen (approved):

NickPhura (approved):
Looks good.  That hidden chunk of fields is now missing from the create survey snapshot.  But can be added later, either way.

PR #321, opened 2021-05-14
Some API tests

anissa-agahchen:


api/src/paths/project/{projectId}/attachments/list.test.ts
```diff
@@ -0,0 +1,139 @@
...
+      query: mockQuery
+    });
+
+    sinon.stub(project_attachments_queries, 'getProjectAttachmentsSQL').returns(SQL`something`);
+
+    const result = listAttachments.getAttachments();
+
+    await result(sampleReq, sampleRes as any, (null as unknown) as any);
+
+    expect(actualResult).to.not.be.null;
```
sdevalapurkar: should be able to assert the actual value of the result, not just that it isnt null

anissa-agahchen: Yeah, not super happy with this test.  I seem to not be able to see the id in the actualResult and the others are undefined.

anissa-agahchen: should be fixed now

sdevalapurkar (approved):
download

PR #322, opened 2021-05-14
Study species table, species on the survey (not project) and more major changes

cgarrettjones:
# Overview

## Links to jira tickets

- [BHBC-1082](https://quartech.atlassian.net/browse/BHBC-1082)
- [BHBC-1069](https://quartech.atlassian.net/browse/BHBC-1069)
- [BHBC-529](https://quartech.atlassian.net/browse/BHBC-529)

## This PR contains the following changes

- added table study species
- added table survey attachment
- added table survey funding source
- various data integrity checks
- Study species table, species on the survey (not project) and more major changes

## This PR contains the following types of changes

- [x] New feature (change which adds functionality)
- [x] Enhancement (improvements to existing functionality)
- [x] Misc cleanup / Refactoring / Documentation

## How Has This Been Tested?

- Locally

PR #324, opened 2021-05-14
BHBC-1010

cgarrettjones:
- added permit table
  - various data integrity checks

# Overview

## Links to jira tickets

- [BHBC-1010](https://quartech.atlassian.net/browse/BHBC-1010)

- removed old permit tables
- added table permit
- various data integrity contraints

- [ ] Version change

## Checklist

### General

### Code

- [x] My code uses types/interfaces to help describe values/parameters/etc, help ensure type safety, and improve readability

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

## How Has This Been Tested?

## Screenshots


.prettierignore
```diff
@@ -1,11 +1,11 @@
...
+**/node_modules
+
+dist
+coverage
+
+openshift
+sonar-runner
+uploads
+
+.vscode
```
NickPhura, line 11: Looks like this file got moved from `/api` to `/`, I assume accidentally.  Just needs to get moved back

NickPhura:
Just to confirm, @cgarrettjones does this PR contain all of the changes from #322?  If it does, then we can close 322 and just PR this one.

Edit: 

Actually, maybe we shouldn't, that way we can address the API/APP changes of the first PR on its own, and then worry about any changes from this one afterwards as its own piece of work.

NickPhura:
> 
> 
> Just to confirm, @cgarrettjones does this PR contain all of the changes from #322? If it does, then we can close 322 and just PR this one.

Actually, maybe we shouldn't, that way we can address the API/APP changes of the first PR on its own, and then worry about any changes from this one afterwards as its own piece of work.

database/src/migrations/release.0.16/biohub.sql
```diff
@@ -675,17 +593,19 @@ COMMENT ON TABLE management_action_type IS 'List of Management Actions.'
 ;
 
 -- 
--- TABLE: no_sample_permit 
+-- TABLE: permit 
 --
 
-CREATE TABLE no_sample_permit(
+CREATE TABLE permit(
```
NickPhura, line 599: It looks like there is no `isSamplingConducted` boolean in this new permit table.  
I think we still need it though, otherwise we won't be able to populate that part of the UI correctly.

database/src/migrations/release.0.16/biohub.sql
```diff
@@ -675,17 +593,19 @@ COMMENT ON TABLE management_action_type IS 'List of Management Actions.'
...
     id                           integer           GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+    p_id                         integer,
+    s_id                         integer,
```
NickPhura, line 602: I wonder if it would be better to handle survey permits like you did with survey funding, where its a separate table with a project permit id foreign key.

Mostly thinking about the case where a single project permit is assigned to multiple surveys.  Unless this never happens?

cgarrettjones: 1) a permit can be associated with one or zero projects
2) a permit can only be associated with one survey
3) a permits association with a survey implies that sampling was conducted

the model currently supports the above.

PR #326, opened 2021-05-18
bhbc1064 -> updates to date stuff based on robin's feedback

sdevalapurkar:
# Overview

bhbc1064 -> updates to date stuff based on robin's feedback

remove date-pick blocker; add survey help text for date fields and modify error message for date, so that users know what the project start and end dates are

### DIFF
--- a/app/src/components/fields/StartEndDateFields.tsx
+++ b/app/src/components/fields/StartEndDateFields.tsx
@@ -1,15 +1,14 @@
 import Grid from '@material-ui/core/Grid';
 import TextField from '@material-ui/core/TextField';
-import { DATE_FORMAT, DATE_LIMIT } from 'constants/dateFormats';
+import { DATE_LIMIT } from 'constants/dateFormats';
 import React from 'react';
-import { getFormattedDate } from 'utils/Utils';
 
 interface IStartEndDateFieldsProps {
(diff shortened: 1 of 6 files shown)


NickPhura (approved):
Looks good.

I didn't actually run it, but the changes are simple enough that I can't imagine there being any issues given whatever testing you've already done.

One question: 

The wording of the error message is of the form "Survey start date cannot be before...". I thought I remembered Jeremy saying it should be shorter when possible, something like "Cannot be before ..." or "Cannot precede...".

Not necessarily a super critical issue, as we probably need to do a pass over all of the messaging in the app to unify it, at some point.

sdevalapurkar:
> Looks good.
> 
> I didn't actually run it, but the changes are simple enough that I can't imagine there being any issues given whatever testing you've already done.
> 
> One question:
> 
> The wording of the error message is of the form "Survey start date cannot be before...". I thought I remembered Jeremy saying it should be shorter when possible, something like "Cannot be before ..." or "Cannot precede...".
> 
> Not necessarily a super critical issue, as we probably need to do a pass over all of the messaging in the app to unify it, at some point.

thanks!
yeah not worrying bout the wording details, jer can update later and we can revisit :)

PR #331, opened 2021-05-25
Store artifacts associated to a survey - File upload

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-529

- Store artifacts associated to a survey - File upload
- View file from fe
- delete from fe

## How Has This Been Tested?

- Locally, unit tests

## Screenshots

image


api/src/paths/project/{projectId}/survey/{surveyId}/attachments/{attachmentId}/getSignedUrl.ts
```diff
@@ -0,0 +1,99 @@
...
+      );
+
+      await connection.commit();
+
+      const s3Key = result && result.rows.length && result.rows[0]?.key;
+
+      const s3SignedUrl = await getS3SignedURL(s3Key);
+
+      if (!s3SignedUrl) {
+        return null;
```
NickPhura: What does returning null yield (200 status with no content)?

Might be better to return an actual status here, even if we still want to return `null` as the content.

Same for `delete.ts`

sdevalapurkar: sounds good

NickPhura (approved):

PR #340, opened 2021-06-02
Add tests for API

sdevalapurkar:
# Overview

Add tests for API

### DIFF
--- a/api/src/paths/project/{projectId}/survey/{surveyId}/update.ts
+++ b/api/src/paths/project/{projectId}/survey/{surveyId}/update.ts
@@ -273,14 +273,7 @@ export const getSurveyProprietorData = async (
 
   const response = await connection.query(sqlStatement.text, sqlStatement.values);
 
-  const result =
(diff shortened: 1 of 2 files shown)


api/src/paths/project/{projectId}/survey/{surveyId}/update.ts
```diff
@@ -273,7 +273,14 @@ export const getSurveyProprietorData = async (
...
 
-  return (response && response.rows && response.rows[0] && new GetSurveyProprietorData(response.rows[0])) || null;
+  const result =
+    (response && response.rows && response.rows[0] && new GetSurveyProprietorData(response.rows[0])) || null;
+
+  if (!result) {
+    throw new HTTP400('Failed to get project survey proprietor data');
+  }
+
+  return result;
```
NickPhura: I think you were on vacation, but this was changed to intentionally return null if there is no proprietor data.  So we don't want to throw an error here.

NickPhura (dismissed):

NickPhura (approved):

PR #365, opened 2021-06-15
Block observation get for update basic endpoint

sdevalapurkar:
# Overview

## Links to jira tickets

- Start of https://quartech.atlassian.net/browse/BHBC-1152

- Block observation get for update basic endpoint

## How Has This Been Tested?

- Needs to be properly wired up and tested/used


anissa-agahchen:
/observations/1/block is broken

Screen Shot 2021-06-17 at 8 17 59 AM

app/src/features/projects/ProjectsRouter.tsx
```diff
@@ -87,6 +87,13 @@ const ProjectsRouter: React.FC<IProjectsRouterProps> = (props) => {
         component={BlockObservationPage}
         componentProps={props}
       />
+      <PrivateRoute
+        exact
+        layout={ProjectsLayout}
+        path="/projects/:id?/surveys/:survey_id?/observations/:observation_id?/block"
```
anissa-agahchen, line 93: should this say?
path="/projects/:id?/surveys/:survey_id?/observations/:observation_id?/edit" 
or
path="/projects/:id?/surveys/:survey_id?/observations/:observation_id?/details"

sdevalapurkar: lets discuss this on call -> no this should be just block cause it is specific for block observations
I have never seen the error above so if you could show me how you reproduced it that would be awesome

anissa-agahchen: All good!   I needed to cleanout the database.

anissa-agahchen (dismissed):
Looks great!

anissa-agahchen (approved):

app/src/interfaces/useObservationApi.interface.ts
```diff
@@ -9,3 +11,25 @@ export interface IGetBlocksListResponse {
...
+
+/**
+ * Create observation post object.
+ *
+ * @export
+ * @interface ICreateObservationRequest
+ */
+export interface ICreateObservationRequest {
+  metaData: IBlockObservationForm;
+  tableData: any[][];
```
NickPhura, line 23: Heads up, there is a slight discrepancy between this definition, and the one in [Anissa's PR](https://github.com/bcgov/biohubbc/pull/367/files#diff-7978d58e06195e444e4da2d9adeda22c5478bdd772e82273c53b46f5babb5822R21).

I think we want to go with the way Anissa has it currently, where tableData is an object, with a data property that is a double array.  That way we can accommodate additional table related properties if we need to.

PR #371, opened 2021-06-17
Update Block Observation!!!

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1152

- update block observations (put)

## How Has This Been Tested?

- Locally


api/src/paths/project/{projectId}/survey/{surveyId}/observations/{observationId}/update.ts
```diff
@@ -86,6 +89,98 @@ GET.apiDoc = {
...
+    content: {
+      'application/json': {
+        schema: {
+          title: 'Observation request object',
+          type: 'object',
+          required: ['observation_type'],
+          properties: {
+            observation_type: {
+              title: 'type of observation (eg block)',
+              type: 'string'
```
NickPhura: nit: we could define this as an eum

app/src/features/observations/BlockObservationPage.tsx
```diff
@@ -54,16 +59,20 @@ const BlockObservationPage = () => {
 
   // Ability to bypass showing the 'Are you sure you want to cancel' dialog
   const [enableCancelCheck, setEnableCancelCheck] = useState(true);
-  const [observationWithDetails, setObservationWithDetails] = useState<IBlockObservationForm>(
-    BlockObservationInitialValues
-  );
   const [tableData, setTableData] = useState<any[][]>([[, , , , , , , , , , , , , ,]]);
```
NickPhura, line 62: Not necessarily something to action now: we'll need to devise all of this table stuff, in general, to handle different headers/schemas/data/etc.  In this case what caught my eye was the hardcoded default that presumably has the right number of cells of the moose, but wouldn't be right for other kinds of data.  Possibly we would need to handle a null default, or have the default based on the headers config, or something along those lines.

sdevalapurkar: for sure, good point :) a lot of the stuff is done specifically to handle the moose use case for now 
but a good thing to keep in mind moving forward 💯

app/src/features/observations/BlockObservationPage.tsx
```diff
@@ -189,7 +197,7 @@ const BlockObservationPage = () => {
     return <CircularProgress className="pageProgress" size={40} />;
   }
 
-  const handleSaveAndExit = async () => {
+  const handleSaveAndExit = async (action: string) => {
```
NickPhura: Do we need this `action` param? Or lower down when we make the api call, can't we determine which to call based on the existence of the observationId?

If that isn't possible for some reason, then we should at least make this string an enum so we only get passed known options.

sdevalapurkar: no we definitely can :) good one

anissa-agahchen (approved):
Looks good!

PR #377, opened 2021-06-22
Fix weird CSS issues in Dev

NickPhura:
Update the makeStyles import to be consistent with the one used in the rest of the app.
- I believe either version works, as long as they are always the same.  But, I think the one from `/core/` is the default one (and so the one we have been using in 90% of cases).

Added some bonus logic to strip trailing `/` from routes, so that the route handling doesn't have to worry about bother versions of a url (ending in a slash, and not ending in a slash).


app/src/AppRouter.tsx
```diff
@@ -9,17 +9,20 @@ import AccessRequestPage from 'pages/access/AccessRequestPage';
...
 const AppRouter: React.FC = (props: any) => {
+  const location = useLocation();
+
   const getTitle = (page: string) => {
     return `BioHub - ${page}`;
   };
 
   return (
     <Switch>
+      <Redirect from="/:url*(/+)" to={location.pathname.slice(0, -1)} />
```
sdevalapurkar, line 25: what is this for?

sdevalapurkar (commented):
awesome work

app/src/AppRouter.tsx
NickPhura: Oh oops, I thought I had updated the description (might have just been the commit message).

This is just some code that strips out trailing `/` in routes.
The routers have different behaviour if the url does or does not end in a slash, so this removes that concern.

Example:

You have a link that says `<Link to="newpage"/>`

The current URL in the browser is: `www.host.come/projects/`
That link will take you to: `www.host.come/currentpage/newpage`

The current URL in the browser is: `www.host.come/currentpage` (no trailing slash)
That link will take you to: `www.host.come/newpage`

Etc.

sdevalapurkar (approved):

PR #378, opened 2021-06-22
Delete project functionality!!!

sdevalapurkar:
# Overview

- Delete project functionality!!! 🗑️
- Also wipe attachments associated to project and survey from s3
- Deal with observations and occurrences later (block observation table does get wiped, but detailed deletion of records under survey level will be dealt with later on)

## How Has This Been Tested?

- Locally

### DIFF
--- a/api/src/paths/project/{projectId}/delete.ts
+++ b/api/src/paths/project/{projectId}/delete.ts
@@ -1,0 +1,143 @@
+import { RequestHandler } from 'express';
+import { Operation } from 'express-openapi';
+import { SYSTEM_ROLE } from '../../../constants/roles';
+import { getDBConnection } from '../../../database/db';
+import { HTTP400 } from '../../../errors/CustomError';
+import { getProjectAttachmentsSQL } from '../../../queries/project/project-attachments-queries';
+import { getLogger } from '../../../utils/logger';
+import { getSurveyIdsSQL } from '../../../queries/survey/survey-view-queries';
+import { getSurveyAttachmentS3Keys } from './survey/{surveyId}/delete';
+import { deleteProjectSQL } from '../../../queries/project/project-delete-queries';
+import { deleteFileFromS3 } from '../../../utils/file-utils';
+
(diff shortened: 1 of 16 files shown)


api/src/paths/project/{projectId}/delete.ts
```diff
@@ -0,0 +1,143 @@
...
+      if (!getSurveyIdsResult || !getSurveyIdsResult.rows) {
+        throw new HTTP400('Failed to get survey ids associated to project');
+      }
+
+      const surveyAttachmentS3Keys: string[] = Array.prototype.concat.apply(
+        [],
+        await Promise.all(
+          getSurveyIdsResult.rows.map((survey: any) => getSurveyAttachmentS3Keys(survey.id, connection))
+        )
+      );
```
NickPhura, line 101: Can this not be simplified to something like:

```typescript
 const surveyAttachmentS3Keys: string[] = await Promise.all(
    getSurveyIdsResult.rows.map((survey: any) => getSurveyAttachmentS3Keys(survey.id, connection))
  );
```

sdevalapurkar: it complains that they are incompatible types:
```
Type 'any[][]' is not assignable to type 'string[]'.
Type 'any[]' is not assignable to type 'string'.
```

sdevalapurkar: it returns nested array that I am just flattening

database/src/migrations/release.0.18/api_delete_project.sql
```diff
@@ -20,6 +20,7 @@ begin
   delete from survey_proprietor where s_id in (select id from survey where p_id = __project_id);
   delete from survey_attachment where s_id in (select id from survey where p_id = __project_id);
   delete from study_species where s_id in (select id from survey where p_id = __project_id);
+  delete from block_observation where s_id in (select id from survey where p_id = __project_id);
```
anissa-agahchen, line 23: nice!

anissa-agahchen (approved):
Looks good

PR #392, opened 2021-06-29
Publish a project

anissa-agahchen:
# Overview

- Project is published based on a front-end toggle to publish/unpublish a project
- When a project a published, the publish_timestamp is inserted into the project table
- When a project is unpublished, the timestamp is replaced with null

- [ ] Version change

## Checklist

### General

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

## How Has This Been Tested?

## Screenshots


api/src/paths/project/{projectId}/publish.ts
```diff
@@ -0,0 +1,130 @@
...
+    content: {
+      'application/json': {
+        schema: {
+          title: 'Publish request object',
+          type: 'object',
+          required: ['publish'],
+          properties: {
+            publish: {
+              title: 'publish?',
+              type: 'boolean'
```
sdevalapurkar, line 44: is this accurate? I thought you are passing the `publish_timestamp`

api/src/paths/project/{projectId}/publish.ts
```diff
@@ -0,0 +1,130 @@
...
+      $ref: '#/components/responses/500'
+    },
+    default: {
+      $ref: '#/components/responses/default'
+    }
+  }
+};
+
+/**
+ * Update a project.
```
sdevalapurkar, line 82:
```suggestion
 * Publish a project.
```

api/src/paths/project/{projectId}/publish.ts
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * @returns {RequestHandler}
+ */
+function publishProject(): RequestHandler {
+  return async (req, res) => {
+    const connection = getDBConnection(req['keycloak_token']);
+
+    try {
+      const projectId = Number(req.params?.projectId);
```
sdevalapurkar:
```suggestion
      const projectId = Number(req.params.projectId);
```

api/src/paths/project/{projectId}/publish.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      const publish: boolean = req.body?.publish;
```
sdevalapurkar:
```suggestion
      const publish: boolean = req.body.publish;
```

api/src/paths/project/{projectId}/publish.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      if (!projectId) {
+        throw new HTTP400('Missing required path parameter: projectId');
+      }
+
+      if (!projectId) {
```
sdevalapurkar: this is duplicated, I think you meant to check `req.body`

api/src/paths/project/{projectId}/publish.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      if (!response || !result) {
+        throw new HTTP500('Failed to update project publish status');
+      }
+
+      await connection.commit();
+
+      return res.status(200).send(result);
+    } catch (error) {
+      defaultLog.debug({ label: 'updateProject', message: 'error', error });
```
sdevalapurkar:
```suggestion
      defaultLog.debug({ label: 'publishProject', message: 'error', error });
```

api/src/queries/project/project-update-queries.ts
```diff
@@ -391,3 +391,35 @@ export const putProjectFundingSourceSQL = (
 
   return sqlStatement;
 };
+
+/**
+ * SQL query to update the publish status of a project.
+ *
+ * @param {number} projectId
+ * @param {boolean} publish
```
sdevalapurkar, line 399: can we change it to pass the timestamp to this so that we can just set it

app/src/features/projects/view/ProjectPage.tsx
```diff
@@ -111,6 +118,29 @@ const ProjectPage: React.FC = () => {
...
 
+  const publishProject = async (publish: boolean) => {
+    if (!projectWithDetails) {
+      return;
+    }
+
+    try {
+      const response = await biohubApi.project.publishProject(
+        projectWithDetails.id,
+        !projectWithDetails.project.publish_date
```
sdevalapurkar: if we just pass the `date` here, itll prob make things slightly simpler

app/src/features/projects/view/ProjectPage.tsx
```diff
@@ -177,15 +207,48 @@ const ProjectPage: React.FC = () => {
...
+                      variant="outlined"
+                      color="primary"
+                      data-testid="publish-project-button"
+                      startIcon={<Icon path={mdiToggleSwitchOffOutline} size={1} />}
+                      onClick={async () => {
+                        publishProject(true);
+                      }}>
+                      Publish Project
+                    </Button>
+                  </Box>
```
sdevalapurkar: this can be combined into a single section with just checks to determine which text needs to be displayed and which icons need to be shown

app/src/hooks/api/useProjectApi.ts
```diff
@@ -4,14 +4,15 @@ import {
...
   IGetProjectsListResponse,
-  UPDATE_GET_ENTITIES,
-  IGetProjectForUpdateResponse,
+  IProjectAdvancedFilterRequest,
   IUpdateProjectRequest,
-  IGetProjectAttachmentsResponse,
-  IProjectAdvancedFilterRequest
+  UPDATE_GET_ENTITIES
 } from 'interfaces/useProjectApi.interface';
+//import moment from 'moment';
```
sdevalapurkar: remove if not used

api/src/paths/project/{projectId}/publish.ts
anissa-agahchen: Yeah I changed it to boolean .. because when we unpublish, we need to send either an empty string or null.  It's just as easy for the DB to create the timestamp ... so I went with that.

app/src/features/projects/view/ProjectPage.tsx
sdevalapurkar:
```
<Box ml={4} mt={4} mb={4}>
    <Button
        variant="outlined"
        color="primary"
        data-testid="publish-project-button"
        startIcon={<Icon path={projectWithDetails.project.publish_date ? mdiToggleSwitch : mdiToggleSwitchOffOutline} 
        size={1} />}
        onClick={async () => {
              if (projectWithDetails.project.publish_date) {
                      await publishProject(false);
              } else {
                      await publishProject(true);
               }
        }}>
        {projectWithDetails.project.publish_date ? 'Unpublish Project' : 'Publish Project'}
    </Button>
</Box>
```

anissa-agahchen: That worked well !  Thank you.

api/src/paths/project/{projectId}/publish.ts
NickPhura: We definitely want the timestamp to be created by the server.  So the endpoint can either take a boolean (and support both publish/unpublish), or we can have two separate endpoints (one for publish, one for unpublish), and the user doesnt have to provide any data (other than the project ID, but that is already part of the route).

api/src/queries/project/project-update-queries.ts
```diff
@@ -391,3 +391,35 @@ export const putProjectFundingSourceSQL = (
...
+  defaultLog.debug({ label: 'updateProjectPublishStatusSQL', message: 'params', projectId, publish });
+
+  if (!projectId) {
+    return null;
+  }
+
+  const sqlStatement: SQLStatement = SQL`
+      UPDATE project
+      SET publish_timestamp = ${publish ? 'now()' : null}
+      WHERE id = ${projectId}
```
NickPhura: If the project is already published, we don't want to update the row with a new publish_timestamp.
Can we add an extra condition here to check for "and publish_timetamp = null" (or w/e the correct syntax is).

NickPhura: Not sure what the return value is in this case though, might impact the response checks in the endpoint code `publish.ts`?

app/src/features/projects/view/ProjectPage.tsx
```diff
@@ -111,6 +118,29 @@ const ProjectPage: React.FC = () => {
     onYes: () => dialogContext.setYesNoDialog({ open: false })
   };
 
+  const publishProject = async (publish: boolean) => {
```
NickPhura, line 121: It looks like the publish param is never actually used, as the boolean is being determine based on the publish_date (line 129). I think we should use the parameter though.  That way the api call being made is more deterministic, and anyone can add a publish or unpublish button to the page, without worrying about the real state of the records publish_date.

api/src/paths/project/{projectId}/publish.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      const result = publish.publishProject();
+
+      await result(
+        { ...sampleReq, body: (null as unknown) as any },
+        (null as unknown) as any,
+        (null as unknown) as any
+      );
+      expect.fail();
+    } catch (actualError) {
+      expect(actualError.status).to.equal(400);
```
NickPhura, line 110: For all of these error tests, we should add an expect for `actualError.message`.  Otherwise its possible that the 400 being caught isn't the one we were testing for (as the code has multiple checks that all throw a 400).

NickPhura (approved):

PR #393, opened 2021-06-29
Update search feature map popup with link and project view with statuses

sdevalapurkar:
# Overview

- Update search feature map popup with link and project view with statuses
- Small bug fix with relation to project completion status calculation (for projects with no end date)

## How Has This Been Tested?

- Local

## Screenshots

image

image


api/src/models/project-view.ts
```diff
@@ -26,6 +28,9 @@ export class GetProjectData {
     this.start_date = projectData?.start_date || '';
     this.end_date = projectData?.end_date || '';
     this.comments = projectData?.comments || '';
+    this.completion_status =
+      (projectData && projectData.end_date && moment(projectData.end_date).isBefore(new Date()) && 'Completed') ||
```
NickPhura: Do we need to account for when the end_date is today? Maybe a question for the business: do they consider end_date = July 1st to mean 11:59:59 pm on July 1st, or 12:00:00 am on July 1st.  I'd assume the former?

So for the former, something like:

```typescript
moment(projectData.end_date).endOf('day').isBefore(moment())
```

In this case, end_date = July 1st @ 11:59:59 pm only becomes old when the current day ticks over to July 2nd @ 12:00:00 am.

NickPhura: Oh, do we need to account for when the end_date is today? Maybe a question for the business: do they consider end_date = July 1st to mean 11:59pm on July 1st, or 12:01am on July 1st.  I'd assume the former?

In any case, moment has a `isSameOrBefore` function, if needed.

api/src/paths/projects.ts
```diff
@@ -187,7 +187,7 @@ export function _extractProjects(rows: any[]): any[] {
       end_date: row.end_date,
       coordinator_agency: row.coordinator_agency_name,
       publish_status: row.publish_timestamp ? 'Published' : 'Unpublished',
-      completion_status: moment(row.end_date).isSameOrAfter(new Date()) ? 'Active' : 'Completed',
+      completion_status: (row.end_date && moment(row.end_date).isBefore(new Date()) && 'Completed') || 'Active',
```
NickPhura: Same question: https://github.com/bcgov/biohubbc/pull/393/files#r662533152

app/src/components/map/SearchFeaturePopup.tsx
```diff
@@ -1,24 +1,20 @@
...
-  const { featureData } = props;
-
-  const popupItems = [];
-  for (const property in featureData) {
-    if (property !== 'geometry') {
-      popupItems.push(`${property}: ${featureData[property]}`);
-    }
-  }
+export const SearchFeaturePopup: React.FC<{ featureData: any; history: any }> = (props) => {
+  const { featureData, history } = props;
```
NickPhura: Why does history have to be passed in as a prop? 
Can we not just have `const history = useHistory();` here directly?

anissa-agahchen:
@sdevalapurkar I tested your PR locally, and pulled in the changes from dev, and dealt with the merge conflicts.  Aside from addressing Nick's feedback, the PR looks good.

api/src/models/project-view.ts
sdevalapurkar: that makes sense :) thanks

NickPhura (approved):

PR #396, opened 2021-07-01
BHBC-1140: DWC ZIP and EXCEL parsing + validation

NickPhura:
# Overview

## Links to jira tickets

https://quartech.atlassian.net/browse/BHBC-1140

Migrate code from e2e branch. 
Enhance to support not only dwc zip files, but also excel workbooks.
General updates/improvements.
Add unit tests.

- [ ] Version change

## Checklist

### General

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

## How Has This Been Tested?

## Screenshots


api/src/paths/dwc/validate.ts
```diff
@@ -0,0 +1,223 @@
...
+function prepDWCArchive(): RequestHandler {
+  return async (req, res, next) => {
+    defaultLog.debug({ label: 'prepDWCArchive', message: 'files.length', files: req?.files?.length });
+
+    if (!req.files || !req.files.length) {
+      // no media objects included
+      throw new HTTP400('Missing upload data');
+    }
+
+    if (req.files.length !== 1) {
```
sdevalapurkar, line 173: question:

is the reason we accept an array of files instead of just a single file because we eventually wanna support multiple files at once but for now we are only doing the case for one? cause if this is not the case, just wondering if it can be a single file accepted so we don't need a check like this

api/src/paths/project/{projectId}/survey/{surveyId}/observations/upload.ts
```diff
@@ -0,0 +1,326 @@
...
+    }
+  };
+}
+
+/**
+ * Validate a Darwin Core csv file.
+ *
+ * @returns {RequestHandler}
+ */
+function validateDWCArchive(): RequestHandler {
```
sdevalapurkar, line 226: question:

the prep function and this function in this file seem very similar to the one inside `validate.ts`
can they not be reused?

api/src/paths/spi/validate.ts
```diff
@@ -0,0 +1,257 @@
...
+import { getLogger } from '../../utils/logger';
+import { ICsvState, XLSXCSV } from '../../utils/media/csv/csv-file';
+import { isSPITemplateValid } from '../../utils/media/csv/spi/spi-validator';
+import { IMediaState } from '../../utils/media/media-file';
+import { parseUnknownMedia } from '../../utils/media/media-utils';
+import { logRequest } from '../../utils/path-utils';
+
+const defaultLog = getLogger('paths/project');
+
+export const POST: Operation = [logRequest('paths/spi/transform', 'POST'), prepSPITemplate(), validateSPITemplate()];
```
sdevalapurkar:
```suggestion
export const POST: Operation = [logRequest('paths/spi/validate', 'POST'), prepSPITemplate(), validateSPITemplate()];
```

sdevalapurkar (commented):
havent looked at the entire pr, it is quite large 😸 
a few initial questions

api/src/paths/project/{projectId}/survey/{surveyId}/observations/upload.ts
NickPhura: Its pretty similar yeah, technically the one in upload also checks for a survey Id, which is different.  There is probably some opportunity to make some more re-usable functions, but probably not a concern just yet. Dont want to optimize it too much this early on

api/src/paths/dwc/validate.ts
NickPhura: Think we talked about this.  The schema was out of date and should have indicated a single file.  That being said, the schema doesnt actually prevent someone sending multiple files, so this check is still needed for now

api/src/paths/project/{projectId}/survey/{surveyId}/observations/upload.ts
sdevalapurkar: ok that's fair

api/src/queries/occurrence/occurrence-create-queries.ts
```diff
@@ -0,0 +1,107 @@
+import { SQL, SQLStatement } from 'sql-template-strings';
+import { PostOccurrence } from '../../models/occurrence-create';
+import { getLogger } from '../../utils/logger';
+import { parseUTMString } from '../../utils/spatial-utils';
+
+const defaultLog = getLogger('queries/occurrence/occurrence-create-queries');
+
+export const postOccurrenceSubmissionSQL = (surveyId: number): SQLStatement | null => {
```
sdevalapurkar, line 8: let's add tests for these two functions in the queries file, cause it is low hanging test coverage that can be added in now

sdevalapurkar (commented):
overall looking pretty good to me :)

just unit tests that need to be added and I don't have as much knowledge about this code so would be great if you could add them - not all in this pr obv but soon while its fresh in memory

sdevalapurkar (approved):
lgtm, but would be great to have a separate pr soon with some test coverage :)

PR #408, opened 2021-07-13
BHBC 1081: Implement funding to the survey metadata

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1081

- Implement funding to the survey metadata

## How Has This Been Tested?

- Locally and unit tests

## Screenshots

image

image

image


api/src/paths/project/{projectId}/survey/funding-sources/list.ts
```diff
@@ -0,0 +1,120 @@
...
+        null;
+
+      if (!getSurveyFundingSourcesData) {
+        return res.status(200).json(null);
+      }
+
+      return res.status(200).json(getSurveyFundingSourcesData.fundingSources);
+    } catch (error) {
+      defaultLog.debug({ label: 'getSurveyFundingSources', message: 'error', error });
+      throw error;
```
NickPhura: Does this need a connection.rollback call?

api/src/paths/project/{projectId}/survey/create.ts
```diff
@@ -260,3 +272,23 @@ export const insertSurveyPermit = async (
...
+    throw new HTTP400('Failed to build SQL statement for insertSurveyFundingSource');
+  }
+
+  const response = await connection.query(sqlStatement.text, sqlStatement.values);
+
+  if (!response) {
+    throw new HTTP400('Failed to insert survey funding source data');
+  }
+
+  return true;
```
NickPhura: I know we have this in a few spots to help with testing, but we shouldn't need to return something here if the actual function doesn't need to.

Rather than assert it returns true in the tests, we can just be satisfied if no errors are thrown.  
Or if you want to go one step further, assert that the mock connection query object was called 1 time (and optionally assert what values it was called with).

app/src/features/projects/view/ProjectPage.tsx
```diff
@@ -294,16 +294,18 @@ const ProjectPage: React.FC = () => {
...
-                    variant="outlined"
-                    color="primary"
-                    className={classes.actionButton}
-                    data-testid="delete-project-button"
-                    startIcon={<Icon path={mdiTrashCanOutline} size={1} />}
-                    onClick={showDeleteProjectDialog}
-                    disabled={disableDeleteProjectButton}>
-                    Delete Project
-                  </Button>
+                  <>
```
NickPhura, line 297: Are these fragment tags needed?

app/src/features/surveys/components/GeneralInformationForm.test.tsx
```diff
@@ -46,6 +48,24 @@ describe('General Information Form', () => {
...
+            ].map((item) => {
+              return {
+                value: item.pfsId,
+                label: `${item.agencyName} | ${getFormattedAmount(item.amount)} | ${getFormattedDateRangeString(
+                  DATE_FORMAT.ShortMediumDateFormat,
+                  item.startDate,
+                  item.endDate
+                )}`
+              };
+            })}
```
NickPhura: Not a huge concern, but since this is just a test, any reason for doing this mapping here, rather than just supplying the final array of key/values?

app/src/features/projects/view/ProjectPage.tsx
sdevalapurkar: yeah, cause the jest tests were giving a console error when the tooltip was given a potentially disabled button as a child

NickPhura (dismissed):
A few tiny comments, but looks good.

Gave it a quite test in openshift (adding/editing surveys with/without funding sources, etc).

app/src/features/projects/view/ProjectPage.tsx
NickPhura: Huh, interesting

api/src/paths/project/{projectId}/survey/create.ts
sdevalapurkar: ok sounds good

app/src/features/surveys/components/GeneralInformationForm.test.tsx
sdevalapurkar: mainly just so that the endpoint is more generic and doesnt have to return a weird string with a super fe specific formatted string

NickPhura (approved):

PR #424, opened 2021-07-21
Refactor filesize helper, improved template observations page, and added tests

anissa-agahchen:
Improved the Template Observations Page, and put in some tests

### DIFF
--- a/app/src/components/template/TemplateObservationList.tsx
+++ b/app/src/components/template/TemplateObservationList.tsx
@@ -10,7 +10,7 @@ import TablePagination from '@material-ui/core/TablePagination';
 import TableRow from '@material-ui/core/TableRow';
 import { DATE_FORMAT } from 'constants/dateTimeFormats';
 import { useBiohubApi } from 'hooks/useBioHubApi';
-import { IGetProjectAttachment } from 'interfaces/useProjectApi.interface';
+import { IGetTemplateObservations } from 'interfaces/useProjectApi.interface';
 import React, { useState } from 'react';
 import { handleChangePage, handleChangeRowsPerPage } from 'utils/tablePaginationUtils';
 import { getFormattedDate, getFormattedFileSize } from 'utils/Utils';
@@ -30,8 +30,8 @@ const useStyles = makeStyles({
 export interface ITemplateObservationsListProps {
   projectId: number;
   surveyId?: number;
-  attachmentsList: IGetProjectAttachment[];
-  getAttachments: (forceFetch: boolean) => void;
+  templateObservationsList: IGetTemplateObservations[];
+  getTemplateObservations: (forceFetch: boolean) => void;
 }
 
 const TemplateObservationsList: React.FC<ITemplateObservationsListProps> = (props) => {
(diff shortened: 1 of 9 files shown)


app/src/utils/Utils.test.ts
```diff
@@ -193,3 +194,25 @@ describe('getLogOutUrl', () => {
     );
   });
 });
+
+describe('getFormattedFileSize', () => {
+  it('returns `NaN GB` if no file size exists', async () => {
```
sdevalapurkar:
```suggestion
  it('returns `0 KB` if no file size exists', async () => {
```

app/src/utils/Utils.ts
```diff
@@ -118,6 +118,10 @@ export const getLogOutUrl = (config: IConfig): string | undefined => {
 };
 
 export const getFormattedFileSize = (fileSize: number) => {
+  if (!fileSize) {
+    return `0 KB`;
```
NickPhura: mega nit: can just use regular quotes here:
```suggestion
    return '0 KB';
```

app/src/components/template/TemplateObservationList.tsx
```diff
@@ -77,33 +77,39 @@ const TemplateObservationsList: React.FC<ITemplateObservationsListProps> = (prop
...
+                          variant="body2"
+                          onClick={() => viewFileContents(row)}>
+                          {row.fileName}
+                        </Link>
+                      </TableCell>
+                      <TableCell>{getFormattedDate(DATE_FORMAT.ShortMediumDateTimeFormat, row.lastModified)}</TableCell>
+                      <TableCell>{getFormattedFileSize(row.size)}</TableCell>
+                    </TableRow>
+                  ))}
+              {!props.templateObservationsList.length && (
```
sdevalapurkar: I think this may print out 0 onto the screen...
might need to do:
```
{props.templateObservationsList.length === 0 && (
```

sdevalapurkar (commented):
there seem to be some missing dependencies in the useeffect callback array in `TemplateObservationPage.tsx` maybe could you test adding them in and seeing if it actually breaks stuff or not?

also, you need to add a commit so the pr checks run (I guess the pr was converted from draft to real pr so the pr deployment hasnt run)

app/src/components/template/TemplateObservationList.tsx
sdevalapurkar: if I am mistaken and it doesnt, then this is all good 😄

NickPhura: ya thats a funny issue.  Not quite sure when it happens, but Im going to take a guess and say it wont happen in this case  Mainly Im thinking that the `!` will turn it into a boolean, and so wont print?  Unless it prints 'false' or something haha.  Curious to know what it does tho

NickPhura: Looked in openshift, works as expected, no issues

anissa-agahchen: yeah, it works as expected, but I changed it to `{props.templateObservationsList.length === 0 && (` anyway to avoid issues down the line.

app/src/features/observations/BlockObservationPage.tsx
```diff
@@ -126,7 +126,7 @@ const BlockObservationPage = () => {
       revision_count: observationWithDetailsResponse.revision_count
     });
     setTableData(observationWithDetailsResponse.data.tableData.data);
-  }, [biohubApi.observation, urlParams]);
+  }, [biohubApi.observation, observationId, projectId, surveyId]);
```
sdevalapurkar: are we sure this doesnt cause unexpected rerenders?

app/src/features/projects/create/CreateProjectPage.tsx
```diff
@@ -342,7 +342,7 @@ const CreateProjectPage: React.FC = () => {
         isTouched: false
       }
     ]);
-  }, [codes, stepForms, initialProjectFieldData, hasLoadedDraftData]);
+  }, [codes, stepForms, initialProjectFieldData, hasLoadedDraftData, nonSamplingPermits]);
```
sdevalapurkar, line 345: same question here

app/src/features/observations/BlockObservationPage.tsx
anissa-agahchen: added urlparams back, didn't notice issues with the others but will keep an eye.

sdevalapurkar (approved):
if renders is not an issue and all tested, then lgtm

PR #433, opened 2021-07-27
Inferring region layer info and displaying read-only

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1258
- https://quartech.atlassian.net/browse/BHBC-1260

- Implement UI and API changes to support project and survey boundaries
- DB changes for project and surveys to link boundaries to parks, regions, WMU

- Pending to fix UI and make cleaner

## How Has This Been Tested?

- Locally

## Screenshots

image


api/src/queries/project/project-delete-queries.ts
```diff
@@ -3,40 +3,6 @@ import { getLogger } from '../../utils/logger';
...
-    label: 'deleteRegionsSQL',
-    message: 'params',
-    projectId
-  });
-
-  if (!projectId) {
-    return null;
-  }
-
-  const sqlStatement: SQLStatement = SQL`
```
cgarrettjones, line 23: just to confirm, the intent is the delete **all** regions associated with a project? if the intent is to only delete a particular region associated with a project then use project_region_id to identify the record to delete.

database/src/migrations/20210726170005_remove_project_region.ts
```diff
@@ -0,0 +1,72 @@
+import * as Knex from 'knex';
```
cgarrettjones: sorry, not sure what this migration file is for... is this not part of the base install?

cgarrettjones (dismissed):
lots of work there, nice job!

anissa-agahchen (commented):
For the database changes, use migrations instead of changing the release bundle, because the release bundle is a baseline, and should reflect the latest db release that Charlie put up.

api/src/queries/project/project-delete-queries.ts
cgarrettjones: i guess we should get in the habit of checking users association with the data and their roles as well?

database/src/migrations/20210726170005_remove_project_region.ts
NickPhura: It looks like Shreyas edited the existing migrations to remove all of the foreign key references to the region table, and then dropped region table itself in this new migration.

NickPhura: But, we could move this change into the existing migration, if you prefer.  So we just edit the main SQL file to not create the region table at all, and then this new migration file could be removed. Normally we'd prefer doing new migrations, but as the changes are in both places, we could unify them in this case.

api/src/queries/project/project-delete-queries.ts
NickPhura: This particular change is just showing that shreyas has deleted this whole file.  Since we plan to infer the region based on the boundary, we dont need to store region information anymore, and so the region table can be removed and all associated code, which is a large portion of the changes in this PR.

anissa-agahchen (dismissed):

anissa-agahchen:
In the edit location/project boundary, editing the region doesn't work.

Meaning:  I cannot delete/add a region after the project is made

sdevalapurkar:
> In the edit location/project boundary, editing the region doesn't work.
> 
> Meaning: I cannot delete/add a region after the project is made

cool, please check now with my most recent commit :) thanks!

app/src/components/map/MapContainer.tsx
```diff
@@ -105,6 +203,32 @@ const MapContainer: React.FC<IMapContainerProps> = (props) => {
...
+    }
+  }, [geometryState?.geometry, nonEditableGeometries]);
+
+  useEffect(() => {
+    if (!bounds || !bounds.length) {
+      return;
+    }
+
+    throttledGetFeatureDetails(layersToInfer);
+  }, [bounds]);
```
NickPhura: Any reason why this wouldn't work? I tried it locally and it seems to work, though I don't know all of the edge cases to test, so maybe you can verify it.

```suggestion
  useEffect(() => {
    if (!geometryState?.geometry.length || !nonEditableGeometries?.length) {
      setInferredLayersInfo &&
        setInferredLayersInfo({
          parks: [],
          nrm: [],
          env: [],
          wmu: []
        });
    }

      throttledGetFeatureDetails(layersToInfer);
  }, [geometryState?.geometry, nonEditableGeometries]);
```

NickPhura: The idea being that we just update the feature details on geometry change, rather than bounds + all points (which seemed like a bit of an odd set of triggers for the side effect we want).

sdevalapurkar: great idea!

anissa-agahchen (approved):

PR #435, opened 2021-07-28
BHBC-1188: Timezone bug fix

NickPhura:
# Overview

## Links to jira tickets

https://quartech.atlassian.net/browse/BHBC-1188

- Add function in `db.ts` to stop node-postgres from adding default time/timezone information to psql `DATE` values (which by design don't contain time/timezone information).

Bonus:
- Update docker-compose/env/pipeline files for Db and API to allow a timezone to be specified (rather than just letting it default to whatever default value, UTC probably).

- [ ] Version change

## Checklist

### General

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

## How Has This Been Tested?

Tested locally.

Also used the PR-deploy to test that the timezone for API and DB do change if I change the config.json timezone values.

### DIFF
--- a/api/src/database/db.ts
+++ b/api/src/database/db.ts
@@ -34,7 +34,7 @@ defaultLog.debug({ label: 'create db pool', message: 'pool config', poolConfig }
 // This Can lead to unexpected behaviour when the original psql `DATE` value was intentionally omitting time/zone information.
 // PSQL date types: https://www.postgresql.org/docs/12/datatype-datetime.html
 // node-postgres type handling (see bottom of page): https://node-postgres.com/features/types
-pg.types.setTypeParser(1082, (stringValue: string) => {
+pg.types.setTypeParser(pg.types.builtins.DATE, (stringValue: string) => {
   return stringValue; // 1082 for `DATE` type
 });
 
--- a/api/src/database/db.ts
+++ b/api/src/database/db.ts
@@ -1,4 +1,4 @@
-import { Pool, PoolClient, PoolConfig, QueryResult } from 'pg';
+import * as pg from 'pg';
 import { HTTP400, HTTP500 } from '../errors/CustomError';
 import { setSystemUserContextSQL } from '../queries/user-context-queries';
 import { getUserIdentifier, getUserIdentitySource } from '../utils/keycloak-utils';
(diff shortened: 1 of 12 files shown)


rstens (commented):
So the standard TZ on the pod is already: ENV TZ America/Vancouver

I am a bit confused why this has to be reset.

NickPhura:
> 
> 
> So the standard TZ on the pod is already: ENV TZ America/Vancouver
> 
> I am a bit confused why this has to be reset.

Ya I kind of combined the actual change needed, with some bonus stuff.

The actual change that fixed some of the funny app dates is just a few lines in `db.ts`.

Everything else is just bonus to give us the capability of specifying the timezones used by the API and DB, in case we ever want to change it. A bunch of this came about naturally as part of testing the timezones (changing the API/DB to various timezone combinations to see how it affected the dates, etc), and so I just carried it forward into the pipeline since it wasn't a big additional change.

But ya, you can control the timezones individually for both api and db now in the local docker setup, and optionally change that root config.json if we ever wanted to have our openshift environments deployed in a different timezone (UTC, or w/e).

api/src/database/db.ts
```diff
@@ -29,10 +29,19 @@ const poolConfig: PoolConfig = {
 
 defaultLog.debug({ label: 'create db pool', message: 'pool config', poolConfig });
 
-let pool: Pool;
+// Custom type handler for psq `DATE` type to prevent local time/zone information from being added.
+// Why? By default, node-postgres assumes local time/zone for any psql `DATE` or `TIME` types that don't have timezone information.
+// This Can lead to unexpected behaviour when the original psql `DATE` value was intentionally omitting time/zone information.
+// PSQL date types: https://www.postgresql.org/docs/12/datatype-datetime.html
+// node-postgres type handling (see bottom of page): https://node-postgres.com/features/types
+pg.types.setTypeParser(1082, (stringValue: string) => {
```
anissa-agahchen:
```suggestion
pg.types.setTypeParser(pg.types.builtins.DATE, (stringValue: string) => {
```

anissa-agahchen: Do we need to set a similar typeparser for timestamp and timestampz?

NickPhura: oh nice one! I didn't realize there was a built in variable for that.

Right now its only `DATE` that we have issues with.  `timestamp(z)` already includes time/timezone information, so there is no issues as node-postgres doesn't alter that in any way.  Its only an issue for the types that don't have time/timezone information, as node-postgres decides to add it anyways, using some default values.  

I didn't add a function for `TIME`, which is the other type that node-postgres messes with, as we dont have any UI that creates time values yet.  We really only have a handful of date fields in the whole app and all of them are just start/end dates, with no time component.

anissa-agahchen (approved):

PR #449, opened 2021-08-10
BHBC-1241: Validation Endpoint

NickPhura:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1241

- Update validate endpoint

- [ ] Version change

## Checklist

### General

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

## How Has This Been Tested?

## Screenshots

### DIFF
--- a/api/src/paths/dwc/validate.ts
+++ b/api/src/paths/dwc/validate.ts
@@ -2,7 +2,7 @@ import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
 import { SYSTEM_ROLE } from '../../constants/roles';
 import { getDBConnection, IDBConnection } from '../../database/db';
-import { HTTP400 } from '../../errors/CustomError';
+import { HTTP400, HTTP500 } from '../../errors/CustomError';
 import {
   getSurveyOccurrenceSubmissionSQL,
   insertSurveySubmissionMessageSQL,
@@ -100,7 +100,7 @@ function getSubmissionS3Key(): RequestHandler {
 
       const response = await connection.query(sqlStatement.text, sqlStatement.values);
 
-      if (!response || !response.rowCount) {
+      if (!response?.rows?.length) {
         throw new HTTP400('Failed to get survey occurrence submission');
       }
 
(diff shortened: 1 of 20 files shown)


api/src/paths/dwc/validate.ts
```diff
@@ -161,26 +77,78 @@ POST.apiDoc = {
...
+
+      if (!sqlStatement) {
+        throw new HTTP400('Failed to build SQL get statement');
+      }
+
+      await connection.open();
+
+      const response = await connection.query(sqlStatement.text, sqlStatement.values);
+
+      if (!response || !response.rowCount) {
```
sdevalapurkar: we should check for existence of `response.rows` and maybe `response.rows[0]` here so we can confidently get the key from it below

api/src/paths/dwc/validate.ts
```diff
@@ -161,26 +77,78 @@ POST.apiDoc = {
...
     try {
-      const rawMediaArray: Express.Multer.File[] = req.files as Express.Multer.File[];
+      const s3Key = req['s3Key'];
 
-      const rawMediaFile = rawMediaArray[0];
+      const s3File = await getFileFromS3(s3Key);
 
-      const mediaFiles = parseUnknownMedia(rawMediaFile);
+      if (!s3File) {
+        throw new HTTP400('Failed to get occurrence submission file');
```
sdevalapurkar: question: here is it a 400 cause its bad request from user or should it be 500 like we failed to get the file cause internal issue?

sdevalapurkar (commented):
looking good! :)

api/src/paths/dwc/validate.ts
NickPhura: I double checked what rowCount is, and my understanding had been a bit off.  rowCount is the number of affected rows, but not necessarily the number of rows returned, which makes sense.  So ya, changing this to check for rows is appropriate.

api/src/paths/dwc/validate.ts
NickPhura: Yeah I can update this.  

A bit of tech debt for the todo list is to do a general audit of all of our errors, and make sure they are informative. We can also tack on any additional error messaging as the second param in all of the HTTP errors.

sdevalapurkar (dismissed):

sdevalapurkar (approved):

PR #450, opened 2021-08-10
Tests and cleanup

sdevalapurkar:
# Overview

Tests and cleanup


app/src/components/map/WFSFeatureGroup.tsx
```diff
@@ -135,6 +135,7 @@ const WFSFeatureGroup: React.FC<IWFSFeatureGroupProps> = (props) => {
     )?.catch(/* catch and ignore errors */);
 
     setFeatures(newFeatures);
+    // eslint-disable-next-line
```
NickPhura:
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
```

app/src/components/map/MapContainer.tsx
```diff
@@ -219,6 +219,7 @@ const MapContainer: React.FC<IMapContainerProps> = (props) => {
     }
 
     throttledGetFeatureDetails(layersToInfer);
+    // eslint-disable-next-line
```
NickPhura:
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
```

api/src/paths/project/{projectId}/attachments/upload.ts
```diff
@@ -92,7 +92,7 @@ POST.apiDoc = {
  */
 export function uploadMedia(): RequestHandler {
   return async (req, res) => {
-    defaultLog.debug({ label: 'uploadMedia', message: 'files.length', files: req?.files?.length });
+    defaultLog.debug({ label: 'uploadMedia', message: 'files', files: req.files });
```
NickPhura: We cant print req.files, because that will print all of the binary information, which is potentially massive

NickPhura: If we wanted to print the files, we'd need to do something like: `files: {...req.files, buffer: 'too big to print'}`

sdevalapurkar: cool

api/src/paths/project/{projectId}/attachments/upload.ts
```diff
@@ -92,7 +92,7 @@ POST.apiDoc = {
  */
 export function uploadMedia(): RequestHandler {
   return async (req, res) => {
-    defaultLog.debug({ label: 'uploadMedia', message: 'files.length', files: req?.files?.length });
+    defaultLog.debug({ label: 'uploadMedia', message: 'files', files: { ...req.files, buffer: 'Too big to print' } });
```
NickPhura:
```typescript
req.files.map(item => {return {...item, buffer: 'Too big to print'}});
```

NickPhura (approved):

PR #452, opened 2021-08-11
Hook up BioHub validated end-point to store errors and be callable by the workflow engine (XLSX)

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1242

- Hook up BioHub validated end-point to store errors and be callable by the workflow engine

## How Has This Been Tested?

- Locally and tests/db status table


api/src/utils/media/csv/csv-file.ts
```diff
@@ -171,6 +202,30 @@ export class XLSXCSV implements IWorkbook {
...
 
     this.workbook = new CSVWorkBook(rawWorkbook);
+
+    this._init();
+  }
+
+  _init() {
+    switch (this.rawFile.name) {
+      case XLSX_CLASS.SAMPLE_STATION_INFORMATION:
+        this.workbook.worksheets['Sample Station Information'] = new CSVWorksheet(
```
NickPhura: I think you had an enum for these worksheet names: `Sample Station Information`, etc

NickPhura: Eventually we'll have to fetch those from a config somewhere (since not all templates will have the same sheet names) alongside the validation rules, etc.

anissa-agahchen (approved):

PR #454, opened 2021-08-12
Update basemap to be BCGov instead of ESRI Imagery

sdevalapurkar:
# Overview

Update basemap to be BCGov instead of ESRI Imagery

## Screenshots

image


NickPhura (approved):
This PR is a perfect example of why phasing out snapshot tests is the way to go, haha

sdevalapurkar:
> This PR is a perfect example of why phasing out snapshot tests is the way to go, haha

exactly haha

PR #459, opened 2021-08-18
Get map inferred layers info working (final pass for now)

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1251

- Get map inferred layers info working (final pass for now)

## How Has This Been Tested?

- Locally and tests


database/src/migrations/20210818170006_add_geojson_to_project_and_survey.ts
```diff
@@ -0,0 +1,39 @@
...
+    set role biohub_api;
+
+    create or replace view project as select * from ${DB_SCHEMA}.project;
+    create or replace view survey as select * from ${DB_SCHEMA}.survey;
+
+    set role postgres;
+  `);
+}
+
+export async function down(knex: Knex): Promise<void> {
```
NickPhura, line 24: technically this isnt quite right, as the views cant jsut be dropped, they need to be updated (just the reverse of what you have in the up):

1. alter table drop geojson column
2. create or replace view

api/src/queries/survey/survey-update-queries.ts
```diff
@@ -139,6 +139,7 @@ export const putSurveyDetailsSQL = (
       lead_first_name = ${data.lead_first_name},
       lead_last_name = ${data.lead_last_name},
       location_name = ${data.location_name},
+      geojson = ${JSON.stringify(data.geometry)},
```
NickPhura, line 142: do we need to stringify the json?

sdevalapurkar: yes, otherwise it isn't valid

database/src/migrations/20210818170006_add_geojson_to_project_and_survey.ts
sdevalapurkar: spoke to charlie - he said not to worry about the down at all
but I can make the update

app/src/features/projects/view/components/LocationBoundary.tsx
```diff
@@ -125,12 +125,11 @@ const LocationBoundary: React.FC<ILocationBoundaryProps> = (props) => {
...
 
   useEffect(() => {
-    const geometryCollectionResult = generateValidGeometryCollection(location.geometry);
-    const nonEditableGeometriesResult = geometryCollectionResult.geometryCollection.map((geom: Feature) => {
+    const nonEditableGeometriesResult = location.geometry.map((geom: Feature) => {
       return { feature: geom };
     });
 
-    setBounds(geometryCollectionResult.bounds);
+    updateMapBounds(location.geometry, setBounds);
```
NickPhura: Wondering about this function, any reason for passing in the setBounds function, and not just returning the bounds from this function, and then calling setBounds here? In this case, the function would become something like `getMapBounds(geometry: Feature[]): number[][]` (not sure on the exact types)

database/src/migrations/20210818170006_add_geojson_to_project_and_survey.ts
NickPhura: no worries, either way, just wanted to point it out in case

api/src/queries/survey/survey-update-queries.ts
NickPhura: Does it also fail if we just wrap it in quotes?

app/src/features/projects/view/components/LocationBoundary.tsx
sdevalapurkar: ok cool

anissa-agahchen (approved):

PR #462, opened 2021-08-19
BHBC 1289: Soft Delete an Uploaded Observation

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1289

- Soft Delete an Uploaded Observation

## How Has This Been Tested?

- Local and tests


database/src/migrations/20210819170006_add_soft_delete_timestamp_to_occurrence_submission.ts
```diff
@@ -0,0 +1,35 @@
+import * as Knex from 'knex';
+
+const DB_SCHEMA = process.env.DB_SCHEMA;
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set schema '${DB_SCHEMA}';
+    set search_path = ${DB_SCHEMA},public;
+
+    ALTER TABLE ${DB_SCHEMA}.occurrence_submission add column soft_delete_timestamp timestamptz(6);
```
NickPhura, line 10: Not a major issue, but I think just `delete_timestamp` is sufficient.  The soft part is implied by the fact that the record still exists, and was just marked deleted (rather than actually being gone from the DB)

anissa-agahchen (approved):

database/src/migrations/20210819170006_add_soft_delete_timestamp_to_occurrence_submission.ts
NickPhura: Also keeps it in line with the format of the existing `publish_timestamp`

sdevalapurkar: the soft delete terminology was what charlie had put in my ticket so I just went with it :)

sdevalapurkar: so if its ok, ill just leave it as is for now and in my next pr (which will be updating the survey view, i can change it there)

app/src/features/surveys/view/SurveyObservations.tsx
```diff
@@ -79,32 +80,34 @@ const SurveyObservations: React.FC = () => {
...
-          ) {
-            setIsValidating(false);
-            setIsPolling(false);
-
-            setPollingTime(null);
-          } else {
-            setIsValidating(true);
-            setIsPolling(true);
-          }
+  const fetchObservationSubmission = useCallback(async () => {
```
NickPhura, line 83: With this function moved out of the use effect we no longer need to `pollingFunction` variable, as we can set the polling function in the `useInterval` call from the start (and then the polling can just be controlled by the pollingTime variable)

database/src/migrations/20210819170006_add_soft_delete_timestamp_to_occurrence_submission.ts
NickPhura: Sounds good

app/src/features/surveys/view/SurveyObservations.tsx
NickPhura: This wont really change anything, just a small optimization

PR #471, opened 2021-08-25
BHBC-1284: Create a N8N webhook workflow to call the validation end-points

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1284

- Create a N8N webhook workflow to call the validation end-points

## How Has This Been Tested?

- Local and docker

## Screenshots

image


rstens (commented):
We should store the workflow json in the repo as well. However, there is an issue with hardcoded api locations, what we might want to do is building these to read from an environment variable that can be accessed in n8n. So that we can store the API location there and then when we move the workflow into another environment, it would still work.

Also I did find out that once you are in the workflow in OpenShift, you can call the API on the internal network: e.g. http://biohubbc-api-dev-deploy.af2668-dev.svc:6100/api/version.

I actually created a workflow which dynamically builds the url depending on which environmnet we are in: https://n8n-af2668-dev.apps.silver.devops.gov.bc.ca/workflow/12

app/src/features/surveys/view/SurveyObservations.tsx
```diff
@@ -57,14 +57,21 @@ const SurveyObservations = () => {
...
-      return biohubApi.observation.uploadObservationSubmission(
+    return async (files, cancelToken, handleFileUploadProgress) => {
+      const uploadResult = biohubApi.observation.uploadObservationSubmission(
         projectId,
         surveyId,
         files[0],
         cancelToken,
         handleFileUploadProgress
       );
+      const awaitedUploadResult = await uploadResult;
```
NickPhura: Any reason for setting the await here, rather than on line 61? `const uploadResult = biohubApi...`

sdevalapurkar: yeah cause I need to return the promise, not the result of the awaited promise

NickPhura:
```typescript
  const importObservations = (): IUploadHandler => {
    return (files, cancelToken, handleFileUploadProgress) => {
      return biohubApi.observation
        .uploadObservationSubmission(projectId, surveyId, files[0], cancelToken, handleFileUploadProgress)
        .then((result) => {
          if (!result || !result.submissionId) {
            return;
          }

          return biohubApi.observation.initiateSubmissionValidation(awaitedUploadResult.submissionId, files[0].type);
        });
    };
  };
```

env_config/env.docker
```diff
@@ -66,3 +66,27 @@ OBJECT_STORE_BUCKET_NAME=
...
+
+# ------------------------------------------------------------------------------
+# n8n - https://docs.n8n.io/reference/environment-variables.html
+# ------------------------------------------------------------------------------
+N8N_HOST=localhost
+N8N_API_HOST=api
+N8N_PROTOCOL=http
+N8N_PORT=5678
+N8N_PROXY_PORT=5100
+N8N_NGINX_PORT=5101
```
NickPhura: I dont think this is used anywhere

n8n/.config/configstore/update-notifier-npm.json
```diff
@@ -0,0 +1,4 @@
+{
+	"optOut": false,
+	"lastUpdateCheck": 1630011451262
+}
```
NickPhura: I don't think we want to include this (I think it can be ignored)?

Seems like a bit of throw away config generated by n8n?

n8n/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "0.0.0",
+  "description": "N8N setup",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/bcgov/biohubbc.git"
+  },
+  "engines": {
+    "node": ">= 10.0.0",
+    "npm": ">= 6.0.0"
```
NickPhura: Could specify node 14 (and whatever npm version that yields).

(Not really sure when this is useful tbh, but more information is never bad)

sdevalapurkar:
weird issue with active state and production url not working on initial spin-up: https://github.com/n8n-io/n8n/issues/595

sdevalapurkar:
https://github.com/n8n-io/n8n/issues/2155

issue I created referencing the issue - hope to hear back soon

NickPhura (approved):

PR #485, opened 2021-08-31
Public View (Iteration #1)

sdevalapurkar:
# Overview

- First iteration on public views

## How Has This Been Tested?

- Locally


NickPhura:
Something @jeznorth can decide on ultimately, but we may still want the login page.  

It gives us a place to throw the user once they log out, that explicitly informs them that they are no longer logged in (users like knowing definitively if they are no longer logged in). This post goes into more detail: https://ux.stackexchange.com/questions/103274/what-page-should-be-shown-to-the-user-after-s-he-sign-out

It also gives us a bookmarkable link that can be used anywhere, that takes a user right to the admin side of the app. Currently, even if I send someone a link to `/projects` it automatically bounces them to `/public`, which is possibly confusing if they were expecting to login to the admin site.

app/src/components/layout/Header.tsx
```diff
@@ -144,20 +161,28 @@ const Header: React.FC = () => {
                 </picture>
                 Species Inventory Management System
               </Link>
-              <LoggedInUser />
+              {(!keycloakWrapper?.keycloak?.authenticated || !keycloakWrapper?.hasLoadedAllUserInfo) && (
```
NickPhura, line 164: Can be done later: We'll probably want to add a utility function (`isUserAuthenticated()` or something) that checks if the user is authenticated and loaded so we dont have to paste all of this everywhere.

NickPhura (dismissed):

app/src/components/layout/Header.tsx
sdevalapurkar: sounds good, will look into in the next iteration of this :)

NickPhura (approved):

PR #498, opened 2021-09-03
Update queries to get version as well as name for common survey methodology

sdevalapurkar:
# Overview

Update queries to get version as well as name for common survey methodology


anissa-agahchen (approved):
Looks good !

PR #506, opened 2021-09-08
BHBC: 1317 - Scan Uploaded Attachments

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1317

- Scan Uploaded Attachments using Clamav and report error back to frontend to be shown and fail upload of malicious file

## How Has This Been Tested?

- Locally

## Notes

- https://bcgov.github.io/common-service-showcase/services/cavms.html
- https://www.npmjs.com/package/node-clamav


sdevalapurkar:
test files: https://github.com/fire1ce/eicar-standard-antivirus-test-files

anissa-agahchen (dismissed):
Looks good, tested and run locally

app/src/components/attachments/DropZone.tsx
```diff
@@ -75,7 +75,7 @@ export const DropZone: React.FC<IDropZoneProps & IDropZoneConfigProps> = (props)
               <Typography variant="h3" className={classes.textSpacing}>
                 Drag your files here, or <Link className={classes.browseLink}>Browse Files</Link>
               </Typography>
-              {!!maxFileSize && maxFileSize !== Infinity && (
+              {!!maxFileSize && maxFileSize !== Infinity && acceptedFileExtensions && (
```
NickPhura: I don't think we need the maxFileSize check here.  It won't impact the extensions we allow

Makefile
```diff
@@ -136,28 +138,28 @@ run-backend-debug: ## Runs all backend containers in debug mode, where all conta
...
-## - Builds all of the biohub backend+web projects (db, db_setup, api, app, n8n, n8n_nginx, n8n_setup)
+## - Builds all of the biohub backend+web projects (db, db_setup, api, app, n8n, n8n_nginx, n8n_setup clamav)
 ## ------------------------------------------------------------------------------
 
 build-web: ## Builds all backend+web containers
 	@echo "==============================================="
 	@echo "Make: build-web - building web images"
 	@echo "==============================================="
-	@docker-compose -f docker-compose.yml build db db_setup api app n8n n8n_nginx n8n_setup
+	@docker-compose -f docker-compose.yml build db db_setup api app n8n n8n_nginx n8n_setup clamav
```
NickPhura: We should take clamav out of the web commands, and jsut have it be an optional thing someone can spin up if they want to. Otherwise the "skip clamav" variable isn't doing much as we are still spending the time to spin up the service locally, only to then ignore it.

env_config/env.docker
```diff
@@ -89,3 +89,10 @@ N8N_BASIC_AUTH_ACTIVE=true
 N8N_BASIC_AUTH_USER=n8n
 N8N_BASIC_AUTH_PASSWORD=n8n
 N8N_ENCRYPTION_KEY=secret
+
+# ------------------------------------------------------------------------------
+# Clamav - Virus scanning
+# ------------------------------------------------------------------------------
+CLAMAV_PORT=3310
+CLAMAV_HOST=clamav
+SHOULD_CLAMAV_SCAN=true
```
NickPhura: Wondering if there is a better name for this.

Like, `ENABLE_CLAMAV` or something

api/src/utils/file-utils.ts
```diff
@@ -1,7 +1,11 @@
 import AWS from 'aws-sdk';
 import { DeleteObjectOutput, GetObjectOutput, ManagedUpload, Metadata } from 'aws-sdk/clients/s3';
 import { S3_ROLE } from '../constants/roles';
+import clamd from 'clamdjs';
 
+const scanner = process.env.SHOULD_CLAMAV_SCAN
+  ? clamd.createScanner(process.env.CLAMAV_HOST || 'clamav', Number(process.env.CLAMAV_PORT) || 3310)
```
NickPhura: nit: Do we need the hardcoded defaults?

If someone has `SHOULD_CLAMAV_SCAN` enabled, Id think its safe to assume they will also have the HOST/PORT variables set.

api/openshift/api.dc.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   selfLink: ''
   name: biohubbc-api-dc
 parameters:
+  - name: SHOULD_CLAMAV_SCAN
+    value: '1'
```
NickPhura: Do we want the default to be `true` instead of `1`?

In your check in file-utils.ts, you could then check if `process.env.SHOULD_CLAMAV_SCAN === 'true'`

NickPhura: This way, someone setting it to `false` will still result in it being disabled, even those its a truthy string.

Makefile
sdevalapurkar: makes sense

env_config/env.docker
NickPhura: Or maybe something more descriptive, like `ENABLE_FILE_VIRUS_SCAN`

api/src/utils/file-utils.ts
sdevalapurkar: we want it hard coded default so that it just always works in Openshift even without the env variables cause the values will never change regardless of environment

NickPhura (approved):
Nice work

PR #507, opened 2021-09-08
fixed upload of observations with only recommended rules

anissa-agahchen:
This PR ensures that a submission that is only missing recommended headers goes to completion, and that the warning messages are recorded in the database


sdevalapurkar (dismissed):
too small please add more changes

NickPhura (approved):
Nice

PR #511, opened 2021-09-10
BHBC-1350: Parse validation schema into rules

NickPhura:
# Overview

## Links to jira tickets

https://quartech.atlassian.net/browse/BHBC-1350

Schema parser class to turn validation schemas into concrete rules.
Updates to endpoints to leverage schema-based rules.
Update validation functions to apply their check against all rows for 1 column, rather than run against all rows for all columns.
Update existing unit tests to pass
Renamed validation json-schema file.
Removed some old unused package dependencies.
Removed some old unused code.

- [ ] Version change

## Checklist

### General

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

## How Has This Been Tested?


anissa-agahchen (dismissed):
Looks good

api/src/json-schema/validation-schema.test.ts
```diff
@@ -0,0 +1,239 @@
...
+                      description: 'An unclassified sex'
+                    }
+                  ]
+                }
+              }
+            ]
+          },
+          {
+            name: 'lifeStage',
+            desciption: 'A description of the location.',
```
sdevalapurkar: question: the lifestage is the description of the location?

api/src/json-schema/validation-schema.test.ts
```diff
@@ -0,0 +1,239 @@
...
+              name: '',
+              description: '',
+              recommended_columns: ['occurrenceRemarks']
+            }
+          }
+        ]
+      },
+      {
+        name: 'measurementorfact',
+        description: 'The Event file for this submission. Should contain information about events.',
```
sdevalapurkar: seems like copy paste issue for the description here

api/src/json-schema/validation-schema.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        columns: [],
+        validations: []
+      },
+      {
+        name: 'resourcerelationship',
+        description: 'The Event file for this submission. Should contain information about events.',
```
sdevalapurkar: seems like copy paste issue for the description here

api/src/json-schema/validation-schema.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        columns: [],
+        validations: []
+      },
+      {
+        name: 'taxon',
+        description: 'The Event file for this submission. Should contain information about events.',
```
sdevalapurkar: seems like copy paste issue for the description here

api/src/utils/media/csv/validation/csv-header-validator.ts
```diff
@@ -33,23 +33,36 @@ export const getDuplicateHeadersValidator = (): CSVValidator => {
...
   return (csvWorksheet) => {
-    if (!requiredHeaders?.length) {
+    if (!config) {
+      return csvWorksheet;
+    }
+
+    if (!config.file_required_columns_validator.required_columns.length) {
+      // No required columns
       return csvWorksheet;
     }
```
sdevalapurkar, line 59: question: can we combine these if statements?

api/src/utils/media/csv/validation/csv-header-validator.ts
```diff
@@ -77,23 +90,35 @@ export const hasRequiredHeadersValidator = (requiredHeaders?: string[]): CSVVali
...
+export const hasRecommendedHeadersValidator = (config?: FileRecommendedHeaderValidatorConfig): CSVValidator => {
   return (csvWorksheet) => {
-    if (!recommendedHeaders?.length) {
+    if (!config) {
+      return csvWorksheet;
+    }
+
+    if (!config.file_recommended_columns_validator.recommended_columns.length) {
       return csvWorksheet;
     }
```
sdevalapurkar, line 115: same question here, if not relevant in prev comment then obviously not relevant here

api/src/utils/media/csv/validation/csv-row-validator.ts
```diff
@@ -62,127 +59,159 @@ export const getRequiredFieldsValidator = (requiredFieldsByHeader?: string[]): C
...
             }
           ]);
         }
-
-        // Add an error if the cell value is not in the correct range provided in the array
-
-        if (rowValueForColumn < valueRangesByHeader.min_value || rowValueForColumn > valueRangesByHeader.max_value) {
+      } else if (!config.column_range_validator.min_value && config.column_range_validator.max_value) {
+        // Value must be less than max value
+        if (rowValueForColumn > config.column_range_validator.max_value) {
```
sdevalapurkar, line 195: question: the comment says must be less than but the check is for greater than and not for greater than or equal to
maybe either the check or the comment needs to be updated?

api/src/utils/media/csv/validation/csv-row-validator.ts
```diff
@@ -62,127 +59,159 @@ export const getRequiredFieldsValidator = (requiredFieldsByHeader?: string[]): C
...
+              errorCode: 'Out of Range',
+              message: `Invalid value: ${rowValueForColumn}. Value must be less than ${config.column_range_validator.max_value} `,
+              col: config.columnName,
+              row: rowIndex + 2
+            }
+          ]);
+        }
+      } else if (config.column_range_validator.min_value && !config.column_range_validator.max_value) {
+        // Value must be greater than min value
+        if (rowValueForColumn < config.column_range_validator.min_value) {
```
sdevalapurkar, line 208: same here: greater than vs. greater than or equal to

api/src/utils/media/validation/file-type-and-content-validator.ts
```diff
@@ -10,18 +12,112 @@ export const getFileEmptyValidator = (): MediaValidator => {
...
+  };
+};
+
+export const getRequiredFilesValidator = (
+  config?: SubmissionRequiredFilesValidatorConfig
+): DWCArchiveValidator | XLSXCSVValidator => {
+  return (file: any) => {
+    if (!config) {
+      // No required files specified
+
```
sdevalapurkar: super nit: 
```suggestion
```

api/src/utils/media/validation/validation-schema-parser.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  getColumnValidationsJsonpath(fileName: string, columnName: string): string {
+    return `$.files[?(@.name == '${fileName}')].columns[?(@.name == '${columnName}')].validations`;
+  }
+
+  parseJson(json: any): object {
+    let parsedJson;
+
+    try {
+      parsedJson = JSON.parse(json);
```
sdevalapurkar: question: can we just do `return JSON.parse(json)` in the try and then if it doesnt work, then it will catch?
or does it have to be done this way where the return is outside the try/catch block?

app/src/features/surveys/components/GeneralInformationForm.tsx
```diff
@@ -169,7 +169,7 @@ const GeneralInformationForm: React.FC<IGeneralInformationFormProps> = (props) =
           />
         </Grid>
         <Grid item xs={12}>
-          <FormControl fullWidth variant="outlined" style={{ width: '100%' }}>
+          <FormControl fullWidth variant="outlined" required={true} style={{ width: '100%' }}>
```
sdevalapurkar, line 172: question: we decided that methodology is gonna be a required field now?

sdevalapurkar: is this change only gonna be on fe side and the model will still allow it to be optional (not null) for now?

NickPhura: We just added it to the frontend for now, because without picking a methodology, none of the validation runs correctly (as that is a required piece in determining the template/validation schema)

NickPhura: But, the whole template/methodology stuff is likely going to evolve a bit, so simple fix for now is best

api/src/utils/media/validation/validation-schema-parser.ts
NickPhura: Yeah we can, good catch.  If there wasn't a throw in the catch, then it would complain about no return at the end.  But it appears smart enough to know it will always return an object OR throw.

api/src/utils/media/csv/validation/csv-row-validator.ts
NickPhura: I think if the rule says that the range is "0 to 10" then it makes sense that 0 and 10 are valid options.  Otherwise you have to get into weird cases where the range is "-0.1 to 10.1" or something silly.  Or handle customizing the equality checks as part of the config, which is probably more than we need for any of our validation.

api/src/utils/media/csv/validation/csv-row-validator.ts
NickPhura: The comment is maybe just a bit confusing.

The value does indeed need to be less than the max value.  If it is greater, inside that if we add an error.

api/src/utils/media/csv/validation/csv-header-validator.ts
NickPhura: They could be combined, I had them split out just for readability.  But could go either way. For whatever reason It looks nicer to me split out, but not sure why

sdevalapurkar (approved):
code lgtm

PR #522, opened 2021-09-16
Fix issue where making an attachment private makes all other attachments private too

sdevalapurkar:
# Overview

Fix issue where making an attachment private makes all other attachments private too

This was because the rule was being defined based on project id and not based on project attachment id


anissa-agahchen (approved):
Works!

PR #523, opened 2021-09-16
Public view of project metadata

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1312

- Public view of project metadata

## How Has This Been Tested?

- Locally

## Screenshots

image


api/src/models/project-view.ts
```diff
@@ -39,6 +39,41 @@ export class GetProjectData {
...
+    this.project_activities = (activityData?.length && activityData.map((item) => item.name)) || [];
+    this.start_date = projectData?.start_date || '';
+    this.end_date = projectData?.end_date || '';
+    this.comments = projectData?.comments || '';
+    this.completion_status =
+      (projectData &&
+        projectData.end_date &&
+        moment(projectData.end_date).endOf('day').isBefore(moment()) &&
+        'Completed') ||
+      'Active';
```
NickPhura: nit: can we make an enum for these states? And use it wherever they are currently used.

api/src/queries/project/project-view-queries.ts
```diff
@@ -3,6 +3,62 @@ import { getLogger } from '../../utils/logger';
...
+      project.coordinator_last_name,
+      project.coordinator_email_address,
+      project.coordinator_agency_name,
+      project.coordinator_public,
+      project.geojson as geometry,
+      project.create_date,
+      project.create_user,
+      project.update_date,
+      project.update_user,
+      project.revision_count,
```
NickPhura: I would remove these.  partially because we don't use them on the frontend (right?), and partially because it contains internal information about the project that probably shouldn't be public (mainly the user info).

api/src/queries/project/project-view-queries.ts
```diff
@@ -3,6 +3,62 @@ import { getLogger } from '../../utils/logger';
...
+      project.location_description,
+      project.start_date,
+      project.end_date,
+      project.caveats,
+      project.comments,
+      project.coordinator_first_name,
+      project.coordinator_last_name,
+      project.coordinator_email_address,
+      project.coordinator_agency_name,
+      project.coordinator_public,
```
NickPhura: Can this be public?

sdevalapurkar: not always, it depends on state of `coordinator_public`
will change this, nice catch

app/src/hooks/api/useProjectApi.ts
```diff
@@ -105,6 +105,18 @@ const useProjectApi = (axios: AxiosInstance) => {
     return data;
   };
 
+  /**
+   * Get public (published) project details based on its ID for viewing purposes.
+   *
+   * @param {number} projectId
+   * @return {*} {Promise<IGetProjectForViewResponse>}
+   */
+  const getPublicProjectForView = async (projectId: number): Promise<IGetProjectForViewResponse> => {
```
NickPhura, line 114: Doesn't need to happen now, but it might be nice to nest these public functions under some sort of `public` object.

So that we can do something like `biohubApi.projects.public.getProjectForView()` or `biohubApi.public.projects.getProjectForView()`

That way we don't need to include the word `public` in every function name, and its a bit clearer which you are calling.

api/src/queries/project/project-view-queries.ts
NickPhura: I think robin mentioned something about the agency always being visible? its just the name/email that is conditional

api/src/models/project-view.ts
```diff
@@ -39,6 +39,41 @@ export class GetProjectData {
   }
 }
 
+/**
+ * Pre-processes GET /projects/{id} public (published) project data
+ *
+ * @export
+ * @class GetPublicProjectData
+ */
+export class GetPublicProjectData {
```
NickPhura: I'm thinking it might be nice to separate the public models from the non-public models.

Maybe put them all under `api/src/models/public`.

api/src/queries/project/project-view-queries.ts
```diff
@@ -3,6 +3,62 @@ import { getLogger } from '../../utils/logger';
 
 const defaultLog = getLogger('queries/project/project-view-queries');
 
+/**
+ * SQL query to get a single public (published) project.
+ *
+ * @param {number} projectId
+ * @returns {SQLStatement} sql query object
+ */
+export const getPublicProjectSQL = (projectId: number): SQLStatement | null => {
```
NickPhura: I'm thinking it might be nice to separate the public queries from the non-public queries.

Maybe put them all under `api/src/queries/<project>/public` or alternatively `api/src/queries/public/<project>`

NickPhura: This would also make them align with the paths, which are already under `api/paths/public`

app/src/hooks/api/useProjectApi.ts
sdevalapurkar: cool!

NickPhura (approved):
Looks good!

Did a quick test in openshift, and everything seems in order.

The snapshot tests are not super ideal, but I'm also not sure how much intense testing we really care to do for the public pages, given its all just non-editable static data. So as long as the API isn't revealing anything it shouldn't, its not really a big deal if the public pages work correctly.

PR #537, opened 2021-09-23
See Survey Observations on a Map + More public view stuff

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1302

- See Survey Observations on a Map

## How Has This Been Tested?

- Locally

## Screenshots

image


api/src/paths/dwc/validate.ts
```diff
@@ -225,47 +224,11 @@ export function persistParseErrors(): RequestHandler {
...
-          connection
-        );
-
-        await connection.commit();
-
-        return res.status(200).json();
-      }
-
-      req['validationSchema'] = validationSchema;
+    req['validationSchema'] = {};
```
anissa-agahchen, line 229: can you restore this to the old functionality?

api/src/paths/xlsx/validate.ts
```diff
@@ -72,6 +76,50 @@ function prepXLSX(): RequestHandler {
   };
 }
 
+export function getValidationSchema(): RequestHandler {
```
anissa-agahchen, line 79: once restored for dwc, delete from here

anissa-agahchen (changes requested):

api/src/paths/dwc/scrape-occurrences.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    const occurrenceSubmissionId = req.body.occurrence_submission_id;
+    const file = req['dwcArchive'];
+
+    const connection = getDBConnection(req['keycloak_token']);
+
+    try {
+      await connection.open();
+
+      const eventHeaders = file.worksheets.event?.getHeaders();
```
anissa-agahchen: I thought we had common code for the validate endpoints that get the sheets, headers and rows.  Can we leverage that?

api/src/paths/xlsx/validate.ts
sdevalapurkar: we need it in xlsx but not in dwc

api/src/paths/dwc/scrape-occurrences.ts
sdevalapurkar: which one?

api/src/paths/xlsx/validate.ts
sdevalapurkar: cause for xlsx we want to get the validation schema but for dwc we dont as of now

anissa-agahchen: the old workflow was in dwc, and called by xlsx.  Either way, we need to make sure they're not able to upload dwc for the time being.

sdevalapurkar: if they are unable to upload dwc then all the work ive done in this ticket is useless and cannot be tested...

sdevalapurkar: can we leave it this way until the transformation piece is in place

anissa-agahchen: right .... I see what you mean.  Can we have the client test in openshift?

@NickPhura which way should we go?

api/src/paths/dwc/scrape-occurrences.ts
anissa-agahchen: one sec ... lets go to discord

api/src/paths/xlsx/validate.ts
sdevalapurkar: I'd say leave it for now and then when transformation is done and in dev, we just change this to not allow dwc and it will still work since the xlsx can be transformed to dwc and this functionality can still be called

anissa-agahchen (approved):

PR #575, opened 2021-10-07
Warning Indicator for Non-supported Browser

sdevalapurkar:
# Overview

## Links to jira tickets

- https://quartech.atlassian.net/browse/BHBC-1357

- Warning Indicator for Non-supported Browser

## How Has This Been Tested?

- Locally


app/src/layouts/PublicLayout.tsx
```diff
@@ -1,15 +1,37 @@
...
+
+    return false;
+  }
+
   return (
     <Box height="100vh" display="flex" flexDirection="column">
       <CssBaseline />
       <DialogContextProvider>
+        {!isSupportedBrowser() && showUnsupportedBrowserAlert && (
+          <Alert onClose={() => setShowUnsupportedBrowserAlert(false)} severity="error">
```
NickPhura: I think we can get rid of the close button entirely, as even if they close it, it pops up when they change pages.

NickPhura: And we don't care if it stays on top, as we dont want them using that browser anyways

NickPhura (approved):

PR #592, opened 2021-10-13
🚅 Let the frontend tests begin, start saying goodbye to snapshots

sdevalapurkar:
# Overview

🚅 Let the frontend tests begin

And the snapshots going away begins too


NickPhura (approved):
Nice!

PR #646, opened 2021-11-18
Test Script update, login and some other details

rstens:
Fixing the login:

- Found a Cypress-Keycloak module that I rebuilt to deal with dependencies. Now added as keycloak.js to the support files.
- We can still use the cypress@idir user that way
- Fixed some test code to work with the new layout, ids
- Had to leave creating a survey behind for now
- Upgraded keycloak to 9.0.0
- Had to solve timing and path issues - The application does behave a bit different when run on OpenShift, it takes longer to figure out the authentication and I was too quick with continuing the test.


NickPhura (dismissed):
Looks good

You mentioned the keycloak script was generated from something? Might be good to include a comment or something in the readme that points to how it was made, etc.

rstens:
> Looks good
> 
> You mentioned the keycloak script was generated from something? Might be good to include a comment or something in the readme that points to how it was made, etc.

Yes, good point. I'll add a reference in the readme

PR #665, opened 2021-12-07
BHBC-1457

cgarrettjones:
# Overview

## Links to jira tickets

- [BHBC-1525](https://quartech.atlassian.net/browse/)

 - instantiate EML test with some initial tests

- [ ] Version change

## Checklist

### General

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

## How Has This Been Tested?

## Screenshots


api/src/paths/dwc/eml.test.ts
```diff
@@ -0,0 +1,662 @@
...
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  it('should throw a 400 error when no data package id is provided', async () => {
+    sinon.stub(db, 'getDBConnection').returns(dbConnectionObj);
+
+    try {
+      const result = eml.getSurveyOccurrenceSubmission(sampleReq.data_package_id, dbConnectionObj);
+      await result;
```
NickPhura, line 67:
```suggestion
      await eml.getSurveyOccurrenceSubmission(sampleReq.data_package_id, dbConnectionObj);
```

NickPhura: Tiny optimization.
Same story for many of the cases below.

In the original code `result` is a promise, and since the test code isn't doing anything with the promise (other than waiting for it to finish) we don't need to assign the promise to a variable, we can just do it all in one line.

api/src/paths/dwc/eml.test.ts
```diff
@@ -0,0 +1,662 @@
...
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  it('should throw a 400 error when no data package id is provided', async () => {
+    sinon.stub(db, 'getDBConnection').returns(dbConnectionObj);
+
+    try {
+      const result = eml.getSurveyDataPackageEML();
+      await result(
```
NickPhura, line 37:
```suggestion
      const requestHandler = eml.getSurveyDataPackageEML();
      await requestHandler(
```

NickPhura: (This isn't something you could have really known, as there are tons of cases of the former naming convention in the code base)

Calling the return value from `getSurveyDataPackageEML` (or any endpoint function) `result` is a bit misleading, as its really just a function (specifically a RequestHandler function).  Moving forward it would be a slight improvement to readability if we called them `requestHandler` (or something along those lines).

api/src/paths/dwc/eml.test.ts
NickPhura: If we did want to reference the resolved value , then we would do something like:

```typescript
const value = await eml.getSurveyOccurrenceSubmission(sampleReq.data_package_id, dbConnectionObj);

// do something with value here
```

NickPhura:
@cgarrettjones  Added a few comments. They could be addressed now, or we can merge this PR and address them after.  None of my comments are crucial, just a few fairly small improvements/notes.

Looks great in general though!

api/src/paths/dwc/eml.test.ts
```diff
@@ -0,0 +1,662 @@
...
+        { ...sampleReq, body: { ...sampleReq.body, data_package_id: null } },
+        (null as unknown) as any,
+        (null as unknown) as any
+      );
+      expect.fail();
+    } catch (actualError) {
+      expect((actualError as CustomError).status).to.equal(400);
+      expect((actualError as CustomError).message).to.equal('Missing required body param `data_package_id`.');
```
NickPhura, line 45: Same as the above comment, most of our existing tests check status before message.  But I want to start changing it so that the message assertion comes before the status assertion.  Because otherwise, if the wrong error is thrown (like a 500), all you get in the log is "400 !== 500" which is super unhelpful when trying to debug the test, since there might be multiple places where a 500 could be thrown from.

anissa-agahchen (approved):
Looks good, the only thing I am wondering about it whether you need to rebase this branch, as the commit history is quite long.

PR #679, opened 2021-12-16
View user details

KjartanE:
# Overview

## Links to jira tickets

- {Include links to all of the applicable Jira tickets}

- {List all the relevant changes. If there are many changes across many Jira tickets, organize the changes by Jira ticket}

- [ ] Version change

## Checklist

### General

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

## How Has This Been Tested?

## Screenshots


api/src/queries/project-participation/project-participation-queries.ts
```diff
@@ -3,6 +3,51 @@ import { getLogger } from '../../utils/logger';
...
+  defaultLog.debug({
+    label: 'getAllUserProjectsSQL',
+    message: 'params',
+    userId
+  });
+
+  if (!userId) {
+    return null;
+  }
+  console.log('INSIDE THE SQL CREATION ');
```
NickPhura: nit: remove

api/src/paths/user/{userId}/get-projects.ts
```diff
@@ -0,0 +1,145 @@
+import { RequestHandler } from 'express';
```
NickPhura: (Regarding this whole file)

Similar to the previous endpoint, lets change this to something like:
`paths/user/{userId}/projects/get`

Alternatively, it could also just be:
`paths/user/{userId}/get`
And in this case, it could be a general `get` for that user, and could return a whole packet of info about them (user details, projects they are one, etc).  Depending on what information is needed on the user info page you added, this might be an option.  But it might also be premature if we don't need other user information yet.

app/src/features/admin/users/UsersDetailPage.tsx
```diff
@@ -0,0 +1,60 @@
...
+  );
+
+  const getUser = () => {
+    if (userDetails !== undefined) {
+      setSelectedUser(userDetails);
+      window.localStorage.setItem('selectedUser', JSON.stringify(userDetails));
+    } else {
+      setSelectedUser(JSON.parse(window.localStorage.getItem('selectedUser') || ''));
+    }
+  };
```
NickPhura: I don't think we need to use local storage here.  Any data this component needs should either be passed in as props, or just re-fetched from the API.

app/src/hooks/api/useUserApi.ts
```diff
@@ -19,6 +19,17 @@ const useUserApi = (axios: AxiosInstance) => {
...
   };
 
+  /**
+   * Get user from userId
+   *
+   * @param {number} userId
+   * @return {*}  {Promise<IGetUserResponse>}
+   */
+  const getUserById = async (userId: number): Promise<IGetUserResponse> => {
+    const { data } = await axios.get(`/api/user/${userId}/get`, { params: { userId: userId } });
```
NickPhura: You don't need the parms object.  The API will extract the userId value directly from the URL (because the url matches the path with the `{userId}` folder which it uses to recognize variables in the URL.

```suggestion
    const { data } = await axios.get(`/api/user/${userId}/get`);
```

(Also slightly confusing on Axios's part, but that params object adds items to the query string.  So what you had originally would actually produce: `/api/user/123/get?userId=123`)

app/src/features/admin/users/UsersDetailHeader.tsx
```diff
@@ -0,0 +1,197 @@
...
+        <Box display="flex" justifyContent="space-between" alignItems="flex-start">
+          <Box pb={3}>
+            <Box mb={1.5} display="flex">
+              <Typography className={classes.spacingRight} variant="h1">
+                User - <span className={classes.projectTitle}>{userDetails.user_identifier}</span>
+              </Typography>
+            </Box>
+            <Box mb={0.75} display="flex" alignItems="center">
+              <Typography component="span" variant="subtitle1" color="textSecondary">
+                &nbsp;&nbsp;{userDetails.role_names[0]}
```
NickPhura: Are the 2 spaces needed? Probably just remove.  If an indent is desired then we can add some margin or padding instead. (but as always, do whatever Jeremy says when it comes to UI stuff)

NickPhura (approved):

PR #741, opened 2022-04-19
Techdebt

anissa-agahchen:
# Overview
* remove default logs
* fix tests
* fix code smells


api/src/paths/public/project/{projectId}/view.ts
```diff
@@ -300,7 +300,7 @@ export function getPublicProjectForView(): RequestHandler {
       const projectService = new ProjectService(connection);
 
       const result = await projectService.getPublicProjectById(Number(req.params.projectId));
-
+      console.log('result is: ', result);
```
NickPhura: 👀

NickPhura (approved):

NickPhura:
Just noticing 1 issue with the public view:
image

vs the admin view:
image

Activities and Project Type aren't rendering correctly on the public page.

Looks like the response from the API is returning the wrong values for those 2 fields:
image

NickPhura (changes requested):
See my other comment (forgot to write it here originally)

NickPhura:
> Just noticing 1 issue with the public view: image
> 
> vs the admin view: image
> 
> Activities and Project Type aren't rendering correctly on the public page.
> 
> Looks like the response from the API is returning the wrong values for those 2 fields: image

Pushed up a commit to address this

NickPhura (dismissed):

anissa-agahchen:
@NickPhura, thank you for the fix.

KjartanE (approved):
Looks good!

PR #746, opened 2022-04-20
BHBC-1688: Add survey radio field

NickPhura:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1688

Add new radio field to survey.
- Field is required (user must pick yes or no)

Bonus:
Add new git action to write openshift URLs to the PR.

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


NickPhura:
@jeznorth Wasn't 100% sure what you wanted project View UI to look like for the new field, so maybe double check it looks okay.

## View:
image


## Create/edit:
image

api/src/paths/project/{projectId}/survey/create.ts
```diff
@@ -100,6 +102,10 @@ POST.apiDoc = {
             ecological_season_id: {
               type: 'number'
             },
+            surveyed_all_areas: {
+              type: 'string',
+              enum: ['true', 'false']
```
curtisupshall, line 107: Any particular reason why we're encoding `surveyed_all_areas` as an enum instead of a boolean?

curtisupshall (dismissed):
Looks good! Like the changes to the pull request template 👍

api/src/paths/project/{projectId}/survey/create.ts
NickPhura: Its mainly because the MUI radio controls don't support boolean values, as dumb as that is. (Not sure if this is how all radio buttons work, or its an MUI issue)

So at some point in the process we have to convert from string to boolean when saving, and vice versa when fetching. I just landed on doing it in the API, mainly because converting on the frontend seemed like more effort and we don't have a pattern in place currently to handle these kinds of conversions on the frontend (like a model to translate fields, etc). Something we could look into though.

curtisupshall (approved):

PR #761, opened 2022-05-02
BHBC-1724: Fix edit dialog control clipping.

curtisupshall:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1724

When in edit mode, first field title is clipped.

### DIFF
--- a/app/src/themes/appTheme.ts
+++ b/app/src/themes/appTheme.ts
@@ -85,11 +85,6 @@ const appTheme = createMuiTheme({
         width: '800px'
       }
     },
-    MuiDialogContent: {
-      root: {
-        paddingTop: 0
-      }
-    },
     MuiDialogTitle: {
       root: {
         padding: '20px 24px'


NickPhura:
There is one instance of `MuiDialogContent-root` being set in `Header.tsx`.  I wonder if that interacts with your change here at all? Now that padding isn't set to 0 in the theme, is the override in Header.tsx needed? Or possibly its totally unrelated?

curtisupshall:
> There is one instance of `MuiDialogContent-root` being set in `Header.tsx`. I wonder if that interacts with your change here at all? Now that padding isn't set to 0 in the theme, is the override in Header.tsx needed? Or possibly its totally unrelated?

Thanks for spotting! I'll confirm that.

curtisupshall:
> > There is one instance of `MuiDialogContent-root` being set in `Header.tsx`. I wonder if that interacts with your change here at all? Now that padding isn't set to 0 in the theme, is the override in Header.tsx needed? Or possibly its totally unrelated?
> 
> Thanks for spotting! I'll confirm that.

`Header.tsx` style is not affected by this change.

anissa-agahchen (approved):

PR #772, opened 2022-05-09
BHBC-1736: Fix Read More/Read Less idempotence bug

curtisupshall:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1736

Tightens qualification for string truncating in the `ReadMoreField` component to avoid an issue where pressing **Read More**/**Read Less** results in no change.


anissa-agahchen (dismissed):

app/src/components/fields/ReadMoreField.tsx
```diff
@@ -17,7 +17,15 @@ export interface IReadMoreFieldProps {
 export const ReadMoreField: React.FC<IReadMoreFieldProps> = (props) => {
   const { text, maxCharLength } = props;
 
-  const [isTruncatedText, setIsTruncatedText] = useState(text?.length > maxCharLength);
+  /*
+    Determines whether or not the given body of text will be truncated based on
+    the max character length.
+  */
+  const willTruncateText = (body: string): boolean => {
+    return body?.trim().length > maxCharLength || false;
```
NickPhura: We pass in `text` but reference it as `body` here. Should this say `text` instead?

curtisupshall: I didn't want to shadow the varable `text` from L19. Maybe we can simply call `t`?

NickPhura: > I didn't want to shadow the varable text from L19

Right, yeah.

> Maybe we can simply call t?

Ideally not just a 1 letter variable, even when the function is short. Maybe `content`?

NickPhura (approved):

PR #778, opened 2022-05-11
BHBC-1740: Replace Moment with Luxon

curtisupshall:
# Overview

https://quartech.atlassian.net/browse/BHBC-1740

[MomentJS](https://www.npmjs.com/package/moment) is officially deprecated and the recommended successor is [Luxon](https://www.npmjs.com/package/luxon)


api/src/models/project-view.ts
```diff
@@ -51,7 +52,7 @@ export class GetProjectData {
     this.completion_status =
       (projectData &&
         projectData.end_date &&
-        moment(projectData.end_date).endOf('day').isBefore(moment()) &&
+        DateTime.fromJSDate(new Date(projectData.end_date)).endOf('day') < DateTime.now() &&
```
NickPhura, line 54: Any reason to use one over the other in this case?
`DateTime.fromJSDate(new Date(projectData.end_date))` vs `DateTime.fromISO(projectData.end_date)`

I haven't spent too much time on Luxon yet, but it seems like the latter would do the trick?

PR #781, opened 2022-05-16
Fixed horizontal autoscaler and consolidate deployment files

rstens:
# Overview

Fixing an issue with the horizontal autoscaler definition when deploying. The deploy code referenced an older version of the api for the scaler, therefore it was not deploying (pods were created all right). This would mean that we do not get that sweet auto-tuning that kubernetes can do for us.

Also consolidated some of the changes I made for N8N.

### DIFF
--- a/containers/n8n/n8n.bc.yaml
+++ b/containers/n8n/n8n.bc.yaml
@@ -1,12 +1,12 @@
 ---
 kind: Template
-apiVersion: v1
+apiVersion: template.openshift.io/v1
 labels:
   app: n8n
   build: "${NAME}"
   template: "${NAME}-bc-template"
 metadata:
-  name: "${NAME}"
+  name: n8n
 objects:
   - kind: ImageStream
     apiVersion: v1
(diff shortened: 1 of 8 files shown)


containers/n8n/n8n.bc.yaml
```diff
@@ -44,14 +44,18 @@ objects:
           ARG N8N_VERSION
           ENV HOME=/opt/app-root/src \
               TZ=America/Vancouver
+          USER root
```
NickPhura, line 47: (not his line specifically) Would it be useful to move this whole dockerfile definition into its own file, rather than writing it inline here?

NickPhura (dismissed):

containers/n8n/n8n.bc.yaml
rstens: It is easier to iterate settings this way, you can do it all on openshift. That was the only reason.
The dockerfile for this is not useful for anything but openshift though. For local we'd just use n8n/n8n:latest.

NickPhura (dismissed):

NickPhura (approved):

PR #792, opened 2022-06-13
BHBC-1796-1810 : bug fix - xlsx observation upload not completing

anissa-agahchen:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1796
https://quartech.atlassian.net/browse/BHBC-1810

Notable changes:
- transform xlsx step is faster
- new endpoint for the API to manage validate, transform and scrape

### The problem:

the UTM values should not have decimal points

There are three consecutive calls made from the app to process a submission: validate, transform, scrape.  The weak point was that since the file had a lot of rows, the last step of the transform that removes duplicates take a long time.  And when the API takes longer than 30 seconds to respond, the APP fails.

### The fix:
1. check the UTM values in the input file.  Down the road, this would be a good one to create a validation for. the format of the UTM values

2. Re-wrote the function that removes duplicates so that it is much faster

3. Introduced a new end-point xlsx/process that manages all the steps: validate, transform, and scrape.  The front-end now just makes one call.

4. The workflow for processing a submission now sends a success response back to the APP once the validation step is complete, and continues to process the submission in the background.  This way the APP doesn’t hang waiting.  This also means that you won’t see the points on the map until the processing is finished.

### To test:

Upload a file

If the file is large, give it a minute before refreshing the screen and you will see the points on the map.

Use this file to test: 
[modified_Moose_Aerial_StratifiedRandomBlock_Composition_Recruitment_Survey_2.5_Teslin2020DP.xlsx](https://github.com/bcgov/biohubbc/files/8920326/modified_Moose_Aerial_StratifiedRandomBlock_Composition_Recruitment_Survey_2.5_Teslin2020DP.xlsx)

### DIFF
--- a/api/src/paths/dwc/scrape-occurrences.ts
+++ b/api/src/paths/dwc/scrape-occurrences.ts
@@ -215,9 +215,7 @@ export const uploadScrapedOccurrence = async (
   scrapedOccurrence: PostOccurrence,
   connection: IDBConnection
 ) => {
-  console.log('Start - inserting occurrences in the occurrence table', new Date());
   const sqlStatement = queries.occurrence.postOccurrenceSQL(occurrenceSubmissionId, scrapedOccurrence);
-  console.log('End - inserting occurrences in the occurrence table', new Date());
 
   if (!sqlStatement) {
     throw new HTTP400('Failed to build SQL post statement');
--- a/api/src/paths/dwc/scrape-occurrences.ts
+++ b/api/src/paths/dwc/scrape-occurrences.ts
@@ -68,6 +68,7 @@ POST.apiDoc = {
         'application/json': {
           schema: {
             type: 'object',
+            required: ['status', 'reason'],
             properties: {
               status: {
                 type: 'string'
--- a/api/src/paths/dwc/scrape-occurrences.ts
+++ b/api/src/paths/dwc/scrape-occurrences.ts
@@ -140,8 +140,6 @@ export function scrapeAndUploadOccurrences(): RequestHandler {
         vernacularNameHeader
       } = getHeadersAndRowsFromFile(dwcArchive);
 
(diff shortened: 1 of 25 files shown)


api/src/paths/dwc/scrape-occurrences.ts
```diff
@@ -63,7 +63,22 @@ POST.apiDoc = {
   },
   responses: {
     200: {
-      description: 'Successfully scraped and uploaded occurrence information.'
+      description: 'Successfully scraped and uploaded occurrence information.',
+      content: {
+        'application/json': {
+          schema: {
+            type: 'object',
```
KjartanE, line 70: does it make sense to make these required?

app/src/features/surveys/view/SurveyObservations.tsx
```diff
@@ -78,10 +76,10 @@ const SurveyObservations: React.FC<ISurveyObservationsProps> = (props) => {
             return;
           }
 
-          if (config?.N8N_HOST) {
-            biohubApi.n8n.initiateOccurrenceSubmissionProcessing(projectId, result.submissionId, file.type);
-            return;
-          }
+          // if (config?.N8N_HOST) {
+          //   biohubApi.n8n.initiateOccurrenceSubmissionProcessing(projectId, result.submissionId, file.type);
+          //   return;
```
KjartanE, line 81: Can this be deleted?

KjartanE (commented):
App works properly and as intended, small nits

api/src/paths/dwc/scrape-occurrences.ts
anissa-agahchen: yes

KjartanE (approved):
Looks good works well!

NickPhura (approved):

PR #793, opened 2022-06-15
BHBC-1775, BHBC-1809: Templates in prod are still downloaded from S3 Dev

curtisupshall:
# Overview

Currently, templates in the test/production environment are still downloaded from S3 Dev. These templates should be retrieved from the production instance.

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1775
https://quartech.atlassian.net/browse/BHBC-1809

 - Add s3 public hsot URL to frontend env vars; Replace hardcoded URLs.
 - API response validation bug fixes

### DIFF
--- a/env_config/env.docker
+++ b/env_config/env.docker
@@ -142,3 +142,8 @@ GCNOTIFY_SMS_URL=https://api.notification.canada.ca/v2/notifications/sms
 # Taxonomy Elasticsearch Platform API
 # ------------------------------------------------------------------------------
 ELASTICSEARCH_URL=https://elasticsearch-af2668-dev.apps.silver.devops.gov.bc.ca
+
+# ------------------------------------------------------------------------------
+# S3 Public Host URL
+# ------------------------------------------------------------------------------
+S3_PUBLIC_HOST_URL=https://nrs.objectstore.gov.bc.ca/gblhvt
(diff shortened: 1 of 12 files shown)


env_config/env.docker
```diff
@@ -142,3 +142,8 @@ GCNOTIFY_SMS_URL=https://api.notification.canada.ca/v2/notifications/sms
 # Taxonomy Elasticsearch Platform API
 # ------------------------------------------------------------------------------
 ELASTICSEARCH_URL=https://elasticsearch-af2668-dev.apps.silver.devops.gov.bc.ca
+
+# ------------------------------------------------------------------------------
+# S3 Public Host URL
```
KjartanE: Is this the same as line #89 (OBJECT_STORE_URL)

NickPhura (commented):
Kjartan pointed out some changes we can make to reduce redundancy (which I hadn't considered when we were chatting about the approach for this ticket). We can discuss in discord whenever.

anissa-agahchen:
@curtisupshall your appdeploy pod is failing.  I think you can fix it by dropping the pods, and then retriggering a build with an empty commit.

NickPhura:
There is a funny issue happening when you click the links in openshift.  Because the env vars don't include `https://` react is actually opening the urls relative to the app url. 

I'm thinking the simplest solution for now is to just wrap those URLs in the resource page with that ensureProtocol function.

KjartanE (approved):
Looks Great!

NickPhura (approved):

PR #803, opened 2022-07-08
BHBC-1201: Push project/survey metadata (EML) to backbone on project/survey create/update

NickPhura:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1201

Add metadata submission function to build DwCA archive with just EML meta, and submit to the backbone.
 - Also preemptively added a function to build the full DwCA archive and submit it, though it currently doesn't fetch/include the occurrence data, as that will be part of a future ticket.  

Call metadata submission function from all endpoints that involve creating or update projects or surveys.  

Added new env var so we can update the backbone intake path, if it changes.
Added new env var so we can disable/enable submitting to the backbone (defaulting to false currently)
 - Will need update your `.env`

## Testing
You will need to run both SIMS and Platform at the same time. You will have to make some changes to one or both of their `.env` files.
- There are a few env vars in SIMS that need to be populated (towards the top, they reference `backbone`).
- You will need to change the env vars mentioned in the comment block at the top of `env.docker`, mainly ports, so that SIMS and Platform dont reference the same ports. 

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/openshift/api.dc.yaml
```diff
@@ -25,7 +25,14 @@ parameters:
...
     value: ''
   - name: BACKBONE_API_HOST
+    required: true
     description: API host for BioHub Platform Backbone
+  - name: BACKBONE_INTAKE_PATH
+    required: true
+    description: API path for BioHub Platform Backbone DwCA submission intake endpoint
+  - name: BACKBONE_INTAKE_ENABLED
+    required: true
+    description: Controls whether or not SIMS will submit DwCA datasets to teh BioHub Platform Backbone
```
curtisupshall: Spelling error

api/src/paths/project/{projectId}/funding-sources/{pfsId}/delete.ts
```diff
@@ -32,13 +33,14 @@ DELETE.apiDoc = deleteFundingSourceApiDocObject(
 
 export function deleteFundingSource(): RequestHandler {
   return async (req, res) => {
-    defaultLog.debug({ label: 'Delete project funding source', message: 'params', req_params: req.params });
+    const projectId = Number(req.params.projectId);
+    const pfsId = Number(req.params.pfsId);
```
curtisupshall, line 37: Is this shorthand for Platform Service ID?

curtisupshall (dismissed):
It looks great! 🏝️

api/src/paths/project/{projectId}/funding-sources/{pfsId}/delete.ts
NickPhura: It stands for `project funding source id`. Some of this code could use a bit of a refactor/cleanup still, but I was just touching the bits that directly related to my changes.

api/openshift/api.dc.yaml
NickPhura: Fixed in latest commit

curtisupshall (dismissed):
⛱️

curtisupshall (approved):
Pulled down and tested a few scenarios. Looks great!

PR #813, opened 2022-09-07
BHBC-1925

cgarrettjones:
# Overview

- adjusted permit model

## Links to Jira tickets

- [BHBC-1925](https://quartech.atlassian.net/browse/BHBC-1925)

- adjusted permit model for model 1.2
- adjusted db function project deletion routines
- added smoke test file

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


curtisupshall (dismissed):
Looks good to my eyes. 👍

AlfredRosenthal:
This all looks good. I did try running it locally just to see what happens and I think the removal of `project_id` from the `permit` table means we'll have to update a few SQL statements that are still expecting that to exist. That might have been the blow up you were seeing when you were trying to run things @cgarrettjones.

api/src/services/eml-service.ts
```diff
@@ -409,37 +409,39 @@ export class EmlService extends DBService {
...
-            metadata: {
-              permits: {
-                permit: { permitType: item.permit.permit_type, permitNumber: item.permit.permit_number }
-              }
-            }
-          });
-        }
-      });
-    }
+    // TODO add back when survey supports permits
```
AlfredRosenthal, line 412: should these be added back now?

api/src/services/permit-service.test.ts
```diff
@@ -1,313 +1,10 @@
...
-
-      sinon.stub(permit_queries, 'postPermitNoSamplingSQL').returns(SQL`some`);
-
-      const permitService = new PermitService(mockDBConnection);
-
-      const res = await permitService.insertNoSamplePermit(permitData, coordinatorData);
-
-      expect(res).to.equal(12);
-    });
+  it('TODO', () => {
```
AlfredRosenthal: could be removed

api/src/services/survey-service.ts
```diff
@@ -491,31 +489,52 @@ export class SurveyService extends DBService {
...
-    await this.unassociatePermitFromSurvey(surveyId);
+  async updateSurveyPermitData(surveyId: number, surveyData: PutSurveyObject) {
+    const permitService = new PermitService(this.connection);
 
-    if (!surveyData.permit.permit_number) {
-      return;
+    // Get existing permits
+    const existingPermits = await permitService.getPermitBySurveyId(surveyId);
+
+    // Determine if any existing permits need to be deleted (are not included in the incoming permits array)
```
AlfredRosenthal: comment is a little confusing maybe change to
'Incoming permits array is our source of truth, find and remove any hanging/ leftover permits'

NickPhura: Took a stab at updating all of these comments to try and be more descriptive, not sure if its better or just different :)

api/src/services/permit-service.test.ts
NickPhura: Removed, and added real tests (was just a placeholder so as not to forget)

api/src/services/eml-service.ts
NickPhura: we may never add them back tbh, as its questionable if permits belong in the EML at all.  But not certain yet.

api/src/services/survey-service.ts
AlfredRosenthal: I think it's better!

AlfredRosenthal (approved):
🐶

PR #836, opened 2022-10-28
BHBC 2009

AlfredRosenthal:
# Overview

## Links to Jira tickets

[BHBC-2009](https://quartech.atlassian.net/browse/BHBC-2009?atlOrigin=eyJpIjoiNzQxNmU2MDYwMmZlNDFmMTlmNTdiNjVmMWEwMjMxNTgiLCJwIjoiaiJ9)

Added a function wrapper for the database class to catch all exceptions thrown.
To test: 
- create a project
- create a survey
- open survey page in two tabs
- make changes and save in one tab
- make changes and save in second tab
- see stale project data error instead of 'Unexpected Error'

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/database/db.test.ts
```diff
@@ -114,17 +115,24 @@ describe('db', () => {
...
-              expectedError = error as Error;
+              expectedError = error as ApiExecuteSQLError;
             }
 
-            expect(expectedError.message).to.equal('DBPool is not initialized');
+            expect(expectedError.message).to.equal('Failed to execute SQL');
 
+            expect(expectedError.errors?.length).to.be.greaterThan(0);
+            expectedError.errors?.forEach((item) => {
+              expect(item instanceof Error).to.be.true;
```
NickPhura: nit: can we change this to use .instanceof? expect(blah).to.be.instanceof(stuff)

Just to match the other tests, though I think there is one other test file that also does it this way and could be updated.

NickPhura:
@AlfredRosenthal Looks like this is just failing because of a corrupt migration table. You probably just need to purge the pr artifacts in openshift, and then re-run the pr deploy, and then it will deploy using that fixed file name from dev.

KjartanE (approved):
🏆

PR #845, opened 2022-11-15
Bhbc 2045 secure documents prototype

anissa-agahchen:
# Overview

## Links to Jira tickets

-WIP stable chunk of code that can be merged into the secure-documents-prototype


api/src/models/project-survey-attachments.ts
```diff
@@ -67,15 +76,18 @@ export class GetReportAttachmentMetadata {
   description: string;
   year_published: number;
   revision_count: number;
+  security_review_timestamp: string;
```
curtisupshall, line 79: Let's make a `security` property which stores `security_review_timestamp` and `security_reasons`.

app/src/components/dialog/ViewFileWithDetailsDialog.tsx
```diff
@@ -48,27 +52,28 @@ const useStyles = makeStyles((theme: Theme) => ({
...
 
 /**
  * General information content for a project.
  *
  * @return {*}
  */
-const ViewFileWithMetaDialog: React.FC<IViewFileWithMetaDialogProps> = (props) => {
+const ViewFileWithDetailsDialog: React.FC<IViewFileWithDetailsDialogProps> = (props) => {
   const classes = useStyles();
-  const { reportMetaData } = props;
```
curtisupshall, line 68: Add this back in?

app/src/components/dialog/ViewFileWithDetailsDialog.tsx
```diff
@@ -196,28 +169,36 @@ const ViewFileWithMetaDialog: React.FC<IViewFileWithMetaDialogProps> = (props) =
...
-                    <TableCell></TableCell>
-                  </TableRow>
+                  {props.reportMetaData?.security_reasons &&
+                    props.reportMetaData?.security_reasons?.length > 0 &&
+                    props.reportMetaData?.security_reasons?.map((row, index) => {
+                      return (
+                        <TableRow key={`${row.category}-${index}`}>
+                          <TableCell>{row.category}</TableCell>
+                          <TableCell>
+                            <Typography style={{ fontWeight: 700 }}>{row.reason}</Typography>
```
curtisupshall, line 182: Make this a class?

app/src/interfaces/useProjectApi.interface.ts
```diff
@@ -295,13 +296,22 @@ export interface IGetReportMetaData {
   last_modified: string;
   revision_count: number;
   authors: IGetReportAuthors[];
+  security_reasons: IGetSecurityReasons[];
```
curtisupshall, line 299: Make this a `security` property

app/src/interfaces/useSurveyApi.interface.ts
```diff
@@ -170,6 +170,7 @@ export interface IGetSurveyAttachment {
   lastModified: string;
   size: number;
   securityToken: any;
+  securityReviewTimestamp: string;
```
curtisupshall, line 173: make security parameter

app/src/interfaces/useProjectApi.interface.ts
```diff
@@ -14,6 +14,7 @@ export interface IGetProjectAttachment {
   lastModified: string;
   size: number;
   securityToken: string;
+  securityReviewTimestamp: string;
```
curtisupshall, line 17: security parameter

app/src/interfaces/useSurveyApi.interface.ts
NickPhura: Might need to elaborate on what you mean.

I think this looks correct, unless I'm forgetting something. The database column this is mapping to is `security_review_timestamp`

NickPhura: Oh, unless you mean putting the security stuff in its own object like we were talking about? That definitely makes sense when it comes to the list of all security rules applied to an attachment, but this particular column is part of the attachment table, so it makes sense to live in this object.

I think ultimately, we still want to maintain a separate object that represents a single attachment record, and a separate object that represents the rules applied to a single attachment. That way we can fetch them separately via services, etc. And then our attachment related endpoints can mix and match the service calls to return attachment(s) with or without security rules alongside them.

PR #866, opened 2022-11-22
Bhbc 2047

anissa-agahchen:
# Overview

## Links to Jira tickets

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/paths/project/{projectId}/attachments/security/add.ts
```diff
@@ -54,11 +54,20 @@ POST.apiDoc = {
...
               }
             },
-            attachment_ids: {
+            attachments: {
               type: 'array',
               items: {
-                type: 'number',
-                minimum: 1
+                type: 'object',
+                minimum: 1,
```
NickPhura: nit: remove minimum, I dont think this is doing anything at this level, might just be a remnant from the previous version

api/src/repositories/attachment-repository.ts
```diff
@@ -109,11 +109,82 @@ export class AttachmentRepository extends BaseRepository {
     return response.rows;
   }
 
+  // TODO functions need to handle duplicate keys (re adding existing security reasons)
   async addSecurityToAttachments(securityIds: number[], attachmentId: number): Promise<void> {
-    // TODO
+    const insertStatement = SQL`
```
NickPhura, line 194: Cant remember the syntax, but wondering if there is a cleaner way to do these array inserts using knex? Something we can tech debt later as well.

NickPhura: That being said, I'm positive we haven't been consistent on when to use knex and when not, so probably we have a mixture of both patterns through the repo

app/src/features/projects/view/ProjectAttachments.tsx
```diff
@@ -123,8 +138,14 @@ const ProjectAttachments: React.FC<IProjectAttachmentsProps> = () => {
 
       <SecurityDialog
         open={securityDialogOpen}
-        onAccept={() => {
-          setSecurityDialogOpen(false);
+        onAccept={(securityReasons) => {
+          if (selectedAttachmentRows.length > 0) {
+            // formik form is retuning array of strings not numbers if printed out in console
+            // linter wrongly believes formik to be number[] so wrapped map in string to force values into number[]
```
NickPhura, line 144: Ah yeah, because the value attribute of a checkbox either doesn't accept numbers or always converts them to strings (either way they only support strings, which is dumb).

api/src/paths/project/{projectId}/attachments/security/add.ts
AlfredRosenthal: removed

api/src/repositories/attachment-repository.ts
AlfredRosenthal: It works, it is probably best to move forward for now. I've created a tech debt ticket to track these changes

AlfredRosenthal (approved):
🐶

PR #873, opened 2022-11-23
BHBC-1097: Secure documents prototype

curtisupshall:
# Overview
Show state of documents sp that data administrators can understand whether or not a document needs to be reviewed OR if it is secured / unsecured from public consumption.

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1097


AlfredRosenthal (approved):
🐠

PR #886, opened 2022-12-05
BHBC-2086 - Don't Show Expired Taxonomic Codes

curtisupshall:
# Overview

New data submitted to SIMS to should use **current** taxonomic codes so new data isn’t associated with codes that are no longer in use.

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-2086

- Modified species search in the Taxonomy service to include a boolean filter on records whose end date has passed;
- Added a new environment variable to deliniate the index name used to search for taxonomic codes in Elasticsearch;
- Enhanced JSDoc and type defs in the Taxonomy service;
- Made all private `TaxonomyService` methods public and prefixed them by "`_`".

## PR Checklist

### Code

- [x] My code uses types/interfaces to help describe values/parameters/etc, help ensure type safety, and improve readability

### Style

- [x] My code follows the established style conventions
- [x] My code uses native material-ui components/icons/conventions when possible

### Documentation

- [x] I have added/updated README's and related documentation, as needed

### Tests

- [x] I have added/updated unit tests for any code I've added/updated
- [x] I have added/updated the Postman requests/tests to account for any API endpoints I've added/updated

### Linting/Formatting

### SonarCloud

- [x] I have addressed all SonarCloud Bugs, Vulnerabilities, Security Hotspots, and Code Smells


api/openshift/api.dc.yaml
```diff
@@ -48,6 +48,10 @@ parameters:
     description: Platform Elasticsearch URL
     required: true
     value: 'https://elasticsearch-af2668-dev.apps.silver.devops.gov.bc.ca'
+  - name: ELASTICSEARCH_TAXONOMY_INDEX
+    description: Platform Elasticsearch Taxonomy Index
+    required: true
+    value: 'taxonomy_2.0.0'
```
NickPhura: nit: lets drop the default, just so we only ever define this in 1 spot (the top level config.js).

NickPhura: (Some of the other vars could use the same treatment, but lets leave that for a later tech debt ticket)

api/src/services/taxonomy-service.ts
```diff
@@ -1,43 +1,94 @@
...
+   * Searches the taxonomy Elasticsearch index by taxonomic code IDs
+   * @param ids The array of taxonomic code IDs
+   * @returns The response from Elasticsearch
+   */
   async getTaxonomyFromIds(ids: number[]) {
-    const response = await this.elasticSearch({
+    const response = await this._elasticSearch({
       query: {
         terms: {
           _id: ids
```
NickPhura, line 92: Is there any way to handle the filtering through some sort of elastic search concept, so we don't have to filter the results after fetching them?

api/openshift/api.dc.yaml
curtisupshall: Roger that; I removed `value: 'taxonomy_2.0.0'`. Do I have that right?

api/src/services/taxonomy-service.ts
```diff
@@ -1,43 +1,94 @@
...
+    }
+
+    return true;
+  };
+
+  /**
+   * Sanitizes species data retrieved from Elasticsearch.
+   * @param {SearchHit<ITaxonomySource>[]} data The data response fromEelasticsearch
+   * @returns {{ id: string, label: string }} An ID and label pair for each taxonomic code
+   */
```
NickPhura: 👍 for adding more jsdoc

NickPhura: Though, I might suggest using that [Document This](https://marketplace.visualstudio.com/items?itemName=oouo-diogo-perdigao.docthis) extension, which will populate the base info for you, if you're not already.

The return here is showing an object, rather than an array of objects.

api/src/services/taxonomy-service.ts
curtisupshall: There is actually - currently working on implementing it

AlfredRosenthal (approved):
🐯

NickPhura (approved):

PR #888, opened 2022-12-08
BHBC-2085

anissa-agahchen:
# Overview

## Links to Jira tickets
https://quartech.atlassian.net/browse/BHBC-2085

The work in this ticket is based on the assumption that the darwin_core_source column in the occurrence_submission is populated with the result of the transform as a JSON object.

- This code takes that object:
** finds all paths that contain a taxonID
** queries elastic search for the scientific and common name
** enriches the object at the same path with the scientific and common names
** returns the object

At a later point, this work will be merged with the transfom steps.

Note:  look at tests to see how the object transformations


api/src/utils/media/xlsx/transformation/xlsx-transformation.ts
```diff
@@ -58,7 +58,7 @@ export class XLSXTransformation {
   }
 
   /**
-   * Flattens the worksheet data into arrays of objects.
+   * Flattens the worksheet data into arrays of objects.par
```
AlfredRosenthal: .par?

api/src/services/validation-service.ts
```diff
@@ -135,6 +135,7 @@ export class ValidationService extends DBService {
       // template transformation
       await this.templateTransformation(submissionId, submissionPrep.xlsx, submissionPrep.s3InputKey, surveyId);
 
+      // insert template validated status
```
AlfredRosenthal: nit: next line is inserting transformed status?

api/src/utils/media/xlsx/transformation/xlsx-transformation.ts
anissa-agahchen: lol ... no idea what that's about

AlfredRosenthal (dismissed):
🐯

api/src/services/dwc-service.ts
```diff
@@ -0,0 +1,80 @@
...
+    this._packageId = options.packageId;
+
+    this.projectId = options.projectId;
+
+    this.taxonomyService = new TaxonomyService();
+
+    this.xml2jsBuilder = new xml2js.Builder({ renderOpts: { pretty: false } });
+  }
+
+  async enrichTaxonIDs(jsonObject: Record<any, any>): Promise<Record<any, any>> {
```
curtisupshall, line 50: Could we get some JS docs on this method?

api/src/services/taxonomy-service.ts
```diff
@@ -192,4 +192,47 @@ export class TaxonomyService {
 
     return response ? this._sanitizeSpeciesData(response.hits.hits) : [];
   }
+
+  private formatEnrichedData = (data: SearchHit<any>) => {
```
curtisupshall: I believe making service methods private makes mocking the method more difficult or even impossible in the service tests. I think we should prefix this method with an `_` and make it public, like other methods in the class

curtisupshall: Also, is it possible to be more specific with the typing on the `data` parameter? It should be `SearchHit<ITaxonomySource>` although I could be wrong

api/src/services/taxonomy-service.ts
```diff
@@ -192,4 +192,47 @@ export class TaxonomyService {
...
+    const scientific_name = [data._source.unit_name1, data._source.unit_name2, data._source.unit_name3]
+      .filter(Boolean)
+      .join(' ');
+    const english_name = data._source.english_name;
+
+    return { scientific_name, english_name };
+  };
+
+  async getEnrichedDataForSpeciesCode(code: string) {
```
curtisupshall: Nit: JS docs for these two new methods

api/src/services/dwc-service.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        item_with_depth_1: {
+          item_with_depth_2: { taxonID: 'M-OVCA' }
+        }
+      };
+
+      const enrichedJSON = await dwcService.enrichTaxonIDs(jsonObject);
+
+      expect(getEnrichedDataForSpeciesCodeStub).to.have.been.called;
+      expect(getEnrichedDataForSpeciesCodeStub).to.have.been.calledWith('M-OVCA');
+      expect(getEnrichedDataForSpeciesCodeStub).not.to.have.been.calledWith('M-OVCA1');
```
curtisupshall: Nit: I think the above `expect` asserts that this will always be true

curtisupshall (dismissed):
I think it looks good! Just a handful of nitpicks re: comments and type defs

app/src/features/surveys/view/SurveyObservations.tsx
```diff
@@ -106,7 +106,7 @@ export enum SUBMISSION_MESSAGE_TYPE {
   'FAILED_PREP_DWC_ARCHIVE' = 'Failed to prep DarwinCore Archive',
   'FAILED_PREP_XLSX' = 'Failed to prep XLSX',
   'FAILED_PERSIST_PARSE_ERRORS' = 'Failed to persist parse errors',
-  'FAILED_GET_VALIDATION_RULES' = 'Failed to get validation rules',
+  'FAILED_GET_VALIDATION_RULES' = 'Failed to get validation rules.  It is possible that your focal species is incorrect.',
```
curtisupshall: I don't know how I feel about this addition. This enum is just supposed to store error message types; I think suggestions such as "fix your focal species" should belong in the actual error message, not in the message type string - especially given that the phrasing here is also a bit subjective. Moreover, if we did want to change this enum value, we would need a corresponding migration as well.

NickPhura: Yeah makes sense.

I do think it adds additional information which is good, but it also could be misleading if the focal species isnt the issue (and you submitted the wrong template or something). The app side needs an update in general around how better to display error messages (rolling them up, having helper icons for more details, etc). So for now we can just leave it as it was, and a user may just need to reach out to Robin & Co for assistance (which they probably would need to anyways)

curtisupshall (approved):
⛄

PR #890, opened 2022-12-09
BHBC-2041: Key Validation Rule for Template Validation

curtisupshall:
# Overview

Adds a new validation rule that can check if one or more columns are unique (not duplicates) in order to validate keys.

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-2041

 - Added a new submission message error type (with corresponding migration)
 - Augmented CSV state to reflect key errors
 - Added validation rules that run on workbooks
 - Changed the Darwin Core Archive and Summary services to also run workbook validations
 - Added key-parent validation rule for workbooks
 - Fixed and added tests for new functions

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/utils/media/dwc/dwc-archive-file.ts
```diff
@@ -87,6 +87,16 @@ export class DWCArchive {
     }
   }
 
+  _workbookFromWorksheets() {
```
curtisupshall: Alternatively, we could modify this class to maintain a workbook, rather than maintaining a collection of sheets. Then we could ditch this helper method completely. Thoughts?

api/src/utils/media/xlsx/xlsx-file.ts
```diff
@@ -33,6 +38,32 @@ export class XLSXCSV {
...
   }
 
+  /**
+   * Determines the validity of a workbook belonging to a submission. Workbook validators
+   * are different in that they validate an entire submission, but return validations
+   * particular to worksheets within the workbook.
+   * @param validationSchemaParser
+   * @returns
+   */
+  isWorkbookValid(validationSchemaParser: ValidationSchemaParser): ICsvState[] {
```
curtisupshall: This JS doc comment pretty much sums up my justification for creating this method, and the added validation schema parser methods that go along with it. But essentially, `isWorkbookValid` is a fringe method that returns `ICsvState[]` for an entire submission, whereas the other two methods return `IMediaState` for an entire submission, or `ICsvState[]` for a single file in a submission.

api/src/utils/media/xlsx/validation/xlsx-validation.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * For a specified parent sheet, child sheet, and set of parent and child columns, adds an error on each cell in the
+ * child sheet whose key in the corresponding row belonging to the parent sheet cannot be found.
+ *
+ * Note: If the cell is empty, this check will be skipped. Use the `getRequiredFieldsValidator` validator to assert
+ * required fields.
+ *
+ * @param {KeyMatchValidatorConfig} [config]
+ * @return {*}  {CSVValidator}
```
curtisupshall: Just noticed that this JS doc return type is outdated; will address this

api/src/utils/media/xlsx/validation/xlsx-validation.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const parentSerializedRows = parentRowObjects.map(serializer);
+
+    // Add an error for each cell containing a dangling key reference in the child worksheet
+    childRowObjects
+      .map(serializer)
+      .map((serializedRow: string, rowIndex: number) => {
+        return !serializedRow || parentSerializedRows.includes(serializedRow) ? -1 : rowIndex;
+      })
+      .filter((rowIndex: number) => rowIndex >= 1)
```
NickPhura: Should this be `rowIndex >= 0` ?

api/src/utils/media/xlsx/validation/xlsx-validation.ts
```diff
@@ -0,0 +1,75 @@
...
+      .forEach((danglingRowIndex: number) => {
+        const mismatchedColumn =
+          column_names.find((columnName: string) => {
+            return parentRowObjects[danglingRowIndex][columnName] !== childRowObjects[danglingRowIndex][columnName];
+          }) || column_names[column_names.length - 1];
```
NickPhura: Just checking: this would return the first column that was incorrect, even if multiple were incorrect, or else it defaults to whatever the last column in the array is. I think it might be ideal if the error message indicated that the key (be it 1 column or multiple) is invalid, rather than trying to target just the bad one(s) Mainly thinking of how best to explain this issue to the end user clearly, which may need some input from Robin/Albert/Jer.

api/src/utils/media/dwc/dwc-archive-file.ts
NickPhura: Could do yeah. I can't remember the original thinking here, but it was probably because the DWC files are all separate csv files, so I converted them each into a worksheet, but never worried about combining them into a workbook since we never do anything at the workbook level (before now anyways).

api/src/services/validation-service.ts
```diff
@@ -310,7 +310,8 @@ export class ValidationService extends DBService {
       throw SubmissionErrorFromMessageType(SUBMISSION_MESSAGE_TYPE.INVALID_MEDIA);
     }
 
-    const csvState: ICsvState[] = file.isContentValid(parser);
+    const csvState: ICsvState[] = [...file.isWorkbookValid(parser), ...file.isContentValid(parser)];
```
NickPhura: I wonder if we need to spread `isWorkbookValid` here? Won't it have written any errors into the worksheets already, which will be accounted for in the return from `isContentValid`?

I guess the only issue is if there are no `isContentValid` rules defined, but there are `isWorkbookValid` rules. This is where merging them becomes necessary maybe. Assuming I'm following this all correctly.

NickPhura: Actually, what we probably need, is instead of returning anything from these 2 functions. We should just call them, and then get the csvStates afterwards from some separate function. Then it doesn't matter what validation rules are being run (if any), we can get the csvStates once at the end.

api/src/json-schema/validation-schema.ts
```diff
@@ -165,6 +182,40 @@ export const submissionValidationSchema = {
...
+          type: 'object',
+          required: ['child_worksheet_name', 'parent_worksheet_name', 'column_names'],
+          properties: {
+            name: {
+              type: 'string'
+            },
+            description: {
+              type: 'string'
+            },
+            required_files: {},
```
NickPhura: nit: remove

api/src/json-schema/validation-schema.ts
```diff
@@ -165,6 +182,40 @@ export const submissionValidationSchema = {
...
+        'Validates that this workbook submission file does not contain keys belonging to a child sheet that are missing in its parent sheet',
+      type: 'object',
+      properties: {
+        workbook_parent_child_key_match_validator: {
+          type: 'object',
+          required: ['child_worksheet_name', 'parent_worksheet_name', 'column_names'],
+          properties: {
+            name: {
+              type: 'string'
+            },
```
NickPhura: nit: remove. This part of the original schema isn't the best, but the `name` was meant to be the name of the column the rule applies to, which doesn't matter in this case since its captured in the column_names field below

api/src/utils/media/xlsx/validation/xlsx-validation.ts
curtisupshall: Oops, thought I was being clever by over skipping the header row - which is implicitly not included. Will change this to `rowIndex >= 0`

api/src/services/validation-service.ts
NickPhura: Just saw your comment here (https://github.com/bcgov/biohubbc/pull/890/files/0ccd6f87af42e88d9f5dcc906b6958c535a95c6b#r1044892317). Possibly my comments here are irrelevant.

api/src/utils/media/xlsx/validation/xlsx-validation.ts
curtisupshall: Yep that was the intent. Right, maybe we need to add a new method `addKeyErrors` or similar.

NickPhura:
Looks good! I might just add more inline comments to the validation function itself, to really make it clear what each step is doing, as it is doing a few different maps and filters, etc.

api/src/services/validation-service.ts
curtisupshall: Ah you're right. Merging them like this would case duplicates, since the CSV states from `isWorkbookValid` will get caried forward. Short of combing through each list of `ICsvState` and taking the state of a given CSV to be derived from one or the other, maybe this calls for the creation of `IWorkbookState`? 🤥

api/src/json-schema/validation-schema.ts
curtisupshall: Oh, I interpreted it as "the name of the parent-child-key relationship"; will remove

api/src/services/validation-service.ts
curtisupshall: > Actually, what we probably need, is instead of returning anything from these 2 functions. We should just call them, and then get the csvStates afterwards from some separate function. Then it doesn't matter what validation rules are being run (if any), we can get the csvStates once at the end.

Ahh yeah good call. That's what we're doing inside of the `isWorkbookValid`, anyway.

curtisupshall: > Actually, what we probably need, is instead of returning anything from these 2 functions. We should just call them, and then get the csvStates afterwards from some separate function. Then it doesn't matter what validation rules are being run (if any), we can get the csvStates once at the end.

This what I ended up opting for.

KjartanE (commented):
Looks good to go, just fix tests not running

AlfredRosenthal (dismissed):
Looks good!

👓

AlfredRosenthal (approved):
👍

PR #892, opened 2022-12-10
Bhbc 2040

AlfredRosenthal:
# Overview

## Links to Jira tickets

[BHBC-2040](https://quartech.atlassian.net/browse/BHBC-2040?atlOrigin=eyJpIjoiN2YzOWMxMmM0Mjc2NDk2Y2FhMjZhMmQ5MmFmMjhmYjAiLCJwIjoiaiJ9)

- added new template validation rule checking unique keys in a file
- added new validation rule to schema
- added new tests for new template validation rules

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/utils/media/csv/validation/csv-row-validator.ts
```diff
@@ -352,3 +352,55 @@ export const getValidFormatFieldsValidator = (config?: ColumnFormatValidatorConf
     return csvWorksheet;
   };
 };
+
+export type FileColumnUniqueValidatorConfig = {
+  file_column_unique_validator: {
+    key_columns: string[];
```
curtisupshall: If these are just the names of the keys (or equivalently, the names of the columns), maybe we could just rename this to `column_names`?

api/src/utils/media/csv/validation/csv-row-validator.ts
```diff
@@ -352,3 +352,55 @@ export const getValidFormatFieldsValidator = (config?: ColumnFormatValidatorConf
...
+    }
+
+    const keySet = new Set();
+    const rows = csvWorksheet.getRowObjects();
+    const lowercaseHeaders = csvWorksheet.getHeadersLowerCase();
+
+    const columnIndices = config.file_column_unique_validator.key_columns.map((column) =>
+      lowercaseHeaders.indexOf(column.toLocaleLowerCase())
+    );
+    if (columnIndices.indexOf(-1) !== -1) {
```
curtisupshall: Is there a more clear or obvious check that could assert that the worksheet or config columns don't line up? If not, maybe a comment could clarify what is happening here?

api/src/utils/media/csv/validation/csv-row-validator.ts
```diff
@@ -352,3 +352,55 @@ export const getValidFormatFieldsValidator = (config?: ColumnFormatValidatorConf
...
+      // worksheet or config columns don't line up
+      return csvWorksheet;
+    }
+
+    rows.forEach((row, rowIndex) => {
+      const key = config.file_column_unique_validator.key_columns.map((columnIndex) => row[columnIndex]).join(', ');
```
curtisupshall: Might want to sanitize any white space using `.map(s => s.trim())` before serializing

database/src/migrations/20221209131600_validation_rule_unique_error.ts
```diff
@@ -0,0 +1,46 @@
+import { Knex } from 'knex';
+
+const DB_SCHEMA = process.env.DB_SCHEMA;
+
+const summaryMessageTypes = [
+  {
+    name: 'Non Unique Key found',
```
curtisupshall: I think this string needs to match the enum value defined in `constraints/status.ts`, since we search the database using the name of an error in order to find its ID when inserting a submission error, in the same manner that Line 32 through Line 38 of this file uses.

curtisupshall (commented):
Looks good! Just had a handful of comments

api/src/utils/media/csv/validation/csv-row-validator.ts
AlfredRosenthal: I've added a comment that hopefully makes more sense. The check was done in this way because there can be multiple columns provided, so I was checking for a not found instead of looping through. Let me know if it makes more sense

anissa-agahchen (approved):

PR #913, opened 2023-01-13
BHBC-2145, 2087: Occurrence messaging refactoring; Submission message type for mismatched focal species

curtisupshall:
# Overview

Amends observation submission messaging.

## Links to Jira tickets

 - https://quartech.atlassian.net/browse/BHBC-2145
 - https://quartech.atlassian.net/browse/BHBC-2087

 - Refactors the way that submission messages are rendered by the SurveyOccurrences component
 - Clean-up around SurveyOccurrences component logic + API queries
 - Fixes a bug that persists a "Failed to get file from S3" submission message
 - Amends messaging related to template species not matching the focal species of a survey

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/repositories/survey-repository.ts
```diff
@@ -815,4 +897,100 @@ export class SurveyRepository extends BaseRepository {
...
+   * @return {*} {Promise<{ submissionId: number }>} Promise resolving the ID of the submission upon successful insertion
+   */
+  async insertSurveyOccurrenceSubmission(
+    submission: IObservationSubmissionInsertDetails
+  ): Promise<{ submissionId: number }> {
+    defaultLog.debug({ label: 'insertSurveyOccurrenceSubmission', submission });
+    const queryBuilder = getKnex()
+      .table('occurrence_submission')
+      .insert({
+        ['input_file_name']: submission.inputFileName,
```
NickPhura: Are the square brackets needed?

api/src/paths/xlsx/process.ts
```diff
@@ -111,10 +112,17 @@ export function processFile(): RequestHandler {
...
-      // process the resulting transformed dwc data
-      await validationService.processDWCFile(submissionId);
+
+      try {
+        // process the raw template data
+        await validationService.processXLSXFile(submissionId, surveyId);
+        // process the resulting transformed dwc data
+        await validationService.processDWCFile(submissionId);
+      } catch (error: any) {
+        if (!(error instanceof SubmissionError)) {
```
NickPhura, line 124: Might be good to add a comment here briefly explaining why we only throw non `SubmissionError` errors.

api/src/repositories/validation-repository.ts
```diff
@@ -28,31 +31,61 @@ export class ValidationRepository extends BaseRepository {
...
-        'template_methodology_species.validation',
-        'template_methodology_species.transform'
-      )
+      .select('template_methodology_species_id', 'wldtaxonomic_units_id', 'validation', 'transform')
       .from('template_methodology_species')
-      .where('template_methodology_species.template_id', templateRow.template_id)
-      .and.whereIn(
-        'template_methodology_species.wldtaxonomic_units_id',
-        (Array.isArray(surveySpecies) && surveySpecies) || [surveySpecies]
-      )
```
NickPhura, line 43: Are we not checking species anymore? Or was this removed by mistake?

NickPhura: Oh, I see it was moved below

api/src/repositories/validation-repository.ts
```diff
@@ -28,31 +31,61 @@ export class ValidationRepository extends BaseRepository {
...
       throw new HTTP400('Failed to query template methodology species table');
     }
 
-    return response.rows[0];
+    // Failure to find a template methodology species record for this template name and verion; Should yield a validation error.
+    if (response.rows.length === 0) {
+      throw failedToFindValidationRulesError;
+    }
+
+    // Some template methodology species records are found for this template name and version, but none are associated with this
```
NickPhura, line 71: 👍 for all the extra comments explaining the business logic here

api/src/repositories/survey-repository.test.ts
```diff
@@ -799,4 +799,62 @@ describe('SurveyRepository', () => {
...
+
+  describe('updateSurveyOccurrenceSubmission', () => {
+    it('todo', async () => {
+      throw new Error('TODO');
+    });
+  });
+
+  describe('deleteOccurrenceSubmission', () => {
+    it('todo', async () => {
+      throw new Error('TODO');
```
NickPhura: Just in case: todos

api/src/repositories/survey-repository.ts
curtisupshall: Just a stylistic choice; Agreebly, not needed.

api/src/paths/project/{projectId}/survey/{surveyId}/observation/submission/get.ts
```diff
@@ -73,12 +74,48 @@ GET.apiDoc = {
...
+                    messageStatus: {
+                      type: 'string',
+                      description: 'The resulting status of the submission as a consequence of the error'
+                    },
+                    messages: {
+                      type: 'array',
+                      description: 'The array of submission messages belonging to this type of message',
+                      items: {
+                        type: 'object',
+                        description: '', // TODO
```
NickPhura: just in case: todo

api/src/repositories/survey-repository.ts
NickPhura: Not a big deal either way then. I would maybe lean towards removing the brackets and the quotes then, just to simplify it a bit (especially in the context of knex, where things are already a bit more complicated)

api/src/paths/project/{projectId}/survey/{surveyId}/observation/submission/upload.ts
```diff
@@ -199,24 +199,20 @@ export const insertSurveyOccurrenceSubmission = async (
   source: string,
   inputFileName: string,
   connection: IDBConnection
-): Promise<any> => {
-  const insertSqlStatement = queries.survey.insertSurveyOccurrenceSubmissionSQL({
+): Promise<{ submissionId: number }> => {
```
KjartanE, line 202: Is there any reason to keep these functions in this endpoint? Could we move them to a service file?

KjartanE: ie `insertSurveyOccurrenceSubmission`, `updateSurveyOccurrenceSubmissionWithKey` Just call the main service function instead of hoping between fucntions

api/src/repositories/survey-repository.ts
```diff
@@ -345,6 +374,59 @@ export class SurveyRepository extends BaseRepository {
     return result;
   }
 
+  /**
+   * SQL query to get the list of messages for an occurrence submission.
+   *
+   * @param {number} submissionId
+   * @returns {SQLStatement} sql query object
```
KjartanE: Nit: Update return type

api/src/repositories/survey-repository.ts
```diff
@@ -345,6 +374,59 @@ export class SurveyRepository extends BaseRepository {
...
+      ON
+        smc.submission_message_class_id = smt.submission_message_class_id
+      WHERE
+        os.occurrence_submission_id = ${submissionId}
+      ORDER BY sm.submission_message_id;
+    `;
+
+    const response = await this.connection.sql<IOccurrenceSubmissionMessagesResponse>(sqlStatement);
+
+    if (!response?.rows) {
```
KjartanE, line 420: does this error? ![Array] will never be false. fix: `!response.rows.length`

curtisupshall: > ![Array] will never be false.

True, but `response.rows` is not guarenteed to be an array; The intent here to branch if `response` has `{ rows: undefined }` which would happen if the query fails. The reason we don't branch on `!response.rows.length` is because `[]` is an expected response

api/src/paths/project/{projectId}/survey/{surveyId}/observation/submission/upload.ts
curtisupshall: That's a fair question. We could certainly inline these functions - I don't think there's a strong case to have `insertSurveyOccurrenceSubmission` and `updateSurveyOccurrenceSubmissionWithKey` exist as separate functions, if only to make testing the endpoint simpler.

Granted, I think the reason that we don't call these two steps inside another "upload occurrence" service file method is because of the file upload to S3; We first need to insert the occurrence record in order to generate an S3 key, both to update the record and to upload the file. Consequently I don't think moving this logic to a service method provides much value because it would only be called by this endpoint.

KjartanE (approved):
👍

PR #919, opened 2023-01-21
BHBC-2160: Decorate DWCA Occurrences with Decimal Lat/Long

KjartanE:
# Overview

## Links to Jira tickets

- https://quartech.atlassian.net/browse/BHBC-2160

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/services/dwc-service.ts
```diff
@@ -90,4 +70,60 @@ export class DwCService extends DBService {
...
+    const pathsToPatch = JSONPath({
+      path: '$..[verbatimCoordinates]^',
+      json: jsonObject,
+      resultType: 'all'
+    });
+
+    const patchOperations: Operation[] = [];
+
+    await Promise.all(
+      pathsToPatch.map(async (item: any) => {
```
curtisupshall: I might be missing something, but it looks like all of the operations in this callback are synchronous. Instead of `pathsToPatch.map`, you could use `pathsToPatch.forEach`, and remove the outer `await Promise.all`

api/src/services/dwc-service.ts
```diff
@@ -90,4 +70,60 @@ export class DwCService extends DBService {
     // Apply patch operations
     return jsonpatch.applyPatch(jsonObject, patchOperations).newDocument;
   }
+
+  async decorateDWCASourceData(occurrenceSubmissionId: number): Promise<boolean> {
```
curtisupshall, line 81: Nit: JS doc

api/src/utils/spatial-utils.ts
```diff
@@ -158,3 +158,50 @@ export function generateGeometryCollectionSQL(geometry: Feature | Feature[]): SQ
 
   return sqlStatement;
 }
+
+export function utmToLatLng(zone: number, easting: number, northing: number): { lat: number; long: number } {
```
curtisupshall: Source on this would be helpful; Maybe we should draw some consensus on whether to use a library for this or not. Granted it's a well understood math operation, but some testing would be nice

api/src/repositories/occurrence-repository.ts
```diff
@@ -16,6 +16,7 @@ export interface IOccurrenceSubmission {
   input_file_name: string;
   output_key: string;
   output_file_name: string;
+  darwin_core_source: any;
```
curtisupshall: If this property remains as a JSON string (and not a JSON object) during the lifecycle of an `IOCcurrenceSubmission`, could we make this of type `string` for clarity?

api/src/services/dwc-service.ts
```diff
@@ -90,4 +70,60 @@ export class DwCService extends DBService {
...
+      json: jsonObject,
+      resultType: 'all'
+    });
+
+    const patchOperations: Operation[] = [];
+
+    await Promise.all(
+      pathsToPatch.map(async (item: any) => {
+        // eslint-disable-next-line no-prototype-builtins
+        if (item.value.hasOwnProperty('decimalLatitude') && item.value.hasOwnProperty('decimalLongitude')) {
```
curtisupshall: I think the linter makes a valid case: https://eslint.org/docs/latest/rules/no-prototype-builtins

You can ameliorate the complaint by calling `Object.prototype.hasOwnProperty`:

```suggestion
        if (
          Object.prototype.hasOwnProperty.call(item.value, 'decimalLatitude') &&
          Object.prototype.hasOwnProperty.call(item.value, 'decimalLongitude')
        ) {
```

curtisupshall (changes requested):
Couple of comments but looks great! +1 for project + survey seeder 😊🙌

api/src/services/dwc-service.ts
NickPhura: Agreed. Lets prefer avoiding disabling linter/formatter rules whenever possible.

api/src/services/dwc-service.ts
NickPhura: I think you are correct, I don't think any of this needs to be async.

api/src/services/dwc-service.ts
```diff
@@ -90,4 +70,60 @@ export class DwCService extends DBService {
...
+          return;
+        }
```
NickPhura: Do we need to also check if the `decimal Lat / Long` columns are not empty, in addition to the fields existing?

api/src/services/dwc-service.ts
NickPhura, line 81: We should add some tests to this too, make sure there are no surprise circumstances that break this.

curtisupshall (dismissed):
🥺

api/src/services/dwc-service.ts
```diff
@@ -90,4 +70,86 @@ export class DwCService extends DBService {
...
+   * @return {*}  {Promise<boolean>}
+   * @memberof DwCService
+   */
+  async decorateDWCASourceData(occurrenceSubmissionId: number): Promise<boolean> {
+    const occurrenceService = new OccurrenceService(this.connection);
+
+    const submission = await occurrenceService.getOccurrenceSubmission(occurrenceSubmissionId);
+    const jsonObject = submission.darwin_core_source;
+
+    let newJson = await this.decorateLatLong(jsonObject);
```
anissa-agahchen: Can you make a function that decorates the json object ie ```decorateDWCSource``` and have that function call the ```decorateLatLong``` and ```decorateTaxonIDs```?

Reason: we may have more decorations in the future

anissa-agahchen:
In testing locally, I see that the object is decorating correctly and will display correctly.

However on refresh, I get:
Screenshot 2023-01-25 at 12 12 19 PM
Screenshot 2023-01-25 at 12 12 26 PM

anissa-agahchen:
> In testing locally, I see that the object is decorating correctly and will display correctly.
> 
> However on refresh, I get: Screenshot 2023-01-25 at 12 12 19 PM Screenshot 2023-01-25 at 12 12 26 PM

On further testing, it's because I had a record that didn't have either lat long or UTM ... will need to account for that in the back end

api/src/services/dwc-service.ts
KjartanE: I believe the function in comment is that function, ie `decorateDWCASourceData`

KjartanE:
> > In testing locally, I see that the object is decorating correctly and will display correctly.
> > However on refresh, I get: Screenshot 2023-01-25 at 12 12 19 PM Screenshot 2023-01-25 at 12 12 26 PM
> 
> On further testing, it's because I had a record that didn't have either lat long or UTM ... will need to account for that in the back end

Thank you, I figured if I covered possible outcome with the decoration this error would not happen. But to avoid possible bad states and to handle this error, I added extra check for the Lat, Long values before they are used to display a data point.

anissa-agahchen (dismissed):

anissa-agahchen (dismissed):

curtisupshall (approved):

PR #926, opened 2023-01-29
BHBC-2144: Fix Project Team Member interface + Amend system user GUID implementation

curtisupshall:
# Overview

## Links to Jira tickets

 - https://quartech.atlassian.net/browse/BHBC-2144

 - Fixes a bug with adding participants to a project;
 - Removes the **User GUID** text field from the Add System User form;
 - Makes the `user_guid` property in the `system_users` table nullable;
 - Adds a service method for finding users by user identifier + identity source;
 - Adds a function which patches the user's GUID entry whenever their user context is set during a database connection.


api/src/queries/database/user-context-queries.ts
```diff
@@ -1,13 +1,65 @@
...
+/**
+ * Returns an SQL query for setting the user's context. As a side-effect, it first updates the user's
+ * `user_guid` to reflect the given `userGuid` value, but only if it is currently `null` (which happens
+ * when new system users are created and a GUID is not provided).
+ *
+ * @param {string} userGuid the GUID of the user
+ * @param {string} userIdentifier the user's identifier
+ * @param {string} userIdentitySource The user's identity source
+ * @returns {*} {SQLStatement | null} The SQL statement for setting the user's context, or `null` if
+ * userGuid or userIdentifier are undefinedd.
```
NickPhura: Nit: looks like this is outdated

api/src/queries/database/user-context-queries.ts
```diff
@@ -1,13 +1,65 @@
...
   }
 
-  return SQL`select api_set_context(${userGuid}, ${systemUserType});`;
+  return SQL`
+    SELECT api_set_context(${userGuid}, ${userIdentitySource});
+  `;
+};
+
+/**
+ * @TODO jsdoc
```
NickPhura: reminder just in case

NickPhura:
Just curious regarding the set guid sql + set context sql: was there no way to do it in 1 call in the end? Because of the views, or the way functions work?

api/src/database/db.ts
```diff
@@ -314,12 +314,21 @@ export const getDBConnection = function (keycloakToken: object): IDBConnection {
...
 
-    if (!userGuid || !userIdentitySource) {
+    if (!userGuid || !userIdentifier || !userIdentitySource) {
       throw new ApiGeneralError('Failed to identify authenticated user');
     }
 
+    // Patch user GUID
+    const patchUserGuidSqlStatement = queries.database.patchUserGuidSQL(userGuid, userIdentifier, userIdentitySource);
+
+    if (!patchUserGuidSqlStatement) {
```
KjartanE: Is this check needed? On line 321 the variable are checked before used in the queries.database.patchUserGuidSQL where they are checked again. Then you are covering a case that will not exists with this check.

KjartanE (approved):
Looks great! ☄️

PR #927, opened 2023-01-30
BHBC-2145: Fix bug re: validation completion flag

curtisupshall:
# Overview
Fixes a bug which causes the `isValidating` flag to be stuck on `true` for an occurrence submission that has finished validating.


NickPhura (dismissed):

anissa-agahchen (dismissed):

api/src/paths/project/{projectId}/survey/{surveyId}/observation/submission/get.ts
```diff
@@ -165,26 +165,38 @@ export function getOccurrenceSubmission(): RequestHandler {
...
         ].includes(occurrenceSubmission.submission_status_type_name);
 
-      const messageTypes: IMessageTypeGroup[] = hasAdditionalOccurrenceSubmissionMessages
+      const isDoneValidating =
+        willFetchAdditionalMessages ||
+        (occurrenceSubmission.submission_status_type_name &&
+          [
+            // Submission statuses for validation completion
+            SUBMISSION_STATUS_TYPE.TEMPLATE_VALIDATED,
+            SUBMISSION_STATUS_TYPE.DARWIN_CORE_VALIDATED
```
NickPhura, line 188: We may want/need to introduce a new status, that represents a generic completed state. Then we can just check if the submission is in that state to know if it is fully complete or not.

NickPhura: Technically a template isn't fully complete until it has been xlsx validated, transformed, and dwc validated.

But, in future if we accept DWC submissions from the user, we obv won't run the first 2 steps, and it will be compete when just the dwc validation has run.

anissa-agahchen (dismissed):

api/src/paths/project/{projectId}/survey/{surveyId}/observation/submission/get.ts
curtisupshall: > We may want/need to introduce a new status, that represents a generic completed state. Then we can just check if the submission is in that state to know if it is fully complete or not.

Agreed 🤔

anissa-agahchen (approved):

PR #931, opened 2023-01-31
BHBC-2183: Re-enable/fix required column value validator

NickPhura:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-2183

Fix/enable semi-finished validator function that checks if a column's cells are empty when they should not be.

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/services/validation-service.ts
```diff
@@ -445,6 +445,7 @@ export class ValidationService extends DBService {
     defaultLog.debug({
       label: 'persistTransformationResults',
       submissionId,
+      fileBuffers,
```
curtisupshall: I had removed this parameter in a previous commit because the output becomes so huge that it truncates any logs before this one. I suppose this could just be commented out when applicable

api/src/utils/media/csv/validation/csv-row-validator.ts
```diff
@@ -1,58 +1,45 @@
...
-              row: rowIndex + 2
-            }
-          ]);
-        }
+      if (rowValueForColumn == undefined || rowValueForColumn === null || rowValueForColumn === '') {
+        // cell is empty when it is required, add an error for this cell
+        csvWorksheet.csvValidation.addRowErrors([
+          {
+            errorCode: SUBMISSION_MESSAGE_TYPE.MISSING_REQUIRED_FIELD,
+            message: `Value is required and can not be empty`,
```
curtisupshall: Nit: "can not" => "cannot"

api/src/services/validation-service.ts
NickPhura: For sure, not sure how it returned. odd. WIll remove

api/src/utils/media/csv/validation/csv-row-validator.ts
NickPhura: Yeah I was wondering about that haha. Apparently both are allowed, though cannot is more common, so Ill change it.

curtisupshall (approved):

PR #949, opened 2023-02-14
Fixing Read More Styling

jeznorth:
# Overview

## Links to Jira tickets

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

### DIFF
--- a/app/src/components/fields/ReadMoreField.tsx
+++ b/app/src/components/fields/ReadMoreField.tsx
@@ -28,7 +28,7 @@ export const ReadMoreField: React.FC<IReadMoreFieldProps> = (props) => {
   const willTruncateText = sanitizedText.length > maxCharLength;
 
   if (!willTruncateText) {
-    return <span>{sanitizedText}</span>;
+    return <Typography color="textSecondary">{sanitizedText}</Typography>;
   }
 
   let truncationIndex = sanitizedText.slice(0, maxCharLength).lastIndexOf(' ');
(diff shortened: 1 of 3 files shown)


app/src/components/fields/ReadMoreField.tsx
```diff
@@ -28,7 +28,7 @@ export const ReadMoreField: React.FC<IReadMoreFieldProps> = (props) => {
   const willTruncateText = sanitizedText.length > maxCharLength;
 
   if (!willTruncateText) {
-    return <span>{sanitizedText}</span>;
+    return <Typography color="textSecondary">{sanitizedText}</Typography>;
```
curtisupshall, line 31: This `<span>` is already renderted inside a `<Typography color="textSecondary">` (see below).

app/src/components/fields/ReadMoreField.tsx
```diff
@@ -39,7 +39,7 @@ export const ReadMoreField: React.FC<IReadMoreFieldProps> = (props) => {
   const persistentTextPortion = sanitizedText.slice(0, truncationIndex);
 
   return (
-    <Box>
+    <>
       <Typography {...TypographyProps}>
```
curtisupshall, line 43: This component already gets `color="textSecondary"` when it is rendered by the parent:
```tsx
<ReadMoreField text={objectives.objectives} maxCharLength={200} TypographyProps={{ color: 'textSecondary' }} />
```

curtisupshall: The intent here is to allow the parent to style the "read more" Typography component based on the context it's used in

curtisupshall (dismissed):

app/src/components/fields/ReadMoreField.tsx
NickPhura: I don't think this span was wrapped in typography tags before. This return here skips everything below entirely, so it just ends up returning a '< span>' with no styling.

NickPhura (approved):

PR #952, opened 2023-02-15
BHBC-2152: Implement Artifact call to BioHub Artifact Submission API

curtisupshall:
# Overview
Implement SIMS interaction with BioHub Artifact Submission API as described in the workflow here:

 - [SIMS to BioHub Workflows](https://apps.nrs.gov.bc.ca/int/confluence/x/Zpu6C)

   - See diagram titled: “File Artifact Submissions”
## Links to Jira tickets

 - https://quartech.atlassian.net/browse/BHBC-2152

- Collector system can call the BioHub Artifact Submission API endpoint, sending it one ore more artifacts (files).
- Creates the service(s) (and related repo/util code) needed to facilitate the workflow
- Removes authenticated read from S3 upload util functions
## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


AlfredRosenthal:
the workflow mentions updating a publish table once the artifacts are uploaded to biohub, [BHBC-2169](https://github.com/bcgov/biohubbc/pull/953) has a new repo/ service for doing that. I think I covered all the publish tables but just a heads up so you don't duplicate the work

curtisupshall:
> the workflow mentions updating a publish table once the artifacts are uploaded to biohub, [BHBC-2169](https://github.com/bcgov/biohubbc/pull/953) has a new repo/ service for doing that. I think I covered all the publish tables but just a heads up so you don't duplicate the work

Good catch! I'll incorporate those changes when your branch gets merged

api/src/services/platform-service.ts
```diff
@@ -2,39 +2,87 @@ import AdmZip from 'adm-zip';
...
-  BACKBONE_API_HOST = process.env.BACKBONE_API_HOST;
-  BACKBONE_INTAKE_PATH = process.env.BACKBONE_INTAKE_PATH || '/api/dwc/submission/queue';
+  attachmentService: AttachmentService;
+  backboneIntakeEnabled: boolean;
+  backboneApiHost: string;
+  backboneIntakePath: string;
+  backboneArtifactIntakePath: string;
+
+  constructor(connection: IDBConnection) {
+    super(connection);
```
NickPhura, line 84: Generally our services always require a connection because they are always making database calls, but this service has a number of functions that don't care about the database. 

So we may need to push this connection and related service down into the function itself, so that you can still use this service to get a token, for example, without having to provide a connection that isn't used.

NickPhura: Oh wait, I am mixing this service up with the keycloak service! Ignore this comment haha

env_config/env.docker
```diff
@@ -74,16 +75,31 @@ DB_SCHEMA_DAPI_V1=biohub_dapi_v1
 DB_TZ=America/Vancouver
 
 # ------------------------------------------------------------------------------
-# KeyCloak Config
+# KeyCloak Configuration for Keycloak Common Hosted Single Sign-on (CSS)
```
NickPhura, line 78: Could even include the URL to the CSS site if you want

api/.pipeline/config.js
```diff
@@ -83,6 +83,7 @@ const phases = {
     appHost: (isStaticDeployment && staticUrls.dev) || `${appName}-${changeId}-af2668-dev.apps.silver.devops.gov.bc.ca`,
     backboneApiHost: 'https://api-dev-biohub-platform.apps.silver.devops.gov.bc.ca',
     backboneIntakePath: '/api/dwc/submission/intake',
```
NickPhura: Would you mind also adjusting the submission intake path while you're at it. I believe it is now `/queue` rather than `/intake`

api/src/services/platform-service.ts
```diff
@@ -196,4 +244,151 @@ export class PlatformService extends DBService {
...
+   */
+  async uploadProjectAttachmentsToBioHub(
+    dataPackageId: string,
+    projectId: number,
+    attachmentIds: number[],
+    reportAttachmentIds: number[]
+  ): Promise<{ artifact_id: number }[]> {
+    const attachments = (
+      await this.attachmentService.getProjectAttachments(projectId)
+    ).filter((attachment: IProjectAttachment) => attachmentIds.includes(attachment.id));
```
NickPhura: You could consider making a new function that gets attachments based on an array of ids (select * where in) type call. Then you wouldn't have to filter them here.

NickPhura:
@curtisupshall Gave the code a look over, looks good overall. ~~I think the only thing missing is populating the attachment/report publish tables after the call to biohub is made successfully for each attachment. Probably some code there that overlaps with Alfreds PR.~~

Added a couple of last minute comments, but looks good otherwise!

api/src/services/platform-service.test.ts
```diff
@@ -303,4 +304,20 @@ describe('PlatformService', () => {
       expect(insertOccurrence).to.have.been.calledOnce;
     });
   });
+
+  describe('_makeArtifactsFromAttachments', () => {
+    // TODO
+  });
+
+  describe('_submitArtifactToBioHub', () => {
+    // TODO
```
KjartanE: Wrap up tests files and looks good to go

api/.pipeline/config.js
```diff
@@ -111,6 +112,7 @@ const phases = {
     appHost: staticUrls.test,
     backboneApiHost: 'https://api-test-biohub-platform.apps.silver.devops.gov.bc.ca',
     backboneIntakePath: '/api/dwc/submission/intake',
```
NickPhura: Could you update this submission path to be `.../queue` as well, and the one below on line 115.

api/src/services/platform-service.ts
```diff
@@ -259,4 +307,180 @@ export class PlatformService extends DBService {
...
+      const artifact = await this._makeArtifactFromAttachment({ dataPackageId, attachment, file_type: 'Report' });
+      const { artifact_id } = await this._submitArtifactToBioHub(artifact);
+
+      return this.publishService.insertSurveyReportPublishRecord({
+        artifact_id,
+        survey_report_attachment_id: attachment.id
+      });
+    });
+
+    return [...(await Promise.all(attachmentArtifactPromises)), ...(await Promise.all(reportArtifactPromises))];
```
NickPhura: Could also write this as:
```
return Promise.all([...attachmentArtifactPromises, ...reportArtifactPromises]);
```
Should yield the same result

api/src/services/attachment-service.ts
```diff
@@ -33,6 +35,19 @@ export class AttachmentService extends DBService {
...
+    const attachmentIds = (await this.getProjectAttachments(projectId)).map((attachment) => attachment.id);
+    const reportAttachmentIds = (await this.getProjectReportAttachments(projectId)).map((attachment) => attachment.id);
+    await platformService.uploadProjectAttachmentsToBioHub(
+      dataPackageId,
+      projectId,
+      attachmentIds,
+      reportAttachmentIds
+    );
+  }
+
```
NickPhura: We should probably remove this before merging. Good idea on how to test it, and could be used locally to demo it is working, but lets not merge it to dev. We'll be starting on the UIs to trigger this next sprint, so it won't be longer before we can run it for real.

api/src/services/platform-service.ts
curtisupshall: When I write `Promise.all([...attachmentArtifactPromises, ...reportArtifactPromises])`, the compiler complains... Admitedly it looks a bit messy, so I'll refactor

api/src/repositories/attachment-repository.ts
```diff
@@ -132,6 +145,49 @@ export class AttachmentRepository extends BaseRepository {
...
+      .andWhere('project_id', projectId);
+
+    const response = await this.connection.knex<IProjectAttachment>(queryBuilder);
+
+    if (!response.rows) {
+      throw new ApiExecuteSQLError('Failed to get project attachments by attachmentIds', [
+        'AttachmentRepository->getProjectAttachmentsByIds',
+        'rows was null or undefined, expected rows != null'
+      ]);
+    }
```
NickPhura: This check could be removed because rows will always be an array. That being said, it is in a bunch of other places from past code, so we could opt to clean them all up separately later, as tech debt. Up to you though.

api/src/services/survey-service.test.ts
```diff
@@ -93,6 +94,10 @@ describe('SurveyService', () => {
   });
 
   describe('updateSurvey', () => {
+    beforeEach(() => {
+      sinon.stub(AttachmentService.prototype, 'tempSubmitAttachments').resolves();
+    });
+
```
NickPhura: Can this be removed?

api/src/services/platform-service.ts
```diff
@@ -259,4 +310,182 @@ export class PlatformService extends DBService {
...
+      const artifact = await this._makeArtifactFromAttachment({ dataPackageId, attachment, file_type: 'Report' });
+      const { artifact_id } = await this._submitArtifactToBioHub(artifact);
+
+      return this.publishService.insertProjectReportPublishRecord({
+        artifact_id,
+        project_report_attachment_id: attachment.id
+      });
+    });
+
+    return [...(await Promise.all(attachmentArtifactPromises)), ...(await Promise.all(reportArtifactPromises))];
```
NickPhura: Could do the same to the return here as below

api/src/repositories/attachment-repository.ts
```diff
@@ -224,6 +282,56 @@ export class AttachmentRepository extends BaseRepository {
...
+        'title',
+        'description',
+        knex.raw('year::int as year_published'),
+        knex.raw(`
+          CASE
+            WHEN update_date IS NULL
+            THEN create_date::text
+            ELSE update_date::text
+          END AS last_modified
+        `),
```
NickPhura, line 307: Just curious: looks like we only do this for project/survey reports, but not attachments. Any reason? Or just do work with existing interfaces that were designed differently, or something?

KjartanE (approved):
🥇

PR #953, opened 2023-02-17
BHBC-2169: Implement DwCA call to BioHub DwCA Submission API when Observations are Submitted

AlfredRosenthal:
# Overview

## Links to Jira tickets

[BHBC-2169](https://quartech.atlassian.net/browse/BHBC-2169?atlOrigin=eyJpIjoiNWUwN2RhMDdiOGQ5NGQ1NDgxYzhiMDJjOTE3YmNmZGUiLCJwIjoiaiJ9)

- Updated Biohub submission code to hit new queue endpoint
- updated that same call to insert into new publish tables
- created new repo/ service to handle publish tables

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

### DIFF
--- a/api/src/repositories/survey-repository.ts
+++ b/api/src/repositories/survey-repository.ts
@@ -66,6 +66,15 @@ export interface IObservationSubmissionUpdateDetails {
   outputKey?: string;
 }
 
+export interface ISurveyProprietorModel {
+  first_nations_id: number;
+  proprietor_type_id: number;
+  survey_id: number;
+  rational: string;
+  proprietor_name: string;
+  disa_required: boolean;
+}
+
 const defaultLog = getLogger('repositories/survey-repository');
 
 export class SurveyRepository extends BaseRepository {
@@ -242,6 +251,18 @@ export class SurveyRepository extends BaseRepository {
     }
 
     return new GetSurveyFundingSources(result);
+  }
+
+  async getSurveyProprietorDataForSecurityRequest(surveyId: number): Promise<ISurveyProprietorModel> {
+    const sqlStatement = SQL`
+      SELECT * 
+      FROM survey_proprietor as sp
+      WHERE survey_id = ${surveyId};
+    `;
+
+    const response = await this.connection.sql<ISurveyProprietorModel>(sqlStatement);
+    
+    return (response && response.rows && response.rows?.[0]) || null;
   }
 
   async getSurveyProprietorDataForView(surveyId: number): Promise<GetSurveyProprietorData | null> {
--- a/api/src/repositories/survey-repository.ts
+++ b/api/src/repositories/survey-repository.ts
@@ -227,9 +227,10 @@ export class SurveyRepository extends BaseRepository {
   }
 
   /**
-   * Deletes a survey and any associations for a given survey
+   * Get Survey and Funding Source data for a given surveyId
    *
    * @param {number} surveyId
+   * @returns {*} Promise<GetSurveyFundingSources>
    * @memberof SurveyRepository
    */
   async getSurveyFundingSourcesData(surveyId: number): Promise<GetSurveyFundingSources> {
@@ -311,6 +312,13 @@ export class SurveyRepository extends BaseRepository {
     return (response && response.rows && response.rows?.[0]) || null;
   }
 
+  /**
+   * Get Survey Proprietor data for view for a given survey id
+   *
+   * @param {number} surveyId
+   * @returns {*} Promise<GetSurveyProprietorData | null>
+   * @memberof SurveyRepository
+   */
   async getSurveyProprietorDataForView(surveyId: number): Promise<GetSurveyProprietorData | null> {
     const sqlStatement = SQL`
       SELECT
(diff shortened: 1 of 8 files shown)


api/src/services/platform-service.ts
```diff
@@ -185,15 +199,31 @@ export class PlatformService extends DBService {
...
       },
-      dataPackageId: emlService.packageId
+      dataPackageId: emlService.packageId,
+      securityRequest
     };
 
-    return this._submitDwCADatasetToBioHubBackbone(dwCADataset);
+    const queueResponse = await this._submitDwCADatasetToBioHubBackbone(dwCADataset);
+
+    Promise.all([
```
curtisupshall: Should we put an `await` here?

api/src/services/project-service.ts
```diff
@@ -342,7 +343,13 @@ export class ProjectService extends DBService {
     await this.insertParticipantRole(projectId, PROJECT_ROLE.PROJECT_LEAD);
 
     //Submit Eml to biohub
-    await this.platformService.submitDwCAMetadataPackage(projectId);
+    const queueResponse = await this.platformService.submitDwCAMetadataPackage(projectId);
+
+    // take queue id and insert into history publish table
+    if (queueResponse?.queue_id) {
```
curtisupshall: Do we need this check? The return type of `platformService.submitDwCAMetadataPackage` implies that this branch will always be taken.

curtisupshall (dismissed):
Looks great!

api/src/repositories/survey-repository.ts
```diff
@@ -244,6 +253,18 @@ export class SurveyRepository extends BaseRepository {
     return new GetSurveyFundingSources(result);
   }
 
+  async getSurveyProprietorDataForSecurityRequest(surveyId: number): Promise<ISurveyProprietorModel> {
```
KjartanE, line 303: Nit... Want to add JSDoc for the new function

api/src/services/platform-service.ts
KjartanE: I don't think so, don't need an await for last call in a function

api/src/services/survey-service.ts
```diff
@@ -148,6 +150,10 @@ export class SurveyService extends DBService {
     return this.surveyRepository.getLatestSurveyOccurrenceSubmission(surveyId);
   }
 
+  async getSurveyProprietorDataForSecurityRequest(surveyId: number): Promise<ISurveyProprietorModel> {
```
KjartanE, line 244: Nit... JSDoc

api/src/services/project-service.ts
AlfredRosenthal: the `submitDwCAMetaDataPackage` function can possibly fail (try catch) but currently isn't throwing so an undefined is a valid response. Good point of discussion though, should `submitDwCAMetaDataPackage` silently fail? should it just throw the error up at let the parent handle it? @NickPhura @KjartanE

NickPhura: I would move this whole check and history table call into the platform service function (or new platform parent function, w/e makes sense), and then squash any errors there. I don't think we want to prevent a project from being created just because the submit fails. And now that we have the history tables, we can maybe build out something to inform the user or admins, later.

NickPhura: Could probably even not await this call here, since it doesn't impact the surrounding code.

api/src/services/platform-service.ts
NickPhura: > I don't think so, don't need an await for last call in a function

as long as you return the promise all. Then the parent can call await.

But yeah, either await here, or return it. Returning is technically more correct, though I don't think it really matters in the end.

api/src/services/project-service.ts
NickPhura: Unless it relies on the current db connection, then we would need to await, so the connection doesn't get closed in the endpoint before its done running.

api/src/services/platform-service.ts
AlfredRosenthal: updated and added an await

api/src/services/project-service.ts
AlfredRosenthal: added a new function call to wrap the submit and publish inserts

api/src/repositories/survey-repository.ts
AlfredRosenthal: done and then some

api/src/services/platform-service.ts
```diff
@@ -185,15 +199,55 @@ export class PlatformService extends DBService {
...
+      publishRepo.insertOccurrenceSubmissionPublishRecord({
+        occurrence_submission_id: surveyData.id,
+        queue_id: queueResponse.queue_id
+      })
+    ]);
+  }
+
+  async submitAndPublishDwcAMetadata(projectId: number, surveyId?: number) {
+    try {
+      const queueResponse = await this.submitDwCAMetadataPackage(projectId);
```
curtisupshall, line 241: Thoughts on explictly typing the return value of this function + JS doc so it's obvious why `queueResponse` may be undefined? e.g.

```
/**
 * @returns {*} {Promise<{ queue_id: number} | undefined>} Resolves `undefined` if backbone intake is
 * disabled or if submission fails, or the ID of the queue record if the submission is queued successfully.
 */
async submitDwCAMetadataPackage(projectId: number): Promise<{ queue_id: number } | void> {
```

curtisupshall (dismissed):
+1 for JS docs. Looks good!

curtisupshall (approved):
💎

PR #962, opened 2023-03-01
BHBC-2164: Elk SRB/ General template Validation/ Transformation

AlfredRosenthal:
# Overview

## Links to Jira tickets

[BHBC-2164](https://quartech.atlassian.net/browse/BHBC-2164?atlOrigin=eyJpIjoiOWU3NTQ2MGQxZmU0NGY2OWIyOGUxZjJlMWE0ZTQ1M2QiLCJwIjoiaiJ9)

- Fixed issue with certain values throwing false negatives when validating
- Fixed issue with too many calls being made to the taxonomy service

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/services/dwc-service.ts
```diff
@@ -18,6 +18,39 @@ export class DwCService extends DBService {
...
+  }
+
+  /**
+   * Calls elastic search taxonomy service with the given set of taxon codes
+   *
+   * @param {Set<string>} set A set of taxon IDs
+   * @return {*}  {Promise<any>}
+   * @memberof DwCService
+   */
+  async getEnrichedDataForSpeciesSet(set: Set<string>): Promise<any> {
```
NickPhura: Can we make the `Promise<any>` more specific?

NickPhura: Either some kind of `Record<string, whatevergetEnrichedDataForSpeciesCodeReturns>`

or maybe even a `new Map<string, whatevergetEnrichedDataForSpeciesCodeReturns>` (assuming Map is useful here, can't really remember what the pros/cons of Map are tbh, over a plain Record)

AlfredRosenthal: the difference I could see between record and map is that map can be iterated over. It also provides functions like `.has(key)` or `.size`. I'm put a map because it looks a little cleaner to add/ retrieve the data.

api/src/utils/media/csv/validation/csv-row-validator.ts
```diff
@@ -98,18 +97,18 @@ export const getCodeValueFieldsValidator = (config?: ColumnCodeValidatorConfig):
       // compare allowed code values as lowercase strings
       const allowedCodeValuesLowerCase: (string | number)[] = [];
       const allowedCodeValues = config.column_code_validator.allowed_code_values.map((allowedCode) => {
-        allowedCodeValuesLowerCase.push(safeToLowerCase(allowedCode.name));
+        allowedCodeValuesLowerCase.push(String(safeToLowerCase(allowedCode.name)));
```
curtisupshall: If `allowedCode.name` isn't a String, the `safeToLowerCase` will not convert it to lowercase. So the end result of this operation is always a String, but it is not guarenteed to be lowercase. Since we're always attempting to cast `allowedCode.name` to String, should we swap the order of operations to assure that it always gets converted to lowercase?

api/src/utils/media/csv/validation/csv-row-validator.ts
```diff
@@ -256,12 +255,12 @@ export const getNumericFieldsValidator = (config?: ColumnNumericValidatorConfig)
         return csvWorksheet;
       }
 
+      const rowValueForColumn = Number(row[columnIndex]);
```
curtisupshall: Any reason for moving this operation before the branch on L260?

api/src/utils/media/xlsx/validation/xlsx-validation.ts
```diff
@@ -19,7 +19,7 @@ export type ParentChildKeyMatchValidatorConfig = {
  * required fields.
  *
  * @param {ParentChildKeyMatchValidatorConfig} [config] The validator config
- * @return {*}  {WorkBookValidator} The workbook validator
+ * @return {*}  {WorkBookValidator} The workbook validator68
```
curtisupshall: Typo?

api/src/utils/media/csv/validation/csv-row-validator.ts
AlfredRosenthal: I think it was shifted while I was testing things. Moved back

api/src/utils/media/csv/validation/csv-row-validator.ts
AlfredRosenthal: good catch. Updated both instances of this casting to cast to string first, then try to lowercase it

NickPhura (dismissed):

app/src/features/surveys/view/SurveyObservations.tsx
```diff
@@ -286,39 +286,42 @@ const SurveyObservations: React.FC<ISurveyObservationsProps> = (props) => {
...
-                                  return (
-                                    <li key={messageObject.id}>
-                                      <Typography variant="body2">{messageObject.message}</Typography>
-                                    </li>
-                                  );
-                                })}
+                      {
+                        // Alphabetize message types for consistency
+                        submissionMessageTypes
+                          .sort((messageA, messageB) => {
```
NickPhura: I wonder if it would be better to do the sorting first in JS, outside of the return, and then this code can just iterate over the array as provided.

NickPhura: mainly just thinking about how to keep the html simple

AlfredRosenthal: good call, moved it into it's own function

NickPhura (approved):

PR #976, opened 2023-03-15
BHBC-2267: Create Survey context

curtisupshall:
# Overview

Creates a Survey Context, which wraps survey-related app sections, providing a centralized location for fetching/refreshing survey data.

## Links to Jira tickets

 - https://quartech.atlassian.net/browse/BHBC-2267

 - Creates the `useSurveyContext` hook
 - Wraps all survey routes in the `ProjectRouter` with the `SurveyContextProvider`
 - Removes any component props that pass a Survey object

## Testing
 1. Create a project if you don't have one
 2. Create a survey, if you dont' have one
 3. Click on the survey
 - [x] Ensure that the survey loads
 4. Upload an observations
 - [x] Ensure that the observation loads
 5. Upload a report and attachment
 - [x] Ensure that attachments load


app/src/contexts/surveyContext.tsx
```diff
@@ -0,0 +1,67 @@
...
+      surveyDataLoader.refresh(projectId, surveyId)
+    }
+  }
+
+  useEffect(() => {
+    refresh()
+  }, [projectId, surveyId])
+
+  const surveyContext: ISurveyContext = {
+    surveyWithDetails: surveyDataLoader.data || null,
```
NickPhura: I wonder if you could just return the dataloader itself, and then the children could reference `.data`, `.error`, `.refresh`, etc, directly from it.

app/src/contexts/surveyContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+ */
+export const useSurveyContext = (): ISurveyContext => {
+  const urlParams = useParams();
+  const surveyContext = useContext(SurveyContext);
+
+  // Update survey each time the survey ID or project ID changes
+  useEffect(() => {
+    surveyContext._setSurveyId(urlParams['survey_id']);
+    surveyContext._setProjectId(urlParams['id']);
+  }, [surveyContext, urlParams]);
```
NickPhura: Could/should this be `survey_id` and `id` directly, rather than urlParms?

app/src/contexts/surveyContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+  _setSurveyId(surveyId: number): void;
+  _setProjectId(projectId: number): void;
+}
+
+const SurveyContext = createContext<ISurveyContext>({
+  surveyDataLoader: {} as DataLoader<[project_id: number, survey_id: number], IGetSurveyForViewResponse, unknown>,
+  projectId: null,
+  surveyId: null,
+  _setSurveyId: () => null,
+  _setProjectId: () => null
```
NickPhura: nit: return `void` I think?

app/src/components/publish/SubmitSurvey.tsx
```diff
@@ -42,25 +35,30 @@ export const SurveySubmitFormYupSchema = yup.object().shape({
...
   });
-  observationDataLoader.load(
-    surveyDetails.surveyData.survey_details.project_id,
-    surveyDetails.surveyData.survey_details.id
-  );
+
+  if (surveyDetails) {
+    observationDataLoader.load(
+      surveyDetails.surveyData.survey_details.project_id,
+      surveyDetails.surveyData.survey_details.id
```
NickPhura: The project/survey ids could come from useParams as well (Not sure if that changes anything).

app/src/contexts/surveyContext.tsx
curtisupshall: Apparently `void` can't be used as an expression, so it's either `() => null` or `() => {}`, I'm pretty sure.

app/src/components/publish/SubmitSurvey.tsx
curtisupshall: My intent was basically to let the Survey Context be the source of truth for survey and project ID, rather than sometimes getting this info from the URL params and sometimes getting it from the survey context. But yes in this case, we could call `useParams` instead, sinec we're just getting the two IDs

app/src/contexts/surveyContext.tsx
curtisupshall: My thinking was that rather than having Survey Context consumers poke at the URL params, just let the Survey Context fetch them and update the project and survey ID, so subscribers can just reliably get these two IDs from the survey context (assuming that the survey context is only used when these URL params are made available by a router higher up in the state tree)

app/src/contexts/surveyContext.tsx
NickPhura: Ah, I was reading this as an interface, but its an actual initial value for the context. I think undefined might be the equivalent to void? I believe undefined is what gets returned from a function with no `return`.

(Probably doesnt matter much either way)

app/src/components/publish/SubmitSurvey.tsx
NickPhura: > My intent was basically to let the Survey Context be the source of truth for survey and project ID

Yeah thats fair. I think that is totally fine, as long as we are consistent everywhere.

app/src/contexts/surveyContext.tsx
NickPhura: I just mean as far as the useEffect dependencies are concerned. If `urlParams` is included, then other url params outside of project/survey id would cause the useEffect to re-run.

That being said, its quite possible that when `useParams` changes, the whole context is going to update regardless, so maybe narrowing the useEffect dependencies wouldn't help anything. Not totally sure though without testing it.

curtisupshall: Ohh I totally misunderstood. Now I'm thinking maybe it should just be `[urlParams['survey_id'], urlParams['id']]`. If we include `surveyContext`, then it will run twice - once when `urlParams` updates, and again when `surveyContext.surveyId` changes in the parent...

app/src/contexts/surveyContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+  surveyId: number;
+}
+
+export const SurveyContext = createContext<ISurveyContext>({
+  surveyDataLoader: {} as DataLoader<[project_id: number, survey_id: number], IGetSurveyForViewResponse, unknown>,
+  projectId: -1,
+  surveyId: -1
+});
+
+export const SurveyContextProvider = (props: PropsWithChildren<Record<never, any>>) => {
```
anissa-agahchen, line 45: Question:  Is `props: PropsWithChildren<Record<never, any>>` still a good idea?  I wonder if we should specify the props that the context is using.... or maybe even go with React.PropsWithChildren<{}>

curtisupshall: See: https://stackoverflow.com/a/66271367/2932330

> I wonder if we should specify the props that the context is using

Since the Provider doesn't take any props (except children), we wouldn't want to type it with `ISurveyContext`.

> or maybe even go with React.PropsWithChildren<{}>

I tried this, but according the ES lint there are situations where this confuses the compiler.

app/src/contexts/surveyContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const surveyContext: ISurveyContext = useMemo(
+    () => ({
+      surveyDataLoader,
+      projectId,
+      surveyId
+    }),
+    [surveyDataLoader, projectId, surveyId]
+  );
+
+  return <SurveyContext.Provider value={surveyContext} {...props} />;
```
NickPhura: A bit subjective, but can we break this return out, just for clarity: `<stuff>{props.children}</stuff>`

NickPhura: Unless there is some reason that doesnt work with context providers

app/src/contexts/surveyContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+export const SurveyContextProvider = (props: PropsWithChildren<Record<never, any>>) => {
+  const biohubApi = useBiohubApi();
+  const surveyDataLoader = useDataLoader(biohubApi.survey.getSurveyForView);
+  const urlParams = useParams();
+
+  if (!urlParams['id']) {
+    throw new Error(
+      "The project ID found in SurveyContextProvider was invalid. Does your current React route provide an 'id' parameter?"
+    );
+  } else if (!urlParams['survey_id']) {
```
NickPhura: (not really an issue) You can just make this a regular if, rather than an else if, since only one of these ifs can throw at a time anyways.

AlfredRosenthal (commented):
project/ survey creation and editing works as expected! 🐹

app/src/contexts/surveyContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * The Data Loader used to load survey data
+   *
+   * @type {DataLoader<[project_id: number, survey_id: number], IGetSurveyForViewResponse, unknown>}
+   * @memberof ISurveyContext
+   */
+  surveyDataLoader: DataLoader<[project_id: number, survey_id: number], IGetSurveyForViewResponse, unknown>;
+
+  /**
+   * The project ID belonging to the current survey
```
NickPhura: nit: current project

app/src/routers/ProjectsRouter.tsx
```diff
@@ -57,28 +58,10 @@ const ProjectsRouter: React.FC = () => {
...
-        <ProjectsLayout>
-          <SurveyPage />
-        </ProjectsLayout>
-      </AppRoute>
-
-      <AppRoute exact path="/admin/projects/:id/surveys/:survey_id/results" layout={ProjectsLayout}>
-        <ProjectsLayout>
-          <SurveyPage />
-        </ProjectsLayout>
+      <AppRoute path="/admin/projects/:id/surveys/:survey_id" layout={ProjectsLayout}>
```
NickPhura: Could you add the title prop here? (see `AppRouter.tsx`).

The `getTitle` function could be exported, or it could just be removed tbh, as there really is no reason to have it broken out. We may as well just hardcode all the titles. Whatever you prefer.

app/src/contexts/surveyContext.tsx
NickPhura: Yeah `Record<whatever,whatever>` is preferred over curly brace objects in general. There are lots of things in javascript that aren't actually objects, that resolve to some kind of `{}`, so its not a reliable way to denote an "object" of the kind we are typically thinking of.

NickPhura: Or at least, something like that, I cant remember the specifics anymore

app/src/routers/ProjectsRouter.tsx
curtisupshall: Currently we don't maintain titles for Surveys, but I agree. Should it just say `SIMS - Survey`?

app/src/contexts/surveyContext.tsx
NickPhura: And then `object` and `Object` also factor in somewhere... but I think `Record` is still preferred in this case.

app/src/routers/ProjectsRouter.tsx
NickPhura: Yeah SIMS - Survey works, whatever the existing pattern is

app/src/routers/ProjectsRouter.tsx
```diff
@@ -57,28 +59,10 @@ const ProjectsRouter: React.FC = () => {
...
-        <ProjectsLayout>
-          <SurveyPage />
-        </ProjectsLayout>
-      </AppRoute>
-
-      <AppRoute exact path="/admin/projects/:id/surveys/:survey_id/results" layout={ProjectsLayout}>
-        <ProjectsLayout>
-          <SurveyPage />
-        </ProjectsLayout>
+      <AppRoute path="/admin/projects/:id/surveys/:survey_id" title={getTitle('Projects')} layout={ProjectsLayout}>
```
NickPhura: nit: title: Surveys

curtisupshall: 🤦‍♂️

curtisupshall: Fixed!

AlfredRosenthal (dismissed):
🐶

app/src/features/surveys/view/components/SurveyStudyArea.tsx
```diff
@@ -72,14 +70,15 @@ const SurveyStudyArea: React.FC<ISurveyStudyAreaProps> = (props) => {
...
-  } = props;
+  const surveyContext = useContext(SurveyContext);
+  const surveyForViewData = surveyContext.surveyDataLoader.data;
+
+  const { projectForViewData } = props;
+
+  const survey_details = useMemo(() => surveyForViewData?.surveyData?.survey_details, [surveyForViewData]);
+  const occurrence_submission = useMemo(() => surveyForViewData?.surveySupplementaryData?.occurrence_submission, [
+    surveyForViewData
+  ]);
```
NickPhura: Is `useMemo` needed here? Since the value being memoized is also the same value in the dependencies, seems like it wouldn't add any benefit?

app/src/contexts/surveyContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+   */
+  useEffect(() => {
+    if (projectId && surveyId) {
+      surveyDataLoader.refresh(projectId, surveyId);
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [projectId, surveyId]);
+
+  const surveyContext: ISurveyContext = useMemo(
```
NickPhura, line 76: Wondering if `useMemo` is helping here. Since the return is the same as the dependencies, I don't think it would benefit anything?

NickPhura: Unless there is something sneaky with the fact that its wrapped into an object?

NickPhura: Tried logging when it re-runs with and without the memo, and it seems to be the same, so I dont think its helping

curtisupshall: I think there was a time where I needed to include it, since I was getting runtime warnings. I'll remove it.

app/src/features/surveys/view/components/SurveyStudyArea.tsx
curtisupshall: As discussed, `useMemo` is used since the reference to `survey_details` would change every render. After testing, `useMemo` doesn't seem to affect the behaviour, so will remove.

NickPhura (approved):
Tested locally, seems to work well.

anissa-agahchen (approved):

PR #980, opened 2023-03-17
BHBC-2268-2: Update survey supplemental data to include publish data

NickPhura:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-2268

- Update survey supplemental data section to include publish history records (for occurrence, summary, and survey metadata)

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/repositories/survey-repository.ts
```diff
@@ -380,33 +380,26 @@ export class SurveyRepository extends BaseRepository {
...
-   * Get Occurrence submission id for a given survey ID
+   * Get Occurrence submission for a given survey id.
    *
    * @param {number} surveyId
-   * @returns {*} Promise<number>
+   * @return {*}
    * @memberof SurveyRepository
    */
-  async getOccurrenceSubmissionId(surveyId: number): Promise<number> {
+  async getOccurrenceSubmission(surveyId: number) {
```
AlfredRosenthal: these should have a return, not obvious that it could be null

api/src/services/survey-service.ts
```diff
@@ -279,8 +297,15 @@ export class SurveyService extends DBService {
...
-  async getSummaryResultId(surveyId: number): Promise<number> {
-    return this.surveyRepository.getSummaryResultId(surveyId);
+  /**
+   * Get a survey summary submission record for a given survey id.
+   *
+   * @param {number} surveyId
+   * @return {*}
+   * @memberof SurveyService
+   */
+  async getSurveySummarySubmission(surveyId: number) {
```
AlfredRosenthal: nit: would be nice to see the return types here

NickPhura: For sure. Was playing around at one point with letting types bubble up from the lowest level repo function. The types do work if you hover over the function or return value, but I agree its not as obvious at a glance anymore which is annoying.

Will update.

api/src/repositories/survey-repository.ts
NickPhura: updated

api/src/services/survey-service.ts
NickPhura: updated

AlfredRosenthal (approved):
🏎️

anissa-agahchen (approved):

PR #989, opened 2023-03-29
BHBC-2282: Fix draft load

anissa-agahchen:
(https://quartech.atlassian.net/browse/BHBC-2282)

Testing procedure:
Baseline:
1. Log into Dev
2. Throttle your network
3. Load a draft project
4. You will see an empty draft project

As a result of this code change:
1. Go to the app link for this PR
2. Log in
3. Throttle your network
4. Load a draft project
5. You will see the content.


package-lock.json
curtisupshall, line 1: Should we remove this file? At this level there are no `package.json` or node modules

app/src/features/projects/create/CreateProjectPage.tsx
```diff
@@ -66,7 +66,7 @@ const CreateProjectPage: React.FC = () => {
 
   // Reference to pass to the formik component in order to access its state at any time
   // Used by the draft logic to fetch the values of a step form that has not been validated/completed
-  const formikRef = useRef<FormikProps<ICreateProjectRequest>>(null);
+  const [formikRef] = useState(useRef<FormikProps<ICreateProjectRequest>>(null));
```
curtisupshall: Is there a way around creating a state variable that never gets updated? To me it seems like a bit of an anti-pattern

anissa-agahchen: Good question.  I agree that it is odd, however I suspect that the reason we have this pattern established throughout the code is that formik.current.values is constantly updated as users type in the form.  Do you have suggestions for a different setup?

curtisupshall: If we're hooking into useState because we want to preserve the value between renders, we can use the useMemo hook instead 🤔 I'll try pulling the branch and see what I can come up with

anissa-agahchen: hmm interesting, I'll check out useMemo too

anissa-agahchen: @curtisupshall ... turns out the Usestate was unneccessary, as the useRef itself is sufficient.  What we needed to do was check that formik.current is not null before setting the values .. and this solution doesn't have the added complexity of a useState or Use Memo.  

I also looked into useMemo, it doesn't apply to this case, because the intention of useMemo to memoize computational heavy components ... and I don't think the draft functionality is heavy ... but maybe loading the whole project is.  

Either way, seems to work fine now.  Would you please take a look again when you have a chance?

app/src/features/projects/list/ProjectsListPage.tsx
```diff
@@ -168,6 +172,8 @@ const ProjectsListPage: React.FC = () => {
     const getDrafts = async () => {
       const draftsResponse = await biohubApi.draft.getDraftsList();
 
+      console.log('draftsResponse: ', draftsResponse);
```
KjartanE: nit: console log

KjartanE (dismissed):
😸

KjartanE (dismissed):
🌵

PR #1011, opened 2023-04-24
TechDebt: Misc Odd Updates (keycloak, pipeline, favicon, etc)

NickPhura:
# Overview

## Links to Jira tickets

n/a

- Removed some old git workflow files
- Updated keycloak version, which changed it to always being defined (rather than potentially undefined), and so I removed a bunch of unnecessary `?.` when dereferencing the keycloak object.
- Removed a bunch of null user object checks, as the user object would never be undefined.
- Added `change_version` to the version endpoint (mainly an experimental change, as I never really noticed that var before, but in theory it might let us put a version number in the UI)
- Pulled some taxonomy-service changes from other projects, just misc updated
- Added the BCGov favicon to the react app.
- Added proper error handling to the db setup pipeline step, so it will properly fail the git action if it fails in openshift
- Moved a bunch of old tests from manually specifying mock keycloak auth states, to using the `getMockAuthState` test helper function

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


app/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -47,10 +47,10 @@ export interface IKeycloakWrapper {
   /**
    * Original raw keycloak object.
    *
-   * @type {(Keycloak | undefined)}
+   * @type {(Keycloak)}
    * @memberof IKeycloakWrapper
    */
-  keycloak: Keycloak | undefined;
+  keycloak: Keycloak;
```
curtisupshall, line 54: +1 for no more undefined keycloak 🙏

curtisupshall (dismissed):
Code looks great! Lots of welcome changes

curtisupshall (dismissed):

AlfredRosenthal (dismissed):
Ran it locally, tried all the areas that were touched, it looks good

🦺

NickPhura:
@curtisupshall @AlfredRosenthal Thanks for the reviews, I had to fix some conflicts so will need a re-review. Nothing functional should have changed. I updated some tests while I was at it to use `getMockAuthState` rather than hardcoding it in each test. Will see how that evolves long term, but it seemed like we were having issues maintaining our mock auth states and so hopefully this will help centralize it a bit.

curtisupshall (approved):

al-rosenthal (approved):
🕶️

PR #1024, opened 2023-04-27
SIMS-BIOHUB-8

anissa-agahchen:
# Overview

## Links to Jira tickets
https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-8

- SIMS code changes needed to display file errors when required sheets are missing from a template

Test files for Goat:
[content_error_Monashee_Goat.xlsx](https://github.com/bcgov/biohubbc/files/11400422/content_error_Monashee_Goat.xlsx)
[NoTabs_Monashee_Goat_2.0 .xlsx](https://github.com/bcgov/biohubbc/files/11400425/NoTabs_Monashee_Goat_2.0.xlsx)
[Monashee_Mt_Goat_Total_Count_Recuit_Comp_Survey_2.0.xlsx](https://github.com/bcgov/biohubbc/files/11400428/Monashee_Mt_Goat_Total_Count_Recuit_Comp_Survey_2.0.xlsx)


api/src/services/validation-service.ts
```diff
@@ -359,14 +359,11 @@ export class ValidationService extends DBService {
     return validationSchemaParser;
   }
 
-  validateXLSX(file: XLSXCSV, parser: ValidationSchemaParser) {
+  validateXLSX_occurrence(file: XLSXCSV, parser: ValidationSchemaParser) {
     // Run media validations
     file.validateMedia(parser);
 
     const media_state = file.getMediaState();
```
anissa-agahchen, line 366: return an empty csvstate and regualr media state .. repeat for summary

api/src/json-schema/validation-schema.ts
```diff
@@ -118,6 +118,7 @@ export const submissionValidationSchema = {
       title: 'File/Sheet Validation',
       description: 'The validators that can be applied against a file/sheet within a submission file.',
       anyOf: [
+        //below: column header validator
```
anissa-agahchen: replace with descrition field

api/src/services/summary-service.test.ts
```diff
@@ -490,7 +490,7 @@ describe('SummaryService', () => {
         ]);
 
       const getRules = sinon.stub(service, 'getValidationRules').resolves('');
-      const validate = sinon.stub(service, 'validateXLSX').resolves({});
+      const validate = sinon.stub(service, 'validateXLSX_summary').resolves({});
```
anissa-agahchen: revert ... make a ticket for techdebt to address combining .. maybe along with status ticket

api/src/services/validation-service.ts
anissa-agahchen: check if the content will be validated after

curtisupshall (dismissed):
Code looks good!

AlfredRosenthal (dismissed):
🐐

api/src/services/validation-service.ts
```diff
@@ -360,15 +360,15 @@ export class ValidationService extends DBService {
...
   validateXLSX(file: XLSXCSV, parser: ValidationSchemaParser) {
-    // Run media validations
+    defaultLog.debug({ label: 'validateXLSX' });
     file.validateMedia(parser);
 
     const media_state = file.getMediaState();
+
     if (!media_state.isValid) {
-      throw SubmissionErrorFromMessageType(SUBMISSION_MESSAGE_TYPE.INVALID_MEDIA);
+      return { csv_state: ([] as unknown) as ICsvState[], media_state };
```
NickPhura: We should avoid any type casting, and just provide a dummy value for csv_state that is type valid. Otherwise it might lead to unexpected behaviour downstream when functions are expecting ICsvState but secretly receiving something else.

api/src/services/validation-service.ts
```diff
@@ -434,13 +436,13 @@ export class ValidationService extends DBService {
...
-      if (!mediaState.isValid || csvState?.some((item) => !item.isValid)) {
-        // At least 1 error exists, skip remaining steps
-        parseError = true;
-      }
     });
 
+    if (!mediaState.isValid || csvState?.some((item) => !item.isValid)) {
+      // At least 1 error exists, skip remaining steps
+      parseError = true;
+    }
```
NickPhura: We may as well just check this in the `if` below, and throw the SubmissionError directly, rather than assigning `parseError` here and then just immediately checking it below.

api/src/utils/media/xlsx/transformation/xlsx-transform.ts
```diff
@@ -135,17 +135,17 @@ export class XLSXTransform {
     const worksheetJSONWithKey: RowObject[] = [];
 
     for (let i = 0; i < numberOfRows; i++) {
-      const primaryKey = this._getKeyForRowObject(worksheetJSON[i], templateMetaSchema.primaryKey);
+      const primaryKey = this._getKeyForRowObject(worksheetJSON[i], templateMetaSchema.primaryKey, i);
```
NickPhura: We can't make this function always add an index for the row number. It breaks other transformation steps that want their keys to not include a row number.

If we need to include the row number in the key, we need to just add a `[..._row]` part to the key in the config. Should be examples in existing configs.

api/src/services/validation-service.ts
```diff
@@ -397,11 +400,13 @@ export class ValidationService extends DBService {
   async persistValidationResults(csvState: ICsvState[], mediaState: IMediaState): Promise<boolean> {
     defaultLog.debug({ label: 'persistValidationResults', message: 'validationResults' });
 
-    let parseError = false;
+    const parseError = false;
```
NickPhura: Lets remove this `parseError` entirely, and the return at the end of the function. We can just make this function a Promise Void type.

We only ever use the original return value in 1 spot, and its for a test. We can just update that test to assert that if no exception is thrown, it passes, since that is really what this function does now.

anissa-agahchen: resolved in both validate-service and summary-service

NickPhura (approved):

AlfredRosenthal (approved):
🥮

PR #1025, opened 2023-05-02
SIMSBIOHUB-18: Fix application redirection links

curtisupshall:
# Overview

Enhances URL redirection in the SIMS frontend.

## Links to Jira tickets

 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-18

- Fixes the application URL provided in Access Request email links
- Adds a login endpoint to hook into Keycloak sign-in
- Adds the `useRedirect` hook


app/src/components/layout/Header.tsx
```diff
@@ -224,44 +230,48 @@ const Header: React.FC = () => {
...
+          </Container>
+        </Box>
       </AppBar>
 
       <Dialog open={open}>
         <DialogTitle>Contact Support</DialogTitle>
         <DialogContent>
           <Typography variant="body1" component="div" color="textSecondary">
-            For technical support or questions about this application, please email&nbsp;
+            For technical support or questions about this application, please email &zwnj;
```
jeznorth: Why the change here? Just curious.

app/src/pages/landing/LandingPage.tsx
```diff
@@ -1,15 +1,178 @@
...
-  });
+  const hasAdministrativeRole = keycloakWrapper?.hasSystemRole([
+    SYSTEM_ROLE.DATA_ADMINISTRATOR,
+    SYSTEM_ROLE.SYSTEM_ADMIN
+  ]);
 
-  return null;
+  return (
+    <BaseLayout className={classes.baseLayoutContainer}>
+      <Container maxWidth="md">
```
jeznorth: We're going to have to come up with a better strategy here - this layout seems odd on larger viewports.

jeznorth: Encapsulate the content with a negative top margin to align visual center here, not at the containing element.

.config/config.json
```diff
@@ -20,7 +20,7 @@
   "staticUrls": {
     "dev": "dev-biohubbc.apps.silver.devops.gov.bc.ca",
     "test": "test-biohubbc.apps.silver.devops.gov.bc.ca",
-    "prod": "biohubbc.apps.silver.devops.gov.bc.ca"
+    "prod": "sims.nrs.gov.bc.ca"
```
NickPhura, line 23: We'll need to specifically test this change in OpenShift. 

It might be fine, but it might also cause trouble if the pods need the internal openshift route to communicate properly. If that ends up being the case, we could add a 4th field here called something like "prod-vanity-url", and then it can pass the sims URL to the appHost env var, and the rest of them can use the original internal url.

app/src/components/security/RouteGuards.tsx
```diff
@@ -119,26 +119,14 @@ export const UnAuthenticatedRouteGuard = (props: RouteProps) => {
...
-      const redirectUrlParams = qs.stringify(urlParams, { filter: (prefix) => prefix !== 'authLogin' });
-      const redirectUri = `${window.location.origin}${location.pathname}?${redirectUrlParams}`;
-
-      // Trigger login
-      keycloakWrapper?.keycloak?.login({ redirectUri: redirectUri });
-    }
-
-    return <Redirect to="/" />;
+    // Trigger login, then redirect to the desired route
+    return <Redirect to={`/login?next=${encodeURIComponent(location.pathname)}`} />;;
```
NickPhura: nit: the keycloak and siteminder redirects use `redirect_uri` and `returl` respectively. Maybe we can change `next` to something more similar: `redirect` or `redirect_uri` seem like good options, just so its more clear what this field is for.

app/src/hooks/useRedirect.tsx
```diff
@@ -0,0 +1,30 @@
+import { useCallback, useMemo } from 'react';
+import { buildUrl } from 'utils/Utils';
+import { useQuery } from './useQuery';
+
+interface Redirect {
+  redirectUri: string | undefined;
+  redirect: () => void;
+}
+
+export default function useRedirect(fallback?: string): Redirect {
```
NickPhura, line 10: Nice 👍

app/src/components/security/RouteGuards.tsx
curtisupshall: Changed to `redirect`.

AlfredRosenthal (approved):
🦞

app/src/components/layout/Header.tsx
curtisupshall: Just to avoid the nagging `{' '}` literal

anissa-agahchen (approved):

PR #1026, opened 2023-05-04
SIMSBIOHUB-24: First Nations as a funding source on projects/ surveys 

AlfredRosenthal:
# Overview

## Links to Jira tickets

[SIMSBIOHUBS-24](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-24)

- First nations will now appear in a searchable drop down when adding a funding source to a project
- API endpoints updated to account for first nations name/ id 
- updated repos to pull/ save first nations name/ id when appropriate
- Updated front end to account for new strings/ types

## PR Testing
1. Pull branch and run migrations
2. Create/ Edit a project
3. Select `Add funding source`
4. Select an agency and you'll see a typeahead for first nations and agencies (start typing first for a quick example)
5. Select an agency, complete the form and save the changes to a project
6. Create/ Edit a survey associated to the previous project
7. Select the funding sources and pick the newly added funding source (this will fail if BioHub isn't running locally as well)
8. Save the changes and reload the page
9. Edit project and modify a funding sources
10. See changes reflected in survey

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/openapi/schemas/project.ts
```diff
@@ -102,6 +102,9 @@ export const projectCreatePostRequestObject = {
               agency_id: {
                 type: 'number'
               },
+              first_nations_id: {
+                type: 'number'
```
curtisupshall: Should be `type: 'integer', minimum: 1`

api/src/paths/project/{projectId}/survey/funding-sources/list.ts
```diff
@@ -61,19 +61,24 @@ GET.apiDoc = {
                   type: 'number'
                 },
                 agency_id: {
-                  type: 'number'
+                  type: 'number',
```
curtisupshall: While we're in here, may as well change them to `type: 'integer', minimum: 1`

app/src/components/fields/AutocompleteField.tsx
```diff
@@ -27,7 +28,7 @@ const AutocompleteField: React.FC<IAutocompleteField<string | number>> = <T exte
   const { touched, errors, setFieldValue, values } = useFormikContext<IAutocompleteFieldOption<T>>();
 
   const getExistingValue = (existingValue: T): IAutocompleteFieldOption<T> => {
-    const result = props.options.find((option) => existingValue === option.value);
+    const result = props.options.find((option) => existingValue === option[`${props.optionFilter || 'value'}`]);
```
curtisupshall: Is it possible to avoid using a template string to key the `option`?

```typescript
const result = props.options.find((option) => existingValue === option[props.optionFilter || 'value']);
```

app/src/features/projects/components/ProjectFundingItemForm.tsx
```diff
@@ -71,17 +126,34 @@ export interface IProjectFundingItemFormProps {
...
+    formValues?: IProjectFundingFormArrayItem
+  ): IAutocompleteFieldOptionWithType<number> | undefined => {
+    if (formValues) {
+      const initialValue = {
+        value: formValues.agency_id ?? formValues.first_nations_id,
+        label: '',
+        type: formValues.agency_id ? FundingSourceType.FUNDING_SOURCE : FundingSourceType.FIRST_NATIONS
+      } as IAutocompleteFieldOptionWithType<number>;
+
+      initialValue.label = `${findItemLabel(initialValue.value, initialValue.type)}`;
```
curtisupshall: Could simply move this inside the `initialValue` declaration

curtisupshall: Could we wrap this in a `String` constructor instead of using a template string (if the goal is to coerce the return value to be a string)?

app/src/features/surveys/view/components/SurveyGeneralInformation.tsx
```diff
@@ -76,7 +77,7 @@ const SurveyGeneralInformation = () => {
             </Grid>
             <Grid item sm={6}>
               <Typography component="dt" color="textSecondary" variant="subtitle2">
-                Anciliary Species
```
curtisupshall, line 79: I've been saying "ancillary" wrong this entire time???

app/src/utils/Utils.ts
```diff
@@ -252,3 +251,30 @@ export const getFormattedIdentitySource = (identitySource: SYSTEM_IDENTITY_SOURC
...
+ * @param {string} property This function will alphabetize the array of objects on the given property
+ * @returns {any[]} Returns an alphabetized array of objects
+ */
+// eslint-disable-next-line no-extend-native
+Array.prototype.alphabetizeObjects = function <T extends { [key: string]: any }>(property: keyof T): T[] {
+  return this.sort((a: T, b: T) => {
+    const nameA = a[property].toUpperCase();
+    const nameB = b[property].toUpperCase();
+
+    if (nameA < nameB) {
```
curtisupshall: Could simply return `nameB - nameA`? That's how I've seen alphabetical sort implemented, at least

app/src/utils/Utils.ts
```diff
@@ -252,3 +251,30 @@ export const getFormattedIdentitySource = (identitySource: SYSTEM_IDENTITY_SOURC
       return null;
   }
 };
+declare global {
```
curtisupshall: Is there any way we could avoid declaring this in the global scope and just import this as a util function on a case-by-case basis? I'm not super well versed on the pitfalls of changing `Array.prototype`, but I think it's generally advised to avoid it (https://eslint.org/docs/latest/rules/no-extend-native); Unless the code needs to run this as an array protoype method? 🤔

database/src/migrations/20230414000001_tr_project_funding_source.ts
```diff
@@ -0,0 +1,60 @@
...
+  -- Purpose: performs specific data validation
+  --
+  -- MODIFICATION HISTORY
+  -- Person           Date        Comments
+  -- ---------------- ----------- --------------------------------------
+  -- charlie.garrettjones@quartech.com
+  --                  2021-01-03  initial release
+  -- charlie.garrettjones@quartech.com
+  --                  2023-04-17  BHBC-2297 adjustment to accommodate first nations
+  -- *******************************************************************
```
curtisupshall, line 30: Could add an entry to this table explaining what has changed

curtisupshall (commented):
Good looks good! Just had a handful of nitpicks

app/src/features/surveys/view/components/SurveyGeneralInformation.tsx
AlfredRosenthal: I've read and re-read that word so many times it's lost all meaning

app/src/features/projects/components/ProjectFundingItemForm.tsx
AlfredRosenthal: I've updated this to keep the object creation altogether instead of splitting it up

app/src/components/fields/AutocompleteField.tsx
AlfredRosenthal: yes it is

app/src/features/projects/components/ProjectFundingItemForm.tsx
AlfredRosenthal: can do. Is there a reason I should be avoiding template strings @curtisupshall

app/src/utils/Utils.ts
AlfredRosenthal: good call, simplified that

app/src/utils/Utils.ts
AlfredRosenthal: it doesn't need to extend the array class, I wanted to be able to chain the methods together so you could for example `[].alphabetizeObjects().map()`. 

An example is in `ProjectParticipantsPage.tsx`, I think it keeps it a little cleaner with the react code but am happy to change that

database/src/migrations/20230414000001_tr_project_funding_source.ts
AlfredRosenthal: Charlie has a comment on it, he also has a draft PR with the migration changes (I just branched off), I could ask him to add a few more details though as it is a little confusing.

app/src/features/projects/components/ProjectFundingItemForm.tsx
curtisupshall: If you're not using it to build strings comprising a programatic part and a literal part (e.g. something like `Hello, ${name}`), it's preferrable to avoid it because:
  - it adds unneccessary computation
  - the syntax makes it easy to add whitespace or other accidental characters
  - in the case where you're just using it to coerce a value into a string, it's not immediately obvious what you're doing. The `String` constructor makes it very clear that you're converting the object to a string (and constructors in Javascript often have better browser support)

NickPhura (dismissed):

database/src/migrations/20230414000001_tr_project_funding_source.ts
curtisupshall: If Charlie is on it, I'm not worried 😌

app/src/utils/Utils.ts
AlfredRosenthal: turned this into a regular function that accepts a generic array of objects and sorts on a given property

api/src/paths/project/{projectId}/survey/funding-sources/list.ts
AlfredRosenthal: done

app/src/features/projects/components/ProjectFundingItemForm.tsx
AlfredRosenthal: I've updated that object creation to use a String()

NickPhura (approved):

NickPhura (approved):

curtisupshall (approved):
Looks good!

PR #1028, opened 2023-05-10
SIMSBIOHUB-24: Bugfix

AlfredRosenthal:
# Overview

## Links to Jira tickets

- refactored UI to better use space
- added logic to hide labels for missing funding source data

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


app/src/features/projects/components/ProjectFundingForm.tsx
```diff
@@ -155,73 +160,95 @@ const ProjectFundingForm: React.FC<IProjectFundingFormProps> = (props) => {
                     )?.[0]?.label;
 
                     return (
-                      <ListItem disableGutters className={classes.fundingListItem} key={index}>
-                        <Paper variant="outlined">
-                          <Toolbar>
+                      <Card key={index} variant="outlined" className={classes.fundingSourceItem}>
```
NickPhura: Just 1 code smell about the key (doesn't like the index, which is valid). Could probably just make the key some combination of funding source name and id, or something like that. Actually that might not work if the same funding source is added twice. In any case, something unique obv.

al-rosenthal: The funding source doesn't have anything that can make it truly be unique. I did update it but if someone was willing they could just fill out the form in the same way. The first nations is a little tricky as well as nothing on the form is required so if they just pick the same group twice we'd still run into the same problem. The DB id could be used but we'd only have that value when users are editing existing projects

NickPhura: Gotcha, ya fair. I think as long as the index makes it unlikely, then its probably fine. In theory, even if they pick the same funding source, it should be because they have different amounts and/or cover different date ranges.

curtisupshall: A fair compromise could be to form a key using a combination of the index, name, id(?) or other information:

```jsx
key={`${fundingSource.id}-${fundingSource.investment_action_category_name}-${index}`}
```

al-rosenthal: spoke to Nick about this and went with a uuid to guarantee unique items

app/src/features/projects/components/ProjectFundingForm.tsx
```diff
@@ -155,73 +161,95 @@ const ProjectFundingForm: React.FC<IProjectFundingFormProps> = (props) => {
                     )?.[0]?.label;
 
                     return (
-                      <ListItem disableGutters className={classes.fundingListItem} key={index}>
-                        <Paper variant="outlined">
-                          <Toolbar>
+                      <Card key={uuidv4()} variant="outlined" className={classes.fundingSourceItem}>
```
curtisupshall: Using a UUID is bad practice for React indecies because they're used to determine how the component rerenders. Since UUIDs are entirely unique between renders, this can cause problems. Can we not make a compound string of the funding source properties and use that as a key?

```javascript
const key = [
  getCodeValueNameByID(props.funding_sources, fundingSource?.agency_id) ||
    getCodeValueNameByID(props.first_nations, fundingSource?.first_nations_id),
  investment_action_category_value
].join('-');

return (
  <Card key={key}>
     //
```

NickPhura (approved):

anissa-agahchen (approved):

PR #1029, opened 2023-05-11
SIMSBIOHUB-32: Sort Project List by Column Headers

curtisupshall:
# Overview
Updates some frontend libraries to allow the use of [MUI DataGrids](https://mui.com/x/react-data-grid/getting-started/), which is used to enable sorting for the Projects list table.

## Links to Jira tickets

 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-32

- Installed the MUI DataGrid library
- Updated the Projects list table to use the MUI DataGrid
- Updates React from version 16 to 17
- Updated `react-scripts` from version 3 to version 5
- Updated numerous packages and fixed tests


app/src/components/fields/CustomFieldArray.tsx
```diff
@@ -69,7 +67,7 @@ export interface CustomFieldArrayProps<ValueType = unknown> {
  * @param {*} props
  * @return {*}
  */
-const CustomFieldArray = <ValueType extends any>(props: CustomFieldArrayProps<ValueType>) => {
+const CustomFieldArray = <ValueType,>(props: CustomFieldArrayProps<ValueType>) => {
```
NickPhura: nit: odd comma?

curtisupshall: The comma is needed; Because it's a .tsx file, the compiler thinks the angle tags are a React element and not a type specifier. We can alternatively use <Value extends any> instead of <Value,>; But we have to be more specific than "any" because "extend any" is a codesmell

NickPhura: Gotcha, read a bit on it, and yeah makes sense. I wonder if we can just extend unknown instead? Would be nice if that works, as its a bit less obscure than the comma workaround.

Also, is that component ever used anywhere? I cant find any references to it.

NickPhura:
(If I remember correctly) When we went to react-scripts 5 in BioHub (~7 months ago) we ran into issues with common node packages not being available (Buffer, Path, Stream, etc) since react-scripts 5 decided to remove them. So we ended up using react-scripts-rewired to add replacement libraries back in (mainly because some of our dependencies were failing because they relied on them).

Curious if we'll need to do the same thing here

app/src/components/fields/CustomFieldArray.tsx
curtisupshall: This component isn't currently used, no. I believe it was copied from BioHub (or maybe the other way around?). I'll try `extends unknown`.

NickPhura (dismissed):

NickPhura (dismissed):
🚗

al-rosenthal (dismissed):
🥇

NickPhura (dismissed):

NickPhura (approved):

al-rosenthal (approved):
🧇

PR #1033, opened 2023-05-15
SIMSBIOHUB-15: Submit Request to Resubmit or Remove Observations, Summary Results or Documents

KjartanE:
# Overview

## Links to Jira tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-15

- [ ] When Logged in to non-admin account on a project
- [ ] If I upload an (Attachment / Observation / Summary ) to Project or Survey Page
- [ ] When I Click the Kebab Icon, I see Option to Remove or Resubmit button
- [ ] When clicked, Dialog for resubmission should display
- [ ] When filled out, and Submit is clicked
- [ ] Then an email should be sent to yourself and the @biohub email 

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/paths/publish/attachment/resubmit.ts
```diff
@@ -0,0 +1,172 @@
...
+
+const defaultLog = getLogger('/api/publish/attachment/resubmit');
+
+export const POST: Operation = [
+  authorizeRequestHandler((req) => {
+    return {
+      or: [
+        {
+          validSystemRoles: [SYSTEM_ROLE.PROJECT_CREATOR],
+          discriminator: 'SystemRole'
```
NickPhura: Creator shouldn't be included here. The only actual thing a creator can do is create a new project. Everything else under a project is controlled by the project level roles.

We should include system admin and data admin though.

api/src/paths/publish/attachment/resubmit.ts
```diff
@@ -0,0 +1,172 @@
...
+
+      A BioHub Administrator should be in contact with you shortly to discuss your request.`
+    };
+
+    const adminEmail = process.env.GCNOTIFY_ADMIN_EMAIL || '';
+    const adminMessage: IgcNotifyRequestRemovalMessage = {
+      ...message,
+      subject: 'Species Inventory Management System -  Request to Remove or Resubmit',
+      header: ''
+    };
```
NickPhura: It would be more ideal if this content could be moved into the service level. 

In general, all we want to do in the endpoint is parse out request data and establish database connections, make one or more service calls, and then handle sending the final response. Everything else is ideally in the service files.

app/src/components/attachments/list/AttachmentsList.test.tsx
```diff
@@ -42,28 +49,73 @@ describe('AttachmentsList', () => {
...
-      />
+      <Router history={history}>
+        <ProjectContext.Provider value={mockProjectContext}>
+          <SurveyContext.Provider value={mockSurveyContext}>
+            <AttachmentsList
+              attachments={attachmentsList}
+              handleDownload={jest.fn()}
+              handleDelete={jest.fn()}
+              handleViewDetails={jest.fn()}
+            />{' '}
```
NickPhura: `{' '}` intentional? Or just funny side effect of the code getting formatted?

NickPhura: Same thing farther down the file.

app/src/components/publish/PublishDialogs.tsx
NickPhura, line 1: The 2 dialogs in this component don't seem to have any overlap or interaction with one another, so I think it would be better to split them out into 2 components and just include them both in the parent. 

Alternatively, just remove 1 of the `ComponentDialog`s and instead, outside of the return, just determine which props need to be set in the dialog at any given moment. 

I think the first option is maybe better though, since then you could even hard-code the text (or hard code which constants it comes from), and make that component really specific to that status.

Additionally (though this depends maybe a bit on the above comment) I would be more explicit about the name of the prop that triggers the dialog to open (call it `open` or something). Otherwise, when looking at the parent, its hard to know which piece of state to follow if you want to trace when the dialog will be opened.

NickPhura: Can chat if its easier

api/src/services/gcnotify-service.ts
```diff
@@ -95,6 +101,44 @@ export class GCNotifyService {
...
+   * @memberof GCNotifyService
+   */
+  async requestRemovalEmailNotification(
+    emailAddress: string,
+    message: IgcNotifyRequestRemovalMessage
+  ): Promise<IgcNotifyPostReturn> {
+    const data: ISendGCNotifyEmailMessage = {
+      email_address: emailAddress,
+      template_id: REQUEST_REMOVAL_TEMPLATE,
+      personalisation: {
```
curtisupshall: Could just pass the `message` object if all the properties match up:

```typescript
const data: ISendGCNotifyEmailMessage = {
  email_address: emailAddress,
  template_id: REQUEST_REMOVAL_TEMPLATE,
  personalisation: { ...message }
};
```

app/src/components/attachments/list/AttachmentsList.test.tsx
curtisupshall: I've noticed this before; Pretty sure it is a formatter artifact.

app/src/components/attachments/list/AttachmentsListItemMenuButton.tsx
```diff
@@ -22,7 +25,15 @@ interface IAttachmentsListItemMenuButtonProps<T extends IGetProjectAttachment |
...
 ) => {
+  const surveyContext = React.useContext(SurveyContext);
+  const projectContext = React.useContext(ProjectContext);
+  const parentName =
+    surveyContext.surveyDataLoader.data?.surveyData.survey_details.survey_name ||
+    projectContext.projectDataLoader.data?.projectData.project.project_name;
+
   const [anchorEl, setAnchorEl] = useState(null);
+  const [openRemoveOrResubmitDialog, setOpenRemoveOrResubmitDialog] = useState(false);
+  const [RemoveOrResubmitDialogFile, setRemoveOrResubmitDialogFile] = useState<T | null>(null);
```
curtisupshall: Nit: camel case `removeOrResubmitDialogFile`

app/src/components/publish/components/AttachmentsFileCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+  submittedDate?: string;
+}
+
+const AttachmentsFileCard = (props: IAttachmentsFileCardProps) => {
+  const classes = useStyles();
+
+  let submittedDate = 'YYYY-MM-DD';
+
+  if (props.submittedDate) {
+    submittedDate = new Date(props.submittedDate).toISOString().split('T')[0];
```
curtisupshall: Could just turn this into a ternary:

```javascript
const submittedDate: Date | string = props.submittedData
 ? new Date(props.submittedDate).toISOString().split('T')[0];
 : 'YYYY-MM-DD';
```

app/src/components/publish/components/AttachmentsFileCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+          </Box>
+        </Box>
+
+        <Box flex="0 0 auto" display="flex" alignItems="center">
+          <Chip
+            title={props.status === PublishStatus.SUBMITTED ? 'SUBMITTED' : 'UNSUBMITTED'}
+            variant="outlined"
+            className={
+              props.status === PublishStatus.SUBMITTED
+                ? clsx(classes.chip, classes.chipSubmitted)
```
curtisupshall: `clsx` supports boolean operators for class names:

```jsx
className={clsx(classes.chip, { [classes.chipSubmitted]: PublishStatus.SUBMITTED, [classes.chipUnSubmitted]: !PublishStatus.SUBMITTED })}
```

Or a little cleaner

```jsx
className={clsx(classes.chip, PublishStatus.SUBMITTED ? classes.chipSubmitted : classes.chipUnSubmitted)}
```

app/src/features/surveys/view/summary-results/components/FileSummaryResults.tsx
```diff
@@ -69,7 +71,15 @@ const useStyles = makeStyles((theme: Theme) => ({
 const FileSummaryResults = (props: IFileResultsProps) => {
   const { fileData, downloadFile, showDelete } = props;
   const classes = useStyles();
+
+  const surveyContext = React.useContext(SurveyContext);
+  const surveyName = surveyContext.surveyDataLoader.data?.surveyData.survey_details.survey_name;
+
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
+  const [openRemoveOrResubmitDialog, setOpenRemoveOrResubmitDialog] = React.useState(false);
+  const [RemoveOrResubmitDialogFile, setRemoveOrResubmitDialogFile] = React.useState<IGetSummaryResultsResponse | null>(
```
curtisupshall: Nit: camel case

app/src/components/dialog/FailureDialog.tsx
```diff
@@ -0,0 +1,40 @@
+import { DialogContentText } from '@material-ui/core';
+import ComponentDialog from 'components/dialog/ComponentDialog';
+import React from 'react';
+
+export interface IPublishDialogs {
+  failureTitle: string;
+  failureMessage: string;
+  failureBody?: string;
+  open: boolean;
+  setOpen: (isEmpty: boolean) => void;
```
NickPhura: Nit: this should be called something like `onClose` because that is the event that will trigger a call to whatever function the parent assigns to this prop. The parent can then decide if they want to pass a function that closes the dialog, or not.

NickPhura:
@KjartanE Looks good, and seems to work as expected. I did push a commit to wrap the resubmit buttons in system role guards (to match the submit buttons/statuses being hidden, for now).

Actually, a really good tech debt ticket might be to add a new type of guard to check for the environment, or even a feature flag env var. Then we could hide elements of the UI based on something other than role. Created a ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-79

NickPhura (approved):

api/src/constants/notifications.ts
```diff
@@ -4,16 +4,16 @@ import { IgcNotifyGenericMessage } from '../services/gcnotify-service';
...
   footer: ''
 };
 
 // Email template for an access request submitter, informing them that their access request has been approved.
 export const ACCESS_REQUEST_SUBMITTER_APPROVAL_EMAIL: IgcNotifyGenericMessage = {
   subject: 'SIMS: Your request for access has been approved.',
   header: 'Your request for access to the Species Inventory Management System has been approved.',
-  body1: `To access the site, `,
-  body2: 'This is an automated message from the BioHub Species Inventory Management System',
+  main_body1: `To access the site, `,
```
curtisupshall, line 16: Nit: trailing space

anissa-agahchen (approved):

api/src/constants/notifications.ts
NickPhura: It looks like this whole const never gets used anywhere, so we could just remove it. I've got some bonus changes unrelated to this PR on my local, so I can just include this deletion there (in the interest of getting more of these PRs through).

PR #1044, opened 2023-06-08
SIMSBIOHUB-73: EML generation update

al-rosenthal:
# Overview
Code from this PR: [PR-192](https://github.com/bcgov/biohubbc-platform/pull/192) depends on this PR

## Links to Jira tickets

[SIMSBIOHUB-73](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-73)

- updated EML service to modify data available in the EML generated

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

### DIFF
--- a/api/src/services/eml-service.ts
+++ b/api/src/services/eml-service.ts
@@ -631,7 +631,7 @@ export class EmlService extends DBService {
     const indigenous = projectData.partnerships.indigenous_partnerships;
     const names = codes.first_nations.filter((code) => indigenous.includes(code.id)).map((code) => code.name);
 
-    const sorted = [...names, ...stakeholders].sort();
+    const sorted = [...names, ...stakeholders].sort((a, b) => a.localeCompare(b));
 
     return {
       describes: projectData.project.uuid,
--- a/api/src/services/eml-service.ts
+++ b/api/src/services/eml-service.ts
@@ -609,7 +609,7 @@ export class EmlService extends DBService {
         }
       });
     }
-      
+
     additionalMetadata.push(await this._buildPartnershipMetadata(projectData));
 
     // add this metadata field so biohub is aware if EML is a project or survey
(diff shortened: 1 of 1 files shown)


anissa-agahchen (dismissed):

api/src/services/eml-service.ts
```diff
@@ -653,6 +625,26 @@ export class EmlService extends DBService {
     return additionalMetadata;
   }
 
+  async _buildPartnershipMetadata(projectData: IGetProject): Promise<any> {
+    const stakeholders = projectData.partnerships.stakeholder_partnerships;
+    const codes = await this.codes();
+    const indigenous = projectData.partnerships.indigenous_partnerships;
+    const names = codes.first_nations.filter((code) => indigenous.includes(code.id)).map((code) => code.name);
+
+    const sorted = [...names, ...stakeholders].sort((a, b) => a.localeCompare(b));
```
NickPhura: Couple of super minor nits:

`names` might be better as `firstNationsNames` or similar, for clarity.
`sorted` might be better as `sortedPartnershipNames` or similar, for clarity.

I'm not super versed on `localeCompare`, but does it do anything different from plain `.sort()` if you don't provide a locale parameter to add language specific sorting rules?

al-rosenthal: so the reason I didn't just leave the default sort was because of a code smell saying "sort will be alphabetical by default" but that's what we're after so I'm going to remove the local compare to avoid confusion and just ignore the smell on this line.

NickPhura (approved):

api/src/services/eml-service.ts
```diff
@@ -653,6 +626,28 @@ export class EmlService extends DBService {
...
 
+  async _buildPartnershipMetadata(projectData: IGetProject): Promise<any> {
+    const stakeholders = projectData.partnerships.stakeholder_partnerships;
+    const codes = await this.codes();
+    const indigenousPartnerships = projectData.partnerships.indigenous_partnerships;
+    const firstNationsNames = codes.first_nations
+      .filter((code) => indigenousPartnerships.includes(code.id))
+      .map((code) => code.name);
+
+    const sortedPartnerships = _.sortBy([...firstNationsNames, ...stakeholders]);
```
curtisupshall, line 637: Nit: Could we not just use `Array.sort`? If not, could we import `sortBy` from Lodash instead of importing the entire object?

curtisupshall (approved):
Code looks good!

api/src/services/eml-service.ts
al-rosenthal: it was originally but the default `.sort()` call is considered a bug by sonar cloud. Sonar cloud requires adding a custom sort algorithm because the default behavior is to sort alphabetically (which is what I wanted). I figured using lodash here was a cleaner solution instead of just replicating the default sort pattern for the sake of sonar cloud.

PR #1046, opened 2023-06-26
SIMSBIOHUB-117: File deletion sends request to BioHub

al-rosenthal:
# Overview

## Links to Jira tickets

[SIMSBIOHUB-117](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-117)

- Delete endpoint modified to send a request to BioHub to deleted submitted files

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

### DIFF
--- a/api/src/paths/project/{projectId}/attachments/{attachmentId}/delete.ts
+++ b/api/src/paths/project/{projectId}/attachments/{attachmentId}/delete.ts
@@ -1,13 +1,9 @@
 import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
-import { ATTACHMENT_TYPE } from '../../../../../constants/attachments';
 import { PROJECT_ROLE, SYSTEM_ROLE } from '../../../../../constants/roles';
 import { getDBConnection } from '../../../../../database/db';
 import { authorizeRequestHandler } from '../../../../../request-handlers/security/authorization';
 import { AttachmentService } from '../../../../../services/attachment-service';
-import { HistoryPublishService } from '../../../../../services/history-publish-service';
-import { PlatformService } from '../../../../../services/platform-service';
-import { deleteFileFromS3 } from '../../../../../utils/file-utils';
 import { getLogger } from '../../../../../utils/logger';
 import { attachmentApiDocObject } from '../../../../../utils/shared-api-docs';
 
@@ -105,31 +101,14 @@ export function deleteAttachment(): RequestHandler {
       await connection.open();
 
       const attachmentService = new AttachmentService(connection);
-      const historyPublishService = new HistoryPublishService(connection);
 
-      let deleteResult: { key: string; uuid: string };
-
-      if (req.body.attachmentType === ATTACHMENT_TYPE.REPORT) {
-        await historyPublishService.deleteProjectReportAttachmentPublishRecord(Number(req.params.attachmentId));
-        await attachmentService.deleteProjectReportAttachmentAuthors(Number(req.params.attachmentId));
-
-        deleteResult = await attachmentService.deleteProjectReportAttachment(Number(req.params.attachmentId));
-      } else {
-        await historyPublishService.deleteProjectAttachmentPublishRecord(Number(req.params.attachmentId));
-        deleteResult = await attachmentService.deleteProjectAttachment(Number(req.params.attachmentId));
-      }
-
-      const platformService = new PlatformService(connection);
-      // request BIOHUB API to delete attachment
-      await platformService.deleteAttachmentFromBiohub(deleteResult.uuid);
+      await attachmentService.handleDeleteProjectAttachment(
+        Number(req.params.projectId),
+        Number(req.params.attachmentId),
+        req.body.attachmentType
+      );
 
       await connection.commit();
-
-      const deleteFileResult = await deleteFileFromS3(deleteResult.key);
-
-      if (!deleteFileResult) {
-        return res.status(200).json(null);
-      }
 
       return res.status(200).send();
     } catch (error) {
--- a/api/src/paths/project/{projectId}/attachments/{attachmentId}/delete.ts
+++ b/api/src/paths/project/{projectId}/attachments/{attachmentId}/delete.ts
@@ -2,7 +2,8 @@ import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
 import { PROJECT_ROLE, SYSTEM_ROLE } from '../../../../../constants/roles';
 import { getDBConnection } from '../../../../../database/db';
-import { authorizeRequestHandler } from '../../../../../request-handlers/security/authorization';
+import { UserObject } from '../../../../../models/user';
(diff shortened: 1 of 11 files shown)


app/src/components/attachments/list/AttachmentsListItemMenuButton.tsx
```diff
@@ -85,21 +85,21 @@ const AttachmentsListItemMenuButton = (props: IAttachmentsListItemMenuButtonProp
                 View Details
               </MenuItem>
             )}
-            {props.attachmentStatus === PublishStatus.UNSUBMITTED && (
```
NickPhura, line 88: Wouldn't this still apply for non admin users?

KjartanE: image
We have a separate button that is unreleased currently for request to resubmit/remove.
So pure delete seemed like it was a admin only button

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/{attachmentId}/delete.ts
```diff
@@ -123,6 +124,10 @@ export function deleteAttachment(): RequestHandler {
         deleteResult = await attachmentService.deleteSurveyAttachment(Number(req.params.attachmentId));
       }
 
+      const platformService = new PlatformService(connection);
+      // request BIOHUB API to delete attachment
+      await platformService.deleteAttachmentFromBiohub(deleteResult.uuid);
```
NickPhura: We'll need to check, in the case the user is a non-admin user (project lead or editor), if they are allowed to delete the file. If its published, then no, otherwise yes.

KjartanE: image
The endpoints are already restricted by role guards. Should be be double checking?

KjartanE: Oh I see, lead and editor do not have permissions to delete from Backbone

curtisupshall (dismissed):
Code looks good 🦡

api/src/paths/project/{projectId}/attachments/{attachmentId}/delete.ts
```diff
@@ -104,27 +102,19 @@ export function deleteAttachment(): RequestHandler {
...
 
       const attachmentService = new AttachmentService(connection);
-      const historyPublishService = new HistoryPublishService(connection);
 
-      let deleteResult: { key: string };
-      if (req.body.attachmentType === ATTACHMENT_TYPE.REPORT) {
-        await historyPublishService.deleteProjectReportAttachmentPublishRecord(Number(req.params.attachmentId));
-        await attachmentService.deleteProjectReportAttachmentAuthors(Number(req.params.attachmentId));
+      const systemUserObject: UserObject = req['system_user'] || (await getSystemUserObject(connection));
+      const isAdmin = systemUserObject.role_names.includes(SYSTEM_ROLE.SYSTEM_ADMIN);
```
NickPhura: I think this can include data admin as well?

api/src/repositories/attachment-repository.ts
```diff
@@ -369,6 +369,50 @@ export class AttachmentRepository extends BaseRepository {
...
+        survey_attachment
+      WHERE
+        survey_attachment_id = ${attachmentId}
+      AND
+        survey_id = ${surveyId};
+    `;
+
+    const response = await this.connection.sql<ISurveyAttachment>(sqlStatement);
+
+    if (!response.rows) {
```
NickPhura: rows will always be truthy. So we should either remove this, or change the assertion to rowCount === 1 if we are expecting exactly 1 result

api/src/services/attachment-service.ts
```diff
@@ -828,4 +848,112 @@ export class AttachmentService extends DBService {
 
     return { ...attachmentResult, key };
   }
+
+  /**
+   * Handle deletion of Project Attachment.
+   * If attachment is a report, delete authors
+   * Otherwise, delete the publish record and the attachment.
+   * If attachment was published, delete the publish record and delete from Platform.
```
NickPhura: This doc is a bit confusing. "If report, delete authors, otherwise delete publish record and attachment". We want to delete the publish record in both cases regardless of attachment type.

One way to write this out is just to repeat yourself for each type:

```
If report:
 - Delete stuff
 - Delete authors
 - Delete publish record, if one exists
if not report:
 - Delete stuff
 - Delete publish record, if one exists
```

api/src/repositories/attachment-repository.ts
```diff
@@ -369,6 +369,50 @@ export class AttachmentRepository extends BaseRepository {
...
+        survey_attachment
+      WHERE
+        survey_attachment_id = ${attachmentId}
+      AND
+        survey_id = ${surveyId};
+    `;
+
+    const response = await this.connection.sql<ISurveyAttachment>(sqlStatement);
+
+    if (response.rowCount === 1) {
```
NickPhura: Should this be !== 1?

NickPhura: could also update the error message as well, which still says != null

api/src/services/attachment-service.ts
```diff
@@ -828,4 +848,126 @@ export class AttachmentService extends DBService {
...
+   * Handle deletion of Project Attachment.
+   * If report:
+   * - delete authors
+   * - delete publish record
+   * - delete attachment
+   * If attachment:
+   * - delete publish record
+   * - delete attachment
+   * If attachment was published:
+   * - delete from Platform.
```
NickPhura: Sorry, one last nit: The bottom 2 sections both say "If attachment" which makes it seem like the last point (about deleting from the platform) doesn't apply to reports, when it does.

I can help come up with other language if you want, but otherwise all I can recommend is to try and write this more verbosely (comments are free, so no need to be concise) and try and write them from the perspective of someone that doesn't know anything about SIMS attachments or the fact that we have 2 main types (Reports and Non-Reports).


The two kinds of attachments and even the relationship between deleting from SIMS and BioHub are business concepts, rather than technical concepts, so being really comment heavy is always worth it.

NickPhura: That being said, the inline commends in the function below are great, and exactly what we should strive for in any function that has some more complex if/else operations, especially if it involves business logic that isn't necessarily obvious to an outsider, which this has.

curtisupshall (approved):

KjartanE (approved):
🐈

api/src/services/attachment-service.ts
```diff
@@ -828,4 +848,130 @@ export class AttachmentService extends DBService {
...
   }
+
+  /**
+   * Handle deletion of Project Attachment.
+   *
+   * If (attachmentType = report):
+   * - delete authors
+   * - delete publish record
+   * - delete attachment
+   * Else (attachmentType = attachment):
```
NickPhura, line 859: I think the type is `Other` technically

PR #1048, opened 2023-06-29
SIMSBIOHUB-127: Update inferred layers code

NickPhura:
# Overview

## Links to Jira tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-127

- Add backend services for fetching data from the BCGW service
- Add new endpoint for fetching inferred region information (moving away from the frontend making these calls, because we probably need to backend to also make these calls, and don't want to duplicate the same code in both places).
- Some misc updates to a geojson type name.

## TODO
- Write tests (api/app)


api/src/services/bcgw-layer-service.ts
```diff
@@ -0,0 +1,460 @@
...
+    const regionDetails = this.findRegionDetails(feature);
+
+    if (!regionDetails) {
+      // Feature is not a known BCGW feature
+      // Fetch region details for the feature from ALL available layers
+      response = response.concat(
+        await this.getAllRegionDetailsForWktString(geometryWKTString, [
+          BcgwEnvRegionsLayer,
+          BcgwNrmRegionsLayer
+          // BcgwParksAndEcoreservesLayer,
```
curtisupshall: Any reason these commented lines should stay?

api/src/services/bcgw-layer-service.ts
```diff
@@ -0,0 +1,460 @@
...
+      response = response.concat(mappedRegionDetails);
+      // Fetch region details for the feature, excluding the layer whose details were already added above
+      // Why? We want to avoid querying a layer using a feature from that same layer because the actual results will not
+      // be consistent with the expected results. (Ex: LayerA + Feature1 should return Feature1, but will often return
+      // Feature1 + Feature2 + Feature3 due to the layer features containing overlapping coordinates, when ideally they
+      // should not).
+      const layersToProcess = [
+        BcgwEnvRegionsLayer,
+        BcgwNrmRegionsLayer
+        //   BcgwParksAndEcoreservesLayer,
```
curtisupshall: Likewise.

api/src/services/bcgw-layer-service.ts
```diff
@@ -0,0 +1,460 @@
...
+   * @return {*}
+   * @memberof BcgwLayerService
+   */
+  async getAllRegionDetailsForWktString(geometryWktString: string, layersToProcess: string[]) {
+    let response: RegionDetails[] = [];
+
+    for (const layerToProcess of layersToProcess) {
+      switch (layerToProcess) {
+        case BcgwEnvRegionsLayer:
+          response = response.concat(await this.getEnvRegionDetails(geometryWktString));
```
curtisupshall, line 413: Could we improve readibility by using `response.push` instead of `response = response.concat`?

api/src/services/bcgw-layer-service.ts
```diff
@@ -0,0 +1,460 @@
...
+          break;
+        default:
+          break;
+      }
+    }
+
+    return response;
+  }
+
+  async getEnvRegionDetails(geometryWktString: string): Promise<RegionDetails[]> {
```
curtisupshall, line 439: Todo: jsdoc

api/src/services/bcgw-layer-service.ts
curtisupshall: Nevermind, I misunderstood what was happening 😅

api/src/services/bcgw-layer-service.ts
NickPhura: Yeah, its because I was bouncing between using those layers or not. They was support for them in the frontend, but evidently it was disable or something. I moved that all to the API. But when the matching parks are returned, you get a massive list (like 40 items) if the boundary is an NRM region. So it seemed like that wasn't going to be helpful, even though it technically works, so I just commented them out. But, could remove them fully probably.

NickPhura: Removed.

api/src/services/bcgw-layer-service.ts
NickPhura: Removed.

api/src/services/bcgw-layer-service.ts
NickPhura: Added.

api/src/paths/spatial/regions.ts
```diff
@@ -0,0 +1,151 @@
...
+export function getRegions(): RequestHandler {
+  return async (req, res) => {
+    const connection = getDBConnection(req['keycloak_token']);
+
+    try {
+      const features = req.body.features as Feature[];
+
+      let regionsDetails: RegionDetails[] = [];
+
+      try {
```
al-rosenthal: is there a reason to nest this try catch? should it be throwing a custom error compared to the the other try catch?

api/src/services/bcgw-layer-service.ts
```diff
@@ -0,0 +1,480 @@
...
+   * matching NRM regions for a given ENV region, and nice versa, without having to query the BCGW and more importantly,
+   * without running into any inconsistent layer issues (ex: NRM Region 1 should map to ENV Region 2, but due to
+   * the polygons in the layers overlapping, NRM Region 1 will return matches for ENV Region 2 AND ENV Region 3, when
+   * it shouldn't). As a result we cannot reply fully on the layer query results alone.
+   *
+   * @param {RegionDetails[]} regionDetails
+   * @return {*}
+   * @memberof BcgwLayerService
+   */
+  getMappedRegionDetails(regionDetails: RegionDetails[]) {
```
al-rosenthal: nit: a return type would be great here

api/src/services/postgis-service.ts
```diff
@@ -0,0 +1,26 @@
...
+export class PostgisService extends DBService {
+  postgisRepository: PostgisRepository;
+
+  constructor(connection: IDBConnection) {
+    super(connection);
+    this.postgisRepository = new PostgisRepository(connection);
+  }
+
+  /**
+   * Convert the provided geometry into Well-Known Text (WKT) in the provided Spatial Reference ID (SRID).
```
al-rosenthal, line 16: would be great if there was a link explaining what WKT is in this context

api/src/services/postgis-service.ts
```diff
@@ -0,0 +1,26 @@
...
+   *
+   * @param {Geometry} geometry
+   * @param {Srid} srid
+   * @return {*}
+   * @memberof PostgisRepository
+   */
+  async getGeoJsonGeometryAsWkt(geometry: Geometry, srid: Srid) {
```
al-rosenthal: nit: a return type would be great

al-rosenthal (commented):
ran this locally, seems to work great

api/src/paths/spatial/regions.ts
NickPhura: In an earlier version, I had more of the code in the endpoint, and so I was wrapping just the one section that needed a DB Connection. But since refactoring it we totally don't need the extra try catch. Will remove

api/src/services/postgis-service.ts
NickPhura: Added.

api/src/services/postgis-service.ts
NickPhura: Added.

api/src/services/bcgw-layer-service.ts
NickPhura: Added

api/src/paths/spatial/regions.ts
NickPhura: Removed

al-rosenthal (dismissed):
🗺️

api/src/paths/spatial/regions.ts
```diff
@@ -0,0 +1,147 @@
...
+      for (const feature of features) {
+        const result = await bcgwLayerService.getRegionsForFeature(feature, connection);
+        regionsDetails = regionsDetails.concat(result);
+      }
+
+      await connection.commit();
+
+      // Convert array first into JSON, then into Set, then back to array in order to
+      // remove duplicate region information.
+      const regionDetailsJson = regionsDetails.map((value) => JSON.stringify(value));
```
curtisupshall, line 131: @NickPhura Had to change this to perform the `JSON.stringify` conversion, since a `Set` can only determine duplicates for primitives. Thoughts?

curtisupshall (dismissed):
Code looks good! +1 for added type coverage 👏

api/src/paths/spatial/regions.ts
NickPhura: Yeah I saw that, good catch on what I had before which was totally not working.

I think this works just fine atm. I hadn't really thought too much about it tbh. The only caveat would be if we imagine a situation where the object as a whole isn't valid for filtering duplicates (after adding some additional fields to it in future). Might never happen, but that is the only situation I can think of where this might break. Though, not sure what those fields would be.. or why they would be different for each instance of an object with the same region/layer name.. so maybe that isn't a valid scenario that would ever happen.

I found a stackoverflow post discussing this issue (https://stackoverflow.com/questions/2218999/how-to-remove-all-duplicates-from-an-array-of-objects), and the one upvoted solution seems to address that by just checking specific values in the object. I did try in a sandbox and it does work, not something i would have thought up on my own haha.

```
regionsDetails = regionsDetails.filter((value, index, self) =>
    index === self.findIndex((item) => (
        item.regionName === value.regionName && item.sourceLayer === value.sourceLayer
    ))
)
```

al-rosenthal (approved):
🐤

curtisupshall (approved):

PR #1058, opened 2023-07-24
SIMSBIOHUB-149: Update Project Type -> Project Programs and Project activity -> Project Types

al-rosenthal:
# Overview

## Links to Jira tickets

[SIMSBIOHUB-149](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-149)

- Updated UI to show Programs instead of types
- Updated API/ Services/ Repo to handle Programs
- Added migration to create Program and join table and remove type column from projects
- Added migration to move existing types into new programs join table 

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


NickPhura:
Noticed one odd behaviour: If I remove ALL activities, and click save, it doesn't persist (the old values are retained). 

Edit: Tried this in Dev and it does the same thing, so must have been an existing bug. Maybe something we can fix while we're in that code anyways.


Also, I noticed the ProjectAdvancedFilters component, and a few other api/app files, still mention "project_type"

KjartanE (dismissed):
Ran locally, works well! 🐈

database/src/migrations/20230719141800_project_type_to_program.ts
```diff
@@ -0,0 +1,69 @@
+import { Knex } from 'knex';
+
+/**
+ * // add new table for region lookup and connection table to project and survey
```
NickPhura: nit

database/src/migrations/20230719172000_project_type_clean_up.ts
```diff
@@ -0,0 +1,42 @@
+import { Knex } from 'knex';
+
+/**
+ * // add new table for region lookup and connection table to project and survey
```
NickPhura: nit

database/src/migrations/20230719141800_project_type_to_program.ts
```diff
@@ -0,0 +1,69 @@
...
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+	set search_path= biohub, public;
+
+
+    CREATE TABLE program (
```
NickPhura, line 18: nit: Add comments for this table

database/src/migrations/20230719141800_project_type_to_program.ts
```diff
@@ -0,0 +1,69 @@
...
+      program_id              integer                 GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+      name                    varchar(50),
+      description             varchar(50),
+      record_effective_date   date,
+      record_end_date         date,
```
NickPhura, line 23: Will need to add that 'unique end-date constraint' for this table. Examples in that readme in my PR.

Would be on (name, record_end_date) so that you can't have 2 entries with the same name (unless the record_end_date is not null)

database/src/migrations/20230719172000_project_type_clean_up.ts
NickPhura: Also, wondering if this cleanup migration should be in the same migration file as the table creation, since they are 2 steps of the same overall process. 

Not sure if the grouped sql is better or if splitting it out is nicer. Thoughts?

database/src/migrations/20230719172000_project_type_clean_up.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set search_path= biohub, public;
+    
+    -- Get existing types and insert them into programs in order
+    INSERT INTO "program" (name, record_effective_date, description, record_end_date)
+    SELECT name, record_effective_date, description, record_end_date
+    FROM project_type
+    ORDER BY project_type_id;
```
NickPhura: Any reason for the order by? I assume it wouldn't make a difference either way, overall?

Edit: NVM I follow now :)

database/src/migrations/20230719141800_project_type_to_program.ts
al-rosenthal: done

database/src/migrations/20230719172000_project_type_clean_up.ts
al-rosenthal: as per our discussion, moved those into a single migration

database/src/migrations/20230719141800_project_type_to_program.ts
al-rosenthal: done

database/src/migrations/20230719141800_project_type_to_program.ts
al-rosenthal: added

NickPhura (dismissed):

NickPhura (dismissed):

KjartanE (dismissed):
🐱

NickPhura (approved):

PR #1060, opened 2023-07-27
SIMSBIOHUB-154: System user table updates

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-154?

## Description of Changes

- Add migration to:
  - Add new system user permission tables
  - Add new fields to system_user table
  - Add new identity source type: UNVERIFIED
    - Used to classify a user with no IDIR/BCeID information (no verified credential)
- Remove unused `SYSTEM` user identity source type (API/APP)
- Update user seed
  - Account for new system_user fields
- Update database readme

## Testing Notes

- Migration runs successfully
- Application functions as normal (new tables should not impact current functionality)


database/src/migrations/20230727000000_system_user_enhancement.ts
```diff
@@ -0,0 +1,269 @@
...
+      create_date                timestamptz(6)    DEFAULT now() NOT NULL,
+      create_user                integer           NOT NULL,
+      update_date                timestamptz(6),
+      update_user                integer,
+      revision_count             integer           DEFAULT 0 NOT NULL,
+      CONSTRAINT system_permission_pk PRIMARY KEY (system_permission_id)
+    );
+
+    COMMENT ON COLUMN system_permission.system_permission_id       IS 'System generated surrogate primary key identifier.';
+    COMMENT ON COLUMN system_user_info.name                        IS 'The name of the project system permission role.';
```
al-rosenthal: looks like a copy paste issue

database/src/migrations/20230727000000_system_user_enhancement.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    COMMENT ON COLUMN system_user_role.system_user_role_id   IS 'System generated surrogate primary key identifier.';
+    COMMENT ON COLUMN system_user_role.system_user_id        IS 'System generated surrogate primary key identifier.';
+    COMMENT ON COLUMN system_user_role.system_role_id        IS 'System generated surrogate primary key identifier.';
+    COMMENT ON COLUMN system_user_role.create_date           IS 'The datetime the record was created.';
+    COMMENT ON COLUMN system_user_role.create_user           IS 'The id of the user who created the record as identified in the system user table.';
+    COMMENT ON COLUMN system_user_role.update_date           IS 'The datetime the record was updated.';
+    COMMENT ON COLUMN system_user_role.update_user           IS 'The id of the user who updated the record as identified in the system user table.';
+    COMMENT ON COLUMN system_user_role.revision_count        IS 'Revision count used for concurrency control.';
+    COMMENT ON TABLE  system_user_role                       IS 'A associative entity that joins project_group participation and project_group role types.';
```
al-rosenthal: looks like more copy paste errors from the group work, this shows up in a all table comments

al-rosenthal (dismissed):
👤

al-rosenthal (dismissed):
ran this locally and it worked after editing the user name ⚡

api/src/services/user-service.ts
```diff
@@ -227,4 +234,60 @@ export class UserService extends DBService {
...
+        user_guid: keycloakUserInformation.bceid_user_guid,
+        user_identifier: keycloakUserInformation.bceid_username,
+        user_identity_source: SYSTEM_IDENTITY_SOURCE.BCEID_BUSINESS,
+        display_name: keycloakUserInformation.display_name,
+        email: keycloakUserInformation.email,
+        given_name: keycloakUserInformation.given_name,
+        family_name: keycloakUserInformation.family_name,
+        agency: keycloakUserInformation.bceid_business_name
+      };
+    } else {
```
al-rosenthal: is it safe to assume we'll hit one of these types?

api/src/repositories/user-repository.ts
```diff
@@ -407,4 +408,56 @@ export class UserRepository extends BaseRepository {
...
   }
+
+  /**
+   * Update a system user's information.
+   *
+   * Note: Attempts to find a system user record using the `user_guid` first, and if no record is found, attempts
+   * using the `user_identifier` + `user_identity_source`. If a record is found, the record is updated and the
+   * system user id is returned. If no user is found, does nothing and returns null.
+   *
+   * @param {({
```
al-rosenthal: it would be nice to see this in an interface that can be used in the service as well

al-rosenthal (commented):
I also noticed an issue under manage users tabs:
- reviewing a request fails
- adding a user fails

api/src/repositories/user-repository.ts
NickPhura: Yeah I can make one, was just being lazy

api/src/services/user-service.ts
NickPhura: It should be, if the upstream calls are being typesafe.

If you've generated the `keycloakUserInformation` object using the `getKeycloakUserInformationFromKeycloakToken` function (which just calls the zod schema on the raw token), it should be safe to assume it will match one of these here.

NickPhura:
@al-rosenthal Pushed up some changes to fix the manual system/project member add functionality. The UI is a bit ugly, but that is going to change anyways.

api/src/services/user-service.ts
al-rosenthal: ok great

al-rosenthal (dismissed):
ran this locally and things worked as expected 🌻

KjartanE (dismissed):
🐈

KjartanE (dismissed):
🐈

api/src/paths/user/add.ts
```diff
@@ -111,6 +121,14 @@ export function addSystemRoleUser(): RequestHandler {
       throw new HTTP400('Missing required body param: identitySource');
     }
 
+    if (!displayName) {
+      throw new HTTP400('Missing required body param: identitySource');
+    }
+
+    if (!email) {
```
curtisupshall, line 128: Could we not just assert this in the API schema, rather than explicitly check the value?

I know that we do these additional checks for other fields and in other files. Is there any reason we would prefer one strategy over the other?

api/src/services/keycloak-service.ts
```diff
@@ -22,9 +22,20 @@ type BCEIDBusinessAttributes = BCEIDBasicAttributes & {
...
-interface KeycloakGetUserResponse {
-  users: KeycloakUser[];
-  roles: Record<string, string>[];
+interface KeycloakIDIRUserRecord {
+  username: string;
+  firstName: string;
+  lastName: string;
+  email: string;
+  attributes: {
+    idir_user_guid: string[];
```
curtisupshall, line 31: Will we only ever expect one element in this array? If so, we could use:

```javascript
attributes: {
    idir_user_guid: [string];
    idir_username: [string];
    display_name: [string];
}
```

api/src/services/keycloak-service.ts
```diff
@@ -38,22 +49,29 @@ export type KeycloakUser = {
...
- * Keycloak Documentation (select version and see `Administration REST API` section):
- * - https://www.keycloak.org/documentation-archive.html
+ * API Swagger Doc: https://api.loginproxy.gov.bc.ca/openapi/swagger#/Users/get__environment__basic_business_bceid_users
  *
  * @export
  * @class KeycloakService
  */
 export class KeycloakService {
-  keycloakTokenHost: string;
+  // Used to authenticate with the CSS API using the SIMS API credentials
```
curtisupshall, line 60: +1 for labelling all the keycloak variables 👍

api/src/services/keycloak-service.ts
```diff
@@ -85,49 +101,41 @@ export class KeycloakService {
...
       throw new ApiGeneralError('Failed to authenticate with keycloak', [(error as Error).message]);
     }
   }
 
   /**
-   * Fetch keycloak user data by the keycloak username.
-   *
-   * Note on IDIR and BCEID usernames:
-   * - Format is `<username>@idir` or `<username>@bceid`
+   * Search for keycloak IDIR users based on provided criteria.
```
curtisupshall, line 110: Is this a boolean search where the criteria is `AND`ed together? Could specify in the docs.

curtisupshall: This method seems to ignore most of the criteria. I'm assuming this is future-proofing?

api/src/utils/keycloak-utils.ts
```diff
@@ -1,40 +1,112 @@
...
+  given_name: z.string(),
+  family_name: z.string(),
+  email: z.string()
+});
+export type IdirUserInformation = z.infer<typeof IdirUserInformation>;
+
+/**
+ * User information from a verified BCeID Basic Keycloak token.
+ */
+export const BceidBasicUserInformation = z.object({
```
curtisupshall, line 24: Any reason we don't take advantage of `v.extend`?

api/src/utils/keycloak-utils.ts
```diff
@@ -71,20 +140,63 @@ export const coerceUserIdentitySource = (userIdentitySource: string | null): SYS
...
     return null;
   }
 
-  return userIdentifier;
+  return result.data;
+};
+
+export const isIdirUserInformation = (
+  keycloakUserInformation: KeycloakUserInformation
+): keycloakUserInformation is IdirUserInformation => {
```
curtisupshall, line 182: +1 for `is` semantics

api/src/utils/keycloak-utils.ts
```diff
@@ -71,20 +140,63 @@ export const coerceUserIdentitySource = (userIdentitySource: string | null): SYS
...
+export const isBceidBusinessUserInformation = (
+  keycloakUserInformation: KeycloakUserInformation
+): keycloakUserInformation is BceidBusinessUserInformation => {
+  return keycloakUserInformation.identity_provider === SYSTEM_IDENTITY_SOURCE.BCEID_BUSINESS.toLowerCase();
+};
+
+export const isDatabaseUserInformation = (
+  keycloakUserInformation: KeycloakUserInformation
+): keycloakUserInformation is DatabaseUserInformation => {
+  return keycloakUserInformation.identity_provider === SYSTEM_IDENTITY_SOURCE.DATABASE.toLowerCase();
```
curtisupshall, line 201: I'm assuming keycloak currently provides this value as a lowercase string. But is there a possible edge case where we may expect `keycloakUserInformation.identity_provider` to be lowercase but it isn't? That would cause this check to erronesouly fail

api/src/paths/user/add.ts
NickPhura: The explicit checks are just older, and likely not needed. There may be some rare situations where a check is needed, because its doing something openapi doesnt support, but for most of them these manual checks can be removed. I mainly just matched the old pattern when I added these 2 new fields, because this particular change is temporary (the user stuff is all going to change real soon, and I just added these to make testing easier).

api/src/services/keycloak-service.ts
NickPhura: It might be the case that it is only ever 1 item, though we can't be certain, but these are based on the keycloak teams swagger, and they unfortunately don't specify anything more precise than just an array of strings.

api/src/services/keycloak-service.ts
NickPhura: Yeah, there are more keycloak env vars now (or soon will be):  Regular user vars, service client vars, css api vars.

api/src/services/keycloak-service.ts
NickPhura: Good question. I believe its AND, though the doc doesn't explicitly say, but the search results seem to indicate that.

And yeah, just future proofing. These are all of the criteria the CSS API supports, even though we are only using the guid atm. Could probably just remove them as we may never search on the other fields.

api/src/utils/keycloak-utils.ts
NickPhura: I included a call to lowercase in the zod schema, so if used correctly, this identity_provider field should always be lowercase already

api/src/utils/keycloak-utils.ts
NickPhura: Umm, not in particular. I was just thinking it is a bit less hard to read if its all just spelled out plainly. Extending really just saves a few lines of code, in this 1 file, and wouldn't impact how anyone uses them elsewhere.

NickPhura: Probably a bit subjective really. I know my original keycloak types were extending one another, but it might be best just to keep things simple in general. The zod stuff adds another layer of complexity too, having both a zod object and the type.

app/src/features/admin/users/AddSystemUsersForm.tsx
```diff
@@ -80,6 +88,30 @@ const AddSystemUsersForm: React.FC<AddSystemUsersFormProps> = (props) => {
                             }}
                           />
                         </Box>
+                        <Box pl={1} width={220}>
```
curtisupshall, line 91: Seem to be missing padding on the right:

image

curtisupshall (commented):
I added my alt BCeID Basic account (putting `curtisupshalldev` into the **Username** field), and the account was added as a System Administrator successfully. Then I logged into said account, but I was denied access. I then request access from my `curtisupshalldev` account, and approved it as a Data Administrator. So now there are two accounts with the same username - one that I explicitly added, and one that I approved.


image

app/src/features/admin/users/AddSystemUsersForm.tsx
NickPhura: Yeah this is super temporary, just to make some testing easier. All of these manual user controls are going to be drastically changed in the next few sprints

curtisupshall (dismissed):
Code looks good; tested adding a user and it locally and works as expected 👑

al-rosenthal (dismissed):
🕶️

al-rosenthal (dismissed):
🧄

al-rosenthal (approved):
🧯

curtisupshall (approved):

PR #1061, opened 2023-07-27
SIMSBIOHUB-146: Model Clean up

al-rosenthal:
## Links to Jira Tickets

[SIMSBIOHUB-146](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-146)

## Description of Changes
- Added new migration to delete climate initiative tables
- Migration also creates a new `comments` column on the `survey`
- Migration also moves any `caveats` on a project to the new `comments` column on surveys
- Migration also removed `caveats` column from the `project` table
- Added migrations to update project/ survey api delete functions

## Testing Notes


curtisupshall (commented):
Code looks good! Testing locally.

curtisupshall (approved):
Tested locally! Works as expected. 👍

database/src/migrations/20230726080300_clean_project_relations.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    -------------------------------------------------------------------------
+    -- Remove old views
+    -------------------------------------------------------------------------
+    SET search_path = ${DB_SCHEMA_DAPI_V1};
```
NickPhura, line 18: Doesn't really matter either way, but you could replace these two schema vars with the hardcoded strings.

The intended purpose never panned out for various reasons, so it doesn't matter if we hardcode them.

NickPhura (approved):

database/src/migrations/20230726080300_clean_project_relations.ts
al-rosenthal: I've got another PR that I'll replace these on

PR #1063, opened 2023-07-31
Tech debt: Fixed linting errors and skipped tests

curtisupshall:
Fixes frontend linting errors and skipped tests.


al-rosenthal (dismissed):
🍡

NickPhura:
@curtisupshall 

Found an issue: If you edit the map, and select a layer from the dropdown (nrm regional boundaries in my case), you get an infinite loop of requests to the openmaps service. I confirmed this is not the case in Dev. It looks like `throttledGetFeatures` is getting called in a loop.

al-rosenthal (dismissed):
ran this locally and it works as expected
🐇

NickPhura:
@curtisupshall Maps seem happier after the latest changes from what I can tell.

There is still 1 dependency error popping up though, in ProjectAdvancedFilters.tsx L91.

I did give it a try locally, and adding the dependency (biohubApi.taxonomy) it is complaining about seems to work just fine (doesn't trigger any loops or anything).

NickPhura (approved):

KjartanE (approved):
🐈

al-rosenthal (approved):
🗺️

PR #1071, opened 2023-08-18
SIMSBIOHUB - 241 + 242 + 253 - Critter / Observation / Deployment / Survey Stratum tables

MacDeluca:
## Links to Jira Tickets

- [https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-241](url)

## Description of Changes

- New critter table
- New observation table
- New deployment table
- New survey_stratum table
- New trigger for monitoring total critter count to be less than observation total count

## Testing Notes 
- Included simple test sql script on ticket 241
[https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-241](url)
- Only allowed to add as many critters per observation_id as the total_count
- Unique index on critterbase_critter_id (this id is a external Critterbase critter_id)
-

### DIFF
--- a/database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
+++ b/database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
@@ -71,7 +71,7 @@ export async function up(knex: Knex): Promise<void> {
   COMMENT ON COLUMN critter.revision_count           IS 'Revision count used for concurrency control.';
   COMMENT ON TABLE  critter                          IS 'Information about individual animals associated to specific observations.';
 
-  -- Add foreign key constraint from child table to parent table on <column_name_2>
+  -- Add foreign key constraint from child table to parent table on observation_id
   ALTER TABLE critter ADD CONSTRAINT critter_fk1
     FOREIGN KEY (observation_id)
     REFERENCES observation(observation_id);
@@ -82,6 +82,44 @@ export async function up(knex: Knex): Promise<void> {
   -- Create audit and journal triggers
   create trigger audit_critter before insert or update or delete on critter for each row execute procedure tr_audit_trigger();
   create trigger journal_critter after insert or update or delete on critter for each row execute procedure tr_journal_trigger();
+
+  ----------------------------------------------------------------------------------------
+  -- Create Deployment table
+  ----------------------------------------------------------------------------------------
+
+  CREATE TABLE deployment(
+    deployment_id            integer           GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+    critter_id               integer           NOT NULL,
+    bctw_deployment_id       uuid              NOT NULL,
+    create_date              timestamptz(6)    DEFAULT now() NOT NULL,
+    create_user              integer           NOT NULL,
+    update_date              timestamptz(6),
+    update_user              integer,
+    revision_count           integer           DEFAULT 0 NOT NULL,
+    CONSTRAINT deployment_pk PRIMARY KEY (deployment_id)
+  );
+
+  COMMENT ON COLUMN deployment.deployment_id            IS 'System generated surrogate primary key identifier.';
+  COMMENT ON COLUMN deployment.critter_id              IS 'The id of the critter. Internal Sims id.';
+  COMMENT ON COLUMN deployment.bctw_deployment_id       IS 'The external system id of a BCTW deployment / collar->animal assignment.';
+  COMMENT ON COLUMN deployment.create_date              IS 'The datetime the record was created.';
+  COMMENT ON COLUMN deployment.create_user              IS 'The id of the user who created the record as identified in the system user table.';
+  COMMENT ON COLUMN deployment.update_date              IS 'The datetime the record was updated.';
+  COMMENT ON COLUMN deployment.update_user              IS 'The id of the user who updated the record as identified in the system user table.';
+  COMMENT ON COLUMN deployment.revision_count          IS 'Revision count used for concurrency control.';
+  COMMENT ON TABLE  deployment                          IS 'Information about individual critters associated to specific bctw deployments.';
+
+  -- Add foreign key constraint from child table to parent table on critter_id
+  ALTER TABLE deployment ADD CONSTRAINT deployment_fk1
+    FOREIGN KEY (critter_id)
+    REFERENCES critter(critter_id);
+
+  -- Add unique constraint
+  CREATE UNIQUE INDEX deployment_uk1 ON deployment(critter_id, bctw_deployment_id);
+
+  -- Create audit and journal triggers
+  create trigger audit_critter before insert or update or delete on deployment for each row execute procedure tr_audit_trigger();
+  create trigger journal_critter after insert or update or delete on deployment for each row execute procedure tr_journal_trigger();
 
 
   ----------------------------------------------------------------------------------------
@@ -136,6 +174,7 @@ export async function up(knex: Knex): Promise<void> {
 
   create or replace view observation as select * from biohub.observation;
   create or replace view critter as select * from biohub.critter;
+  create or replace view deployment as select * from biohub.deployment;
   `);
 }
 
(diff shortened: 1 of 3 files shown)


database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
```diff
@@ -0,0 +1,184 @@
...
+  ----------------------------------------------------------------------------------------
+  -- Create Observation Table
+  ----------------------------------------------------------------------------------------
+
+  set search_path=biohub;
+
+  CREATE TABLE observation(
+    observation_id           integer           GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+    survey_id                integer           NOT NULL,
+    total_count   	         integer			     NOT NULL CHECK(total_count > 0),
```
NickPhura, line 23: Nice check 👍

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
```diff
@@ -0,0 +1,184 @@
...
+  COMMENT ON COLUMN critter.revision_count           IS 'Revision count used for concurrency control.';
+  COMMENT ON TABLE  critter                          IS 'Information about individual animals associated to specific observations.';
+
+  -- Add foreign key constraint from child table to parent table on observation_id
+  ALTER TABLE critter ADD CONSTRAINT critter_fk1
+    FOREIGN KEY (observation_id)
+    REFERENCES observation(observation_id);
+
+  -- Add unique constraint
+  CREATE UNIQUE INDEX critter_uk1 ON critter(critterbase_critter_id);
```
NickPhura: I think we don't want this index to be unique. Over time, multiple surveys/observations could observe the same critter.

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
```diff
@@ -0,0 +1,184 @@
...
+  COMMENT ON COLUMN critter.create_date              IS 'The datetime the record was created.';
+  COMMENT ON COLUMN critter.create_user              IS 'The id of the user who created the record as identified in the system user table.';
+  COMMENT ON COLUMN critter.update_date              IS 'The datetime the record was updated.';
+  COMMENT ON COLUMN critter.update_user              IS 'The id of the user who updated the record as identified in the system user table.';
+  COMMENT ON COLUMN critter.revision_count           IS 'Revision count used for concurrency control.';
+  COMMENT ON TABLE  critter                          IS 'Information about individual animals associated to specific observations.';
+
+  -- Add foreign key constraint from child table to parent table on observation_id
+  ALTER TABLE critter ADD CONSTRAINT critter_fk1
+    FOREIGN KEY (observation_id)
```
NickPhura, line 77: We should add an index for `observation_id`. 

Another funny thing about postgres, is that foreign keys dont come with indexes automatically like they do in other sql languages, so every time we add a foreign key, we also need to add an index.

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
```diff
@@ -0,0 +1,184 @@
...
+  COMMENT ON COLUMN deployment.create_date              IS 'The datetime the record was created.';
+  COMMENT ON COLUMN deployment.create_user              IS 'The id of the user who created the record as identified in the system user table.';
+  COMMENT ON COLUMN deployment.update_date              IS 'The datetime the record was updated.';
+  COMMENT ON COLUMN deployment.update_user              IS 'The id of the user who updated the record as identified in the system user table.';
+  COMMENT ON COLUMN deployment.revision_count          IS 'Revision count used for concurrency control.';
+  COMMENT ON TABLE  deployment                          IS 'Information about individual critters associated to specific bctw deployments.';
+
+  -- Add foreign key constraint from child table to parent table on critter_id
+  ALTER TABLE deployment ADD CONSTRAINT deployment_fk1
+    FOREIGN KEY (critter_id)
```
NickPhura, line 118: For the same foreign key reasons above, lets add an index for `critter_id`

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
```diff
@@ -0,0 +1,184 @@
...
+    -- Person           Date        Comments
+    -- ---------------- ----------- --------------------------------------
+    -- mac.deluca@quartech.com
+    --                  2023-17-08  initial release
+    -- *******************************************************************
+    DECLARE
+      num_critters integer := (SELECT count(*) FROM critter WHERE critter.observation_id = NEW.observation_id);
+      observation_total_count integer := (SELECT total_count FROM observation WHERE observation.observation_id = NEW.observation_id);
+    BEGIN
+      IF (num_critters >= observation_total_count) THEN
```
NickPhura: Should this be greater than (rather than greater or equal)?

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
NickPhura: Random thought: It does remind me of the existence of non-observation observations (ie: I specifically am recording the absence of an animal). But I think that is its own separate issue to tackle elsewhere/later, and isn't the same as a count of 0 anyways.

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
NickPhura: I think we do want a unique index on `observation_id + critterbase_critter_id` though. Since it wouldn't make sense for the same observation to observe the same animal more than once.

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  -- Create audit and journal triggers
+  create trigger audit_critter before insert or update or delete on critter for each row execute procedure tr_audit_trigger();
+  create trigger journal_critter after insert or update or delete on critter for each row execute procedure tr_journal_trigger();
+
+  ----------------------------------------------------------------------------------------
+  -- Create Deployment table
+  ----------------------------------------------------------------------------------------
+
+  CREATE TABLE deployment(
```
NickPhura, line 94: Does this table need to track devices over time?

Not clear in my head if that is something sims should track, or if that is BCTW's responsibility which it already does.
Probably also relates back to that talk around whether or not its important to track which survey added a collar vs removed it, etc. So maybe for now we just don't worry about it until that is decided on.

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
MacDeluca: I have the trigger firing on before insert. So the count of critters would be the table before the new record is added.

MacDeluca: Do we want the trigger to fire on update aswell or just insert?

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
MacDeluca: Yea im not too sure. I think currently it needs to track the critter and the deployment. We will let BCTW manage all the collar information and deployment dates etc.

Still looking into the solution for tracking the survey for deployments and captures.

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
MacDeluca: added

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
MacDeluca: added

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
MacDeluca: added

database/src/migrations/20230817000000_critter_observation_deployment_tables.ts
MacDeluca: Updated this to not fire on update. (It was throwing an error on update if the total count === critter count)

NickPhura (approved):
👍

al-rosenthal (approved):
☄️

PR #1074, opened 2023-08-23
SIMSBIOHUB-159: Move project_type to survey_type

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-159

#### Migrations
- New migration to add survey_type table, migrate existing project_type values, drop project_type table
#### API/APP
- Update API to remove all project_type code, and add corresponding survey_type code
- Update APP to remove all project_type code, and add corresponding survey_type code

#### Migrations
- Added a few missed indexes/unique constraints on some recent migrations.

#### Seed
- Enhanced the project/survey seed to use more randomly selected lookup values and generated field values.

## Testing Notes

Project type should no longer appear anywhere at the project level.
- Existing project types should be migrated to their child surveys
  - I did test against 2 projects, and multiple surveys, to see that the types were migrated correctly. But a second pass at this by is probably not a bad idea.

Surveys should now have a new survey type field in the UI.
- Creating/Viewing/Editing surveys should work


al-rosenthal (dismissed):
ran this locally, looks good 🌤️

al-rosenthal (dismissed):
☀️

api/src/repositories/survey-repository.ts
```diff
@@ -74,6 +74,47 @@ export interface ISurveyProprietorModel {
   disa_required: boolean;
 }
 
+const SurveyRecord = z.object({
```
curtisupshall, line 77: +1 for Zod types

api/src/services/survey-service.ts
```diff
@@ -142,7 +142,14 @@ export class SurveyService extends DBService {
    * @memberof SurveyService
    */
   async getSurveyData(surveyId: number): Promise<GetSurveyData> {
-    return this.surveyRepository.getSurveyData(surveyId);
+    const [surveyData, surveyTypesData] = await Promise.all([
```
curtisupshall, line 145: Is there any reason we prefer `Promise.all` over something like:
```typescript
const surveyData = await this.surveyRepository.getSurveyData(surveyId);
const surveyTypesData = await this.surveyRepository.getSurveyTypesData(surveyId);
```
I'm guessing it's because using `Promise.all` allows both these promises to be fullfilled concurrently?

api/src/repositories/survey-repository.ts
```diff
@@ -625,6 +686,37 @@ export class SurveyRepository extends BaseRepository {
...
 
+  /**
+   * Inserts a new survey_type record.
+   *
+   * @param {number} focal_species_id
+   * @param {number} surveyId
+   * @return {*}  {Promise<void>}
+   * @memberof SurveyRepository
+   */
+  async insertSurveyType(typeId: number, surveyId: number): Promise<void> {
```
curtisupshall: Thoughts on replacing this with something like:
```typescript
async insertSurveyTypes(typeIds: number[], surveyId: number): Promise<{ survey_type_id: number }[]>
```
Motivation being if we want to attach five types to a survey, we could avoid dispatching five requests to the SQL server.

curtisupshall: Example:

```typescript
async insertStakeholderPartnerships(
    stakeholderPartners: string[],
    surveyId: number
  ): Promise<{ survey_stakeholder_partnership_id: number }[]> {
    defaultLog.debug({ label: 'insertStakeholderPartnerships', stakeholderPartners, surveyId });
    const queryBuilder = getKnex()
      .table('survey_stakeholder_partnership')
      .insert(
        stakeholderPartners.map((stakeholder: string) => ({
          name: stakeholder,
          survey_id: surveyId
        }))
      )
      .returning('survey_stakeholder_partnership_id');

    const response = await this.connection.knex<{ survey_stakeholder_partnership_id: number }>(queryBuilder);

    if (!response || response.rowCount === 0) {
      throw new ApiExecuteSQLError('Failed to insert survey stakeholder partnerships', [
        'ErrorRepository->insertStakeholderPartnerships',
        'rowCount was null or undefined, expected rowCount != 0'
      ]);
    }

    return response.rows;
  }
```

database/src/migrations/20230822115400_move_project_type_to_survey.ts
curtisupshall, line 1: May need to add `CREATE OR REPLACE PROCEDURE api_delete_survey...` to your migration so that survey_types are deleted when a survey is deleted.

api/src/repositories/survey-repository.ts
NickPhura: Ya fair point, I was just being fast and lazy. Updated.

database/src/migrations/20230822115400_move_project_type_to_survey.ts
NickPhura: Good catch. Updated.

api/src/services/survey-service.ts
NickPhura: Yeah just being able to run both at the same time, thats all.

api/src/repositories/survey-repository.ts
NickPhura: 🕺

al-rosenthal (approved):
🏂

curtisupshall (approved):
Code looks good!

PR #1076, opened 2023-08-24
SIMSBIOHUB-158: Move Partnerships from Projects to Surveys

curtisupshall:
## Links to Jira Tickets

 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-158

## Description of Changes
 - Removes old database tables and adds two new tables:
   - `survey_first_nations_partnership`
   - `survey_stakeholder_partnership`
 - Migrates all project partnerships to each project's surveys. (Partnerships for project that don't have surveys will be lost.)
 - Moves UI components to survey create, edit and details page
 - Creates new API service and repo methods


database/src/migrations/20230823000000_survey_partnerships.ts
```diff
@@ -0,0 +1,231 @@
...
+  DROP VIEW biohub_dapi_v1.project_first_nation;
+
+  SET search_path=biohub;
+
+  ----------------------------------------------------------------------------------------
+  -- Remove old 'stakeholder_partnership' and 'survey_first_nation_partnership' tables
+  ----------------------------------------------------------------------------------------
+
+  DROP TABLE stakeholder_partnership;
+  DROP TABLE project_first_nation;
```
NickPhura: We need to migrate any existing prod data from the old tables to the new tables before we can drop the old tables.

database/src/migrations/20230823000000_survey_partnerships.ts
```diff
@@ -0,0 +1,231 @@
...
+
+ 
+  ----------------------------------------------------------------------------------------
+  -- Update api_delete_survey procedure
+  ----------------------------------------------------------------------------------------
+
+
+  set search_path=biohub;
+
+  CREATE OR REPLACE PROCEDURE api_delete_survey(p_survey_id integer)
```
NickPhura, line 208: We probably also need to update the project delete function,as it will still be referencing the old partnerships tables

database/src/migrations/20230823000000_survey_partnerships.ts
curtisupshall: Totally forgot - done.

database/src/migrations/20230823000000_survey_partnerships.ts
curtisupshall: Good catch! Done.

api/src/paths/project/{projectId}/survey/{surveyId}/update.ts
```diff
@@ -167,6 +177,25 @@ PUT.apiDoc = {
                 }
               }
             },
+            partnerships: {
+              title: 'Survey partnerships',
+              type: 'object',
```
NickPhura, line 190: Are the internal properties required (indigenous_partnerships, stakeholder_partnerships)?
Could add a required section at this level as well, if they are.

NickPhura: Same for the view and create endpoints*

api/src/repositories/survey-repository.ts
```diff
@@ -1100,4 +1101,188 @@ export class SurveyRepository extends BaseRepository {
...
+      WHERE
+      sfnp.survey_id = ${surveyId}
+      GROUP BY
+        fn.first_nations_id,
+        fn.name;
+    `;
+
+    const response = await this.connection.query(sqlStatement.text, sqlStatement.values);
+
+    const result = response?.rows;
```
NickPhura: You should be able to drop the question mark as rows will always be defined and an array.

For the if statement below, rows will always be defined, so we should either remove that check, or make it a `response.rowCount` check instead.

NickPhura: Same for a few of the other new functions below*

api/src/repositories/survey-repository.ts
```diff
@@ -1100,4 +1101,188 @@ export class SurveyRepository extends BaseRepository {
...
+        firstNationsIds.map((firstNationsId: number) => ({
+          first_nations_id: firstNationsId,
+          survey_id: surveyId
+        }))
+      )
+      .returning('survey_first_nation_partnership_id');
+
+    const response = await this.connection.knex<{ survey_first_nation_partnership_id: number }>(queryBuilder);
+
+    if (!response || response.rowCount === 0) {
```
NickPhura: response will always be defined, so could drop the initial `!response` check

NickPhura: Lots of our old code makes that check because early on I had the types wrong (was under the impression that response could be undefined, but turns out it cant be. node-pg just throws if it is unable to return a valid response object for w/e reason).

NickPhura:
A few minor comments, but looks good in general!

api/src/repositories/survey-repository.ts
```diff
@@ -1100,4 +1101,188 @@ export class SurveyRepository extends BaseRepository {
...
+      ON
+      sfnp.first_nations_id = fn.first_nations_id
+      WHERE
+      sfnp.survey_id = ${surveyId}
+      GROUP BY
+        fn.first_nations_id,
+        fn.name;
+    `;
+
+    const response = await this.connection.query(sqlStatement.text, sqlStatement.values);
```
NickPhura: A couple of notes: Lets prefer the `.sql` query (its the same as `.query` but you can pass the whole `sqlStatement` to it rather than the text and values parts (trying to slowly deprecate `.query`).

We should pass a zod schema as the second parameter in all of these calls, even if its defined in-line. Still a good step in the right direction for response validation. The generic type can be removed then as well (in spots below that defined one), as the zod schema will ensure the response is typed.

curtisupshall: Yeah I caught the `.query` discrepancy just now. Will work to get some Zod schema incorporated

api/src/repositories/survey-repository.ts
curtisupshall: Will remove these checks!

api/src/repositories/survey-repository.ts
curtisupshall: Will do.

api/src/paths/project/{projectId}/survey/{surveyId}/update.ts
curtisupshall: Updated these files to xplicitly declared which fields were required. 👍

NickPhura:
@curtisupshall Getting an error when I try to edit a project:
image

al-rosenthal (commented):
It migrates partnerships to surveys properly but I am unable to delete a project/ survey. 

From the log when trying to delete a project:
image

Edit: pushed a fix up for this

NickPhura (approved):

al-rosenthal (approved):
⚡

PR #1079, opened 2023-08-29
TechDebt: useAxios 401/403 Request Retry

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Add retry mechanic to useAxios to handle the case where your keycloak token has expired while working in the APP.

This can occur if you sit on an App page for longer than 5 minutes, without interacting with the api. In this case, the token will have quietly expired (5 minutes in Standard realm), and your next http request is guaranteed to fail.

Updated useAxios so that if the response fails, and the status is specifically a 401 or 403, then it attempts to refresh the token, and if it is successfully refreshed, re-run the last failed request.

Add useAxios tests.

## Testing Notes

Let your token expire (wait 5 minutes), and then make some request (ie: save project, or similar), and you should see 2 requests get made, the first of which fails, and the second of which succeeds. The second call should also have a new/different bearer token in the headers.


curtisupshall (approved):

KjartanE (approved):
Worked on local 🐈

PR #1083, opened 2023-08-31
SIMSBIOHUB-192: Add Site Selection Strategy + stratums to survey form

curtisupshall:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-192
- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-247
- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-248

## Description of Changes

 - Added survey site sampling strategies and stratums to the Create Survey and Edit Survey pages
 - Added service and repo methods for creating, updating and deleting survey site selection strategies and stratums
 - Created survey site selection strategy API schema and model properties

## Testing Notes

- TBD.

PR #1085, opened 2023-09-05
SISMBIOHUB-259: Survey Block 

al-rosenthal:
## Links to Jira Tickets

[SIMSBIOHUB-259](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-259)
[SIMSBIOHUB-260](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-260)
[SIMSBIOHUB-261](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-261)

## Description of Changes

- New service and repository added to handle survey block actions
- Added UI for survey block create/ edit survey pages

## Testing Notes


api/src/repositories/survey-block-repository.ts
```diff
@@ -0,0 +1,147 @@
+import SQL from 'sql-template-strings';
+import { z } from 'zod';
+import { ApiExecuteSQLError } from '../errors/api-error';
+import { BaseRepository } from './base-repository';
+
+export const SurveyBlock = z.object({
+  survey_block_id: z.number().nullable(),
```
curtisupshall: Should we remove this property? Is there ever a case where you expect `SurveyBlock.survey_block_id` to be defined, but couldn't otherwise just use `SurveyBlockRecord` instead?

api/src/repositories/survey-block-repository.ts
```diff
@@ -0,0 +1,147 @@
...
+  survey_id: z.number(),
+  name: z.string(),
+  description: z.string()
+});
+
+export type SurveyBlock = z.infer<typeof SurveyBlock>;
+
+// This describes the a row in the database for Survey Block
+export const SurveyBlockRecord = z.object({
```
curtisupshall, line 14: Maybe could use `z.extend` here?

api/src/services/survey-block-service.ts
```diff
@@ -0,0 +1,89 @@
...
+   * @param {number} surveyId
+   * @param {SurveyBlock[]} blocks
+   * @return {*} {Promise<void>}
+   * @memberof SurveyBlockService
+   */
+  async updateInsertSurveyBlocks(surveyId: number, blocks: SurveyBlock[]): Promise<void> {
+    const insertUpdate: Promise<any>[] = [];
+
+    blocks.forEach((item: SurveyBlock) => {
+      item.survey_id = surveyId;
```
curtisupshall: Is this assignment needed?

api/src/services/survey-block-service.ts
```diff
@@ -0,0 +1,89 @@
...
+   * Inserts, Updates and Deletes Block records
+   * All passed in blocks are treated as the source of truth,
+   * Any pre existing blocks that do not collide with passed in blocks are deleted
+   *
+   * @param {number} surveyId
+   * @param {SurveyBlock[]} blocks
+   * @return {*} {Promise<void>}
+   * @memberof SurveyBlockService
+   */
+  async upsertSurveyBlocks(surveyId: number, blocks: SurveyBlock[]): Promise<void> {
```
curtisupshall: I find it a little confusing that there are two methods with very siimilar names. Because they take the same parameters, this would probably cause confusion in the future. My instinct would be to delete the `updateInsertSurveyBlocks` method and move its logic into this one. But it's your call!

curtisupshall: Overall I think your approach to the problem is sound 👍

api/src/services/survey-service.ts
```diff
@@ -445,11 +453,20 @@ export class SurveyService extends DBService {
...
+    if (postSurveyData.blocks) {
+      promises.push(this.upsertBlocks(surveyId, postSurveyData.blocks));
+    }
+
     await Promise.all(promises);
 
     return surveyId;
   }
 
+  async upsertBlocks(surveyId: number, blocks: SurveyBlock[]): Promise<void> {
```
curtisupshall: Nit: jsdoc

app/src/features/surveys/components/BlockForm.tsx
```diff
@@ -0,0 +1,38 @@
...
+import CustomTextField from 'components/fields/CustomTextField';
+import React from 'react';
+
+export interface IBlockData {
+  survey_block_id: number | null;
+  name: string;
+  description: string;
+}
+
+const BlockForm: React.FC = () => {
```
curtisupshall, line 12: Nit: Remove `React.FC`:

```javascript
const BlockForm = () => {
```

app/src/features/surveys/components/BlockForm.tsx
```diff
@@ -0,0 +1,38 @@
...
+              label="Name"
+              maxLength={50}
+              other={{ placeholder: 'Maximum 50 characters', required: true }}
+            />
+          </Box>
+          <CustomTextField
+            name="description"
+            label="Description"
+            maxLength={250}
+            other={{ multiline: true, placeholder: 'Maximum 250 characters', required: true, rows: 3 }}
```
curtisupshall, line 30: Ideally this should be enforced using Yup schema. See: https://github.com/bcgov/biohubbc/pull/1083/files#diff-28a6bb0e6fc7b03b7de1b8d12c0f779a870fc541953a0352dd61f59ea07bf772

app/src/features/surveys/components/CreateSurveyBlockDialog.tsx
```diff
@@ -0,0 +1,74 @@
...
+        }}
+        dialogSaveButtonLabel="Add Block"
+        onCancel={() => onClose()}
+        onSave={(formValues) => {
+          setBlockName(formValues.name);
+          setIsSnackBarOpen(true);
+          onSave(formValues);
+        }}
+      />
+      {/* This is done instead of the dialogContext because that causes the form to rerender, removing any changes the user makes */}
```
curtisupshall: Not entirely sure what this comment is suggesting. Is the Snackbar only included because we need to force a rerender of the form?

curtisupshall: Unless it's a requirement in the ticket, I think adding a snackbar is a bit out of place, since it's not used for any other controls on the Survey form (unless I'm mistaken)

app/src/features/surveys/components/EditSurveyBlockDialog.tsx
curtisupshall, line 1: Could probably get away with having a single Dialog component for both editing and creating Survey Blocks; It's how we (used to) handle Funding Sources, and it's how we handle [Survey Stratums](https://github.com/bcgov/biohubbc/pull/1083/files#diff-28a6bb0e6fc7b03b7de1b8d12c0f779a870fc541953a0352dd61f59ea07bf772)

app/src/features/surveys/components/SurveyBlockSection.tsx
```diff
@@ -0,0 +1,154 @@
...
+      const data = values.blocks;
+      data.splice(editData.index, 1);
+      setFieldValue(name, data);
+    }
+    setAnchorEl(null);
+  };
+
+  return (
+    <>
+      {/* CREATE BLOCK DIALOG */}
```
curtisupshall, line 63: Nit: Remove comments

app/src/features/surveys/CreateSurveyPage.tsx
```diff
@@ -377,6 +379,13 @@ const CreateSurveyPage = () => {
 
                 <Divider className={classes.sectionDivider} />
 
+                <HorizontalSplitFormComponent
+                  title="Block Data"
+                  summary="Data for Blocks"
+                  component={<SurveyBlockSection name="blocks" />}
```
curtisupshall: Is the `name` prop necessary? It's the same value for both instances of `<SurveyBlockSection />`

app/src/features/surveys/CreateSurveyPage.tsx
```diff
@@ -377,6 +379,13 @@ const CreateSurveyPage = () => {
 
                 <Divider className={classes.sectionDivider} />
 
+                <HorizontalSplitFormComponent
```
curtisupshall, line 382: Some UI elements are missing:
image
 vs.

image

curtisupshall (commented):
A couple notes from manually testing:

1. The `<Card>` element used to display each survey block should use the `variant='outlined'` prop.
2. Clicking the `<Card>` shouldn't open the context menu - only the button within the card should open it.
3. The positioning of the context menu is a bit off; Use these props on the `<Menu>`:
```javascript
anchorOrigin={{
  vertical: 'top',
  horizontal: 'right'
}}
transformOrigin={{
  vertical: 'top',
  horizontal: 'right'
}}
```

api/src/services/survey-block-service.ts
al-rosenthal: yup. If the block being added is being added to a brand new survey we won't have the ID until the api request is being made.

api/src/services/survey-block-service.ts
al-rosenthal: I agree it is confusing.

api/src/repositories/survey-block-repository.ts
al-rosenthal: `survey_block_id` is used to check if an incoming block needs to be updated or created net new.

app/src/features/surveys/components/CreateSurveyBlockDialog.tsx
al-rosenthal: It is a requirement in the ticket. I might have to write a bit more but I was trying to explain why I didn't use the dialog context to render the snack bar.

app/src/features/surveys/CreateSurveyPage.tsx
al-rosenthal: No not really. I think just a left over from an earlier iteration

api/src/services/survey-block-service.ts
curtisupshall: Ahh understood

api/src/repositories/survey-block-repository.ts
curtisupshall: But if a block has a `survey_block_id`, then we can also assert it has all of the properties that a `SurveyBlockRecord` has, no?

app/src/features/surveys/components/BlockForm.tsx
al-rosenthal: I'll update it for both. I did have a conversation with Jer about a similar situation for funding sources. It's a strange interaction if the database throws an error about max length but you still let the user input that much. Instead of validating it just to restrict the input. I'll definitely add that to the yup schema

app/src/features/surveys/CreateSurveyPage.tsx
al-rosenthal: good catch, added

api/src/services/survey-block-service.ts
al-rosenthal: combined the functions

api/src/services/survey-service.ts
al-rosenthal: done

app/src/features/surveys/components/SurveyBlockSection.tsx
al-rosenthal: the components are properly named but I think the comments give good space between everything else

KjartanE (dismissed):
🐈 ran locally, worked as intended

api/src/repositories/survey-block-repository.ts
al-rosenthal: I left the zod schema alone and decided to just rename the SurveyBlock to PostSurveyBlock to indicate it is for inserting data. I also turned it into an interface because there was no reason for it to be a zod schema.

api/src/repositories/survey-block-repository.ts
al-rosenthal: Comment is old now but I think it's better in this case to leave these as 2 discrete type definitions instead of extending one of them, keeps it a little easier to read in my opinion.

KjartanE (approved):
🌵

curtisupshall (approved):
Tested in Openshift; Code looks good and works well!

PR #1095, opened 2023-09-15
SIMSBIOHUB-263: Upload Keyx attachments to BCTW

JeremyQuartech:
## Links to Jira Tickets

- [SIMSBIOHUB-263](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-263)

## Description of Changes
- This PR adds all the supporting changes required to identify keyx files during a survey attachment, and send those files to BCTW.
- Adds new /attachments/keyx/upload endpoint for keyx or zip files that handles BCTW transfers.
- Adds 'uploadKeyx' method to BctwService.
- Refined BctwService axios error-handling to support response time-out and connection refusal.
- Adds 'Upload KeyX' option to survey uploads.
- Customized file uploader for keyx files to display extended errors.

## Testing Notes

- It is working reliably at this stage, ~however the error handling could use refinement.~

- ~BCTW API requires a patch for this to function (will remove comment once this is live).~


api/src/paths/project/{projectId}/survey/{surveyId}/attachments/process.ts
```diff
@@ -0,0 +1,118 @@
...
+      },
+      required: true
+    }
+  ],
+  responses: {
+    200: {
+      description: 'Survey process attachments response.',
+      content: {
+        'application/json': {
+          schema: {}
```
GrahamS-Quartech: Should probably add some kind of schema here, even if just a stub

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/process.ts
```diff
@@ -0,0 +1,118 @@
...
+import { PROJECT_PERMISSION, SYSTEM_ROLE } from '../../../../../../constants/roles';
+import { getDBConnection } from '../../../../../../database/db';
+import { authorizeRequestHandler } from '../../../../../../request-handlers/security/authorization';
+import { AttachmentKeyxService } from '../../../../../../services/attachment-keyx-service';
+import { IBctwUser } from '../../../../../../services/bctw-service';
+import { getLogger } from '../../../../../../utils/logger';
+
+// TODO: It may make more sense to move this to the /telemetry endpoint.
+
+const defaultLog = getLogger('/api/project/{projectId}/survey/{surveyId}/attachments/process');
```
GrahamS-Quartech: I wonder if the name for this endpoint should be more specific if it only deals with keyx files? The similarly generic sounding upload endpoint takes all files, but this one won't do anything unless the files all have keyx as the extension.

curtisupshall: I think the name is fitting; If in the future we ever wanted to process other types of files, it should all be kicked off by a single endpoint.

GrahamS-Quartech: Ok valid

api/src/services/attachment-keyx-service.ts
curtisupshall, line 1: I feel like these changes are minimal enough to just put them into the Attachment service. This service is a little too niche I think

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/upload.ts
```diff
@@ -140,6 +142,28 @@ export function uploadMedia(): RequestHandler {
         ATTACHMENT_TYPE.OTHER
       );
 
+      // Check for keyx files in the raw media
+      const keyxArray = [];
+      const ZIP = ['application/zip', 'application/x-zip-compressed', 'application/zip-compressed'];
```
curtisupshall: I would rename this to `zipMimeTypes`

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/processKeyx.ts
```diff
@@ -0,0 +1,118 @@
...
+    const user: IBctwUser = {
+      keycloak_guid: req['system_user']?.user_guid,
+      username: req['system_user']?.user_identifier
+    };
+
+    try {
+      await connection.open();
+
+      const attachmentKeyxService = new AttachmentKeyxService(connection);
+      await attachmentKeyxService.processKeyxRecords(user);
```
curtisupshall: We don't need to pass the user into this method, because calling it should process all KeyX records for all users (unless I'm missing some assumptions). Since you're passing the `connection` object to the constructor, the `user` object can just be obtained inside this method if needed.

api/src/services/attachment-keyx-service.ts
```diff
@@ -0,0 +1,47 @@
...
+   * @return {*}
+   * @memberof AttachmentKeyxService
+   */
+  async insertKeyxReference(attachmentId: number) {
+    return this.attachmentRepository.insertSurveyAttachmentKeyx(attachmentId);
+  }
+
+  /**
+   * Processes keyx records.
+   * TODO: This can likely be made much more efficient.
```
curtisupshall: I think comments about performance can be left out. If a method can't be made efficient on first pass, and a better implemention isn't discerned during code review, then we can probably address optimizations later when we can prove that they are justified

curtisupshall: In my opinion, this method is very reasonably written 👍

api/src/services/attachment-keyx-service.ts
```diff
@@ -0,0 +1,47 @@
...
+   * @return {*}  {Promise<void>}
+   * @memberof AttachmentKeyxService
+   */
+  async processKeyxRecords(user: IBctwUser): Promise<void> {
+    const bctwService = new BctwService(user);
+
+    // get list of unprocessed records
+    const records = await this.attachmentRepository.getSurveyKeyxAttachmentsToProcess();
+    // get files from s3
+    for (const record of records) {
```
curtisupshall: I think if you map each record to a Promise, you can let all of them run concurrently:

```typescript
const records = await this.attachmentRepository.getSurveyKeyxAttachmentsToProcess();
const promises = records.map((record) => new Promise(async (resolve, reject) => {
  try {
    const file = await getFileFromS3(record.key);
    await bctwService.uploadKeyX(file, record.key);
    await this.attachmentRepository.updateSurveyAttachmentKeyx(record.survey_attachment_id);
    resolve();
  } catch (error) {
    reject(error);
  }
));

return Promise.all(promises);
```

Not certain if this is correct though... @NickPhura?

app/src/features/surveys/view/SurveyAttachments.tsx
```diff
@@ -64,6 +64,7 @@ const SurveyAttachments: React.FC = () => {
         onClose={() => {
           setOpenUploadAttachments(false);
           surveyContext.artifactDataLoader.refresh(projectId, surveyId);
+          biohubApi.survey.processSurveyKeyxAttachments(projectId, surveyId); // TODO: Find a more reliable place to call this.
```
curtisupshall: Call this as soon as the upload succeeds (see Line 43):

```typescript
const getUploadHandler = (): IUploadHandler => {
    return (file, cancelToken, handleFileUploadProgress) => {
      return biohubApi.survey.uploadSurveyAttachments(projectId, surveyId, file, cancelToken, handleFileUploadProgress)
        .then((response: IUploadAttachmentResponse) => {
           biohubApi.survey.processSurveyKeyxAttachments(projectId, surveyId);
        })
    };
  };
```

app/src/hooks/api/useSurveyApi.ts
```diff
@@ -217,6 +217,18 @@ const useSurveyApi = (axios: AxiosInstance) => {
...
 
+  /**
+   * Trigger processing of survey attachments
+   *
+   * @param {number} projectId
+   * @param {number} surveyId
+   * @return {*}
+   */
+  const processSurveyKeyxAttachments = async (projectId: number, surveyId: number) => {
+    const { data } = await axios.post(`/api/project/${projectId}/survey/${surveyId}/attachments/processkeyx`);
```
curtisupshall: Opinion: rename this back to `/process`

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/processKeyx.ts
JeremyQuartech: Yeah you're right that it doesn't really need to be passed in, however the bctw API requires a user to be referenced for auditing purposes. Is there a straight-forward way to get the keycloak_guid and username from the connection object?

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/keyx/upload.ts
```diff
@@ -0,0 +1,184 @@
...
+        throw new HTTP400('The file must either be a keyx file or a zip containing only keyx files.');
+      }
+
+      const user: IBctwUser = {
+        keycloak_guid: req['system_user']?.user_guid,
+        username: req['system_user']?.user_identifier
+      };
+      const bctwService = new BctwService(user);
+      const bctwUploadResult = await bctwService.uploadKeyX(rawMediaFile);
+      console.log('bctwUploadResult', bctwUploadResult);
```
GrahamS-Quartech: Remove or use logger

api/src/services/bctw-service.ts
```diff
@@ -255,6 +288,42 @@ export class BctwService {
...
+      headers: {
+        ...formData.getHeaders()
+      }
+    };
+    const response = await this.axiosInstance.post(UPLOAD_KEYX_ENDPOINT, formData, config);
+    const data: IUploadKeyxResponse = response.data;
+    if (data.errors.length) {
+      const actualErrors: string[] = [];
+      for (const error of data.errors) {
+        // Ignore errors that indicate that a keyX already exists
```
GrahamS-Quartech, line 317: Just curious why we ignore this?

app/src/components/file-upload/FileUploadItem.tsx
```diff
@@ -203,13 +208,42 @@ const FileUploadItem: React.FC<IFileUploadItemProps> = (props) => {
         <Box display="flex" flexDirection="row" alignItems="center" p={2} width="100%">
           <Icon path={mdiFileOutline} size={1.5} className={error ? classes.errorColor : classes.fileIconColor} />
           <Box pl={1.5} flex="1 1 auto">
-            <Box display="flex" flexDirection="row" flex="1 1 auto" alignItems="center" height="3rem">
+            <Box display="flex" flexDirection="row" flex="1 1 auto" alignItems="center" minHeight="3rem">
```
GrahamS-Quartech: This seems like a case for sx prop (@curtisupshall ?)

api/src/services/bctw-service.ts
JeremyQuartech: The reason for this is because it won't effect the end result. Basically, if the user adds a keyx file, but it already exists in BCTW, there is no harm done because the BCTW is able to retrieve telemetry.

app/src/components/file-upload/FileUploadItem.tsx
curtisupshall: Yeah I believe all these props can be moved into `sx`. Also, the min height should be absolute, so just use the `sx` units instead of rems.

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/keyx/upload.ts
```diff
@@ -0,0 +1,205 @@
...
+        'application/json': {
+          schema: {
+            type: 'object',
+            required: ['attachmentId', 'revision_count', 'keyxResults'],
+            properties: {
+              keyxResults: {
+                type: 'object',
+                required: ['totalKeyxFiles', 'newRecords', 'existingRecords'],
+                properties: {
+                  totalKeyxFiles: { type: 'number' },
```
curtisupshall: Nit: use `type: 'integer'` instead

GrahamS-Quartech (dismissed):
Nice work 👍

api/src/utils/media/media-utils.ts
```diff
@@ -110,3 +110,28 @@ export const isZipMimetype = (mimetype: string): boolean => {
...
+  const mimeType = mime.getType(file.originalname) ?? '';
+  if (file?.originalname.endsWith('.keyx')) {
+    return true;
+  } else if (isZipMimetype(mimeType)) {
+    const zipEntries = parseUnknownZipFile(file.buffer);
+    if (zipEntries.length === 0) {
+      return false;
+    }
+    for (const zipEntry of zipEntries) {
+      if (!zipEntry.fileName.endsWith('.keyx')) {
```
curtisupshall: If my ZIP contains the following files:
```
[
  'cat.jpg',
  'myfile.keyx',
  'test.csv'
]
```
then this loop would return `false` upon discovering `cat.jpg`, even though the documentation says it should "return `true` if the file is [...] a zip that contains a keyx file."

api/src/utils/media/media-utils.ts
```diff
@@ -110,3 +110,28 @@ export const isZipMimetype = (mimetype: string): boolean => {
...
+        return false;
+      }
+    }
+  }
+  return true;
```
curtisupshall: If the file is not a `.keyx` and not a `.zip`, the default return should be `false`, right?

api/src/utils/media/media-utils.ts
curtisupshall: Unless there's something I'm missing, I would rewrite this:
```typescript
export const isKeyXFileOrArchive = (file: Express.Multer.File): boolean => {
  // File is a KeyX file if it ends in '.keyx'
  if (file?.originalname.endsWith('.keyx')) {
    return true;
  }

  const mimeType = mime.getType(file.originalname) ?? '';
  if (!isZipMimetype(mimeType)) {
    // File cannot be a KeyX file, since it is not an archive nor does it have a .keyx extension
    return false;
  }

  const zipEntries = parseUnknownZipFile(file.buffer);

  // File is a KeyX archive if some file in the archive has a .keyx extension
  return zipEntries.some((zipEntry) => zipEntry.fileName.endsWith('.keyx');
}
```

JeremyQuartech: Yeah good catch, the requirements changed and I forgot to update the description. I cleaned up some of the formatting too.

api/src/utils/media/media-utils.ts
```diff
@@ -110,3 +110,31 @@ export const isZipMimetype = (mimetype: string): boolean => {
...
 };
+
+/**
+ * Returns true if the file is a keyx file, or a zip that contains only keyx files.
+ *
+ * @export
+ * @param {Express.Multer.File} file
+ * @return {*}  {boolean}
+ */
+export function checkFileForKeyx(file: Express.Multer.File): boolean {
```
curtisupshall, line 121: This looks great! I might rename the method to `ensureKeyxFileOrArchive`, but otherwise it looks good. +1 for documentation

app/src/features/surveys/view/SurveyAttachments.tsx
```diff
@@ -54,11 +61,24 @@ const SurveyAttachments: React.FC = () => {
...
+  const getDialogTitle = () => {
+    switch (attachmentType) {
+      case AttachmentType.REPORT:
+        return 'Upload Report';
+      case AttachmentType.KEYX:
+        return 'Upload KeyX';
+      case AttachmentType.OTHER:
+        return 'Upload Attachments';
+      default:
+        return '';
```
curtisupshall, line 73: I'd refactor this to use a non-empty default string:
```javascript
case AttachmentType.OTHER:
default:
  return 'Upload Attachment';
```

app/src/components/file-upload/FileUploadItem.tsx
```diff
@@ -146,6 +151,11 @@ const FileUploadItem: React.FC<IFileUploadItemProps> = (props) => {
     uploadHandler(file, cancelToken, handleFileUploadProgress)
       .then(handleFileUploadSuccess, (error: APIError) => {
         setError(error?.message);
+        setErrorDetails(
+          error?.errors?.map((e) => {
+            return { _id: v4(), message: e?.toString() };
```
curtisupshall, line 156: If this `_id` is being used for providing a React key, using a UUID is generally ill-advised. This is something we need to address in other parts of SIMS so I wouldn't worry too much about it now. The alternative is using the array index, but React also discourages this 🤔

app/src/features/surveys/view/SurveyAttachments.tsx
```diff
@@ -54,11 +61,24 @@ const SurveyAttachments: React.FC = () => {
     };
   };
 
+  const getDialogTitle = () => {
```
curtisupshall, line 64: If a callback is only called during render, I would memoize it as a variable:
```javascript
const dialogTitle = useMemo(() => {
  switch (attachmentType) { ... }
}, [attachmentType]);
```

app/src/components/file-upload/FileUploadItem.tsx
```diff
@@ -203,13 +213,48 @@ const FileUploadItem: React.FC<IFileUploadItemProps> = (props) => {
...
                 <Typography variant="body2" component="div">
                   <strong>{file.name}</strong>
                 </Typography>
                 <Typography variant="caption" component="div">
-                  {error || status}
+                  {error ? (
+                    <>
+                      {error}
+                      {props.enableErrorDetails && errorDetails && errorDetails.length ? (
+                        <IconButton size="small" onClick={() => setShowErrorDetails(!showErrorDetails)}>
```
curtisupshall, line 233: Should be able to avoid using SvgIcon by refactoring:
```javascript
<IconButton size="small" onClick={() => setShowErrorDetails(!showErrorDetails)}>
  <Icon path={showErrorDetails ? mdiChevronUp : mdiChevronDown} size={1} />
</IconButton>
```

app/src/components/file-upload/FileUploadItem.tsx
curtisupshall: If it is just being used as a key, I would rename it to `key` instead of `_id`

curtisupshall (approved):
Code looks good! Just had some minor comments, none are urgent

app/src/components/file-upload/FileUploadItem.tsx
GrahamS-Quartech: FWIW, Sonar will at least not complain about code smell by doing this. Kinda curious since depending on the context, the index would actually be more stable than the UUID 🤔

GrahamS-Quartech (approved):

PR #1097, opened 2023-09-19
SIMSBIOHUB-223: Observations Prototype

curtisupshall:
## Links to Jira Tickets
 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-223
 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-225
 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-280

## Description of Changes

- Adds migration to create the survey observations table
- Adds a DataGrid component for creating, editing and deleting observation records
- Adds the Observation Context for maintaining survey observations
- Adds an endpoint and service and repo methods for update/inserting and retrieving survey observation records

## Testing Notes
 > **Note**: Requires `make clean web`

TBD.


api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
```diff
@@ -0,0 +1,327 @@
+import { RequestHandler } from 'express';
+import { Operation } from 'express-openapi';
+import { authorizeRequestHandler } from '../../../../../../request-handlers/security/authorization';
+import { getLogger } from '../../../../../../utils/logger';
+import { PROJECT_PERMISSION, SYSTEM_ROLE } from '../../../../../../constants/roles';
+import { getDBConnection } from '../../../../../../database/db';
+import { ObservationService } from '../../../../../../services/observation-service';
+import { InsertObservation, UpdateObservation } from '../../../../../../repositories/observation-repository';
+
+const defaultLog = getLogger('/api/project/{projectId}/survey/{surveyId}/observation/get');
```
KjartanE: is the default log get? maybe just leave at `/api/project/{projectId}/survey/{surveyId}/observation/`

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
```diff
@@ -0,0 +1,327 @@
...
+  responses: {
+    200: {
+      description: 'Survey Observations get response.',
+      content: {
+        'application/json': {
+          schema: {
+            title: 'Survey get response object, for view purposes',
+            type: 'object',
+            nullable: true,
+            required: ['surveyObservationData', 'surveyObservationSupplementaryData'],
```
KjartanE: Leftover code 'surveyObservationSupplementaryData' thats not referenced

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
```diff
@@ -0,0 +1,327 @@
...
+    }
+  ],
+  requestBody: {
+    description: 'Survey observation record data',
+    content: {
+      'multipart/form-data': {
+        schema: {
+          type: 'object',
+          properties: {
+            surveyObservations: {
```
KjartanE, line 211: No required on the `surveyObservations`

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
```diff
@@ -0,0 +1,327 @@
...
+  },
+  responses: {
+    200: {
+      description: 'Upload OK',
+      content: {
+        'application/json': {
+          schema: {
+            type: 'object',
+            properties: {
+              // TODO do we need to include anything in this response?
```
KjartanE: Currently a lot of puts are just returning 200 or 201 with no data returned.

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
```diff
@@ -0,0 +1,327 @@
...
+
+    const connection = getDBConnection(req['keycloak_token']);
+
+    try {
+      await connection.open();
+
+      const observationService = new ObservationService(connection);
+      
+      /*
+      const taxonomyService = new TaxonomyService();
```
KjartanE: Any reason to keep around?

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
```diff
@@ -0,0 +1,327 @@
...
+          latitude: record.latitude,
+          longitude: record.longitude,
+          count: record.count,
+          observation_datetime: new Date(`${record.date} ${record.time}`)
+        }
+      });
+
+      await observationService.insertUpdateSurveyObservations(records);
+
+      return res.status(200).json({});
```
KjartanE: return res.status(200).send();

api/src/services/observation-service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * TODO
+   *
+   * @param {((Observation | ObservationRecord)[])} observations
+   * @return {*}  {Promise<ObservationRecord[]>}
+   * @memberof ObservationService
+   */
+  async insertUpdateSurveyObservations(observations: (InsertObservation | UpdateObservation)[]): Promise<ObservationRecord[]> {
+    defaultLog.debug({ label: 'insertUpdateObservationRecords' })
```
KjartanE: not sure if this debug is needed

app/src/contexts/observationsContext.tsx
```diff
@@ -0,0 +1,175 @@
...
+export interface IObservationTableRow extends Partial<IObservationRecord> {
+  id: string;
+}
+
+export const fetchObservationDemoRows = async (): Promise<IObservationRecord[]> => {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      resolve([
+        {
+          observation_id: 1,
```
KjartanE: Hard coded?

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
curtisupshall: Good catch

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
curtisupshall: Yeah, agreed.

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
curtisupshall: Yes, this endpoint is a work in progress. It depends how we want to map `spaciesName` to their respective taxonomic units code.

app/src/contexts/observationsContext.tsx
curtisupshall: Yeah these are just for testing purposes

api/src/services/observation-service.ts
curtisupshall: Agreed, it's just there for testing purposes

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
```diff
@@ -0,0 +1,343 @@
...
+
+    try {
+      await connection.open();
+
+      const observationService = new ObservationService(connection);
+
+      // Sanitize all incoming records
+      const records: (InsertObservation | UpdateObservation)[] = req.body.surveyObservations.map((record: any) => {
+        return {
+          // survey_id: surveyId, // TODO check if this can be removed.
```
GrahamS-Quartech: Can this line be removed?

api/src/repositories/observation-repository.ts
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Performs an upsert for all observation records belonging to the given survey, then
+   * returns the updated rows
+   *
+   * @param {number} surveyId
+   * @param {((InsertObservation | UpdateObservation)[])} observations
+   * @return {*}  {Promise<ObservationRecord[]>}
+   * @memberof ObservationRepository
+   */
+  async insertUpdateSurveyObservations(
```
GrahamS-Quartech, line 94: It may be possible to do this one with the Knex merge function.

app/src/contexts/observationsContext.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const biohubApi = useBiohubApi();
+  const surveyContext = useContext(SurveyContext);
+  const { projectId, surveyId } = useContext(SurveyContext);
+  const observationsDataLoader = useDataLoader(() => biohubApi.observation.getObservationRecords(projectId, surveyId));
+  const [unsavedRecordIds, _setUnsavedRecordIds] = useState<string[]>([]);
+
+  observationsDataLoader.load();
+
+  const markRecordWithUnsavedChanges = (id: string) => {
+    const unsavedRecordSet = new Set<string>([...unsavedRecordIds]);
```
GrahamS-Quartech: Don't think this spread is needed

database/src/migrations/20230925103600_create_survey_observations.ts
```diff
@@ -0,0 +1,199 @@
...
+     --                  2023-08-24  delete partnerships
+     -- curtis.upshall@quartech.com
+     --                  2023-08-24  delete survey blocks and stratums and participation
+     -- curtis.upshall@quartech.com
+     --                  2023-09-25  delete survey observations
+     -- *******************************************************************
+     declare
+
+     begin
+      with occurrence_submissions as (select occurrence_submission_id from occurrence_submission where survey_id = p_survey_id), submission_spatial_components as (select submission_spatial_component_id from submission_spatial_component
```
GrahamS-Quartech, line 144: Maybe this function was just always like this but I feel like all lowercase SQL keywords hurts readability.

database/src/migrations/20230925103600_create_survey_observations.ts
```diff
@@ -0,0 +1,199 @@
...
+  -- Create new survey_observation table
+  ----------------------------------------------------------------------------------------
+
+  SET search_path=biohub;
+
+  CREATE TABLE survey_observation(
+    survey_observation_id                   integer           GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+    survey_id                               integer           NOT NULL,
+    wldtaxonomic_units_id                   integer,
+    latitude                                float             NOT NULL,
```
GrahamS-Quartech: I think check constraints on lat / lon would make sense, the range is well known and is never gonna change.

api/src/paths/project/{projectId}/survey/{surveyId}/observation/index.ts
curtisupshall: It can!

api/src/repositories/observation-repository.ts
curtisupshall: I spent a lot of time workshopping this with Knex `merge` and could not get it to work, chiefly because entering a raw SQL allows me to `OVERRIDING SYSTEM VALUE` when merging the `survey_observation_id`

database/src/migrations/20230925103600_create_survey_observations.ts
curtisupshall: Great point. I did some research and settled on the Postgres [numeric](https://www.postgresql.org/docs/current/datatype-numeric.html) type, which gives you custom precision. I'm giving it 7 points of decimal precision.

database/src/migrations/20230925103600_create_survey_observations.ts
curtisupshall: It has always been like this (I think). While I generally agree, there is no guarentee that any future migrations will respect such a change, since by definition, migrations are one-off scripts that generally don't follow readibility conventions.

app/src/contexts/observationsContext.tsx
curtisupshall: I usually clone arrays when passing into a consturctor, but I agree; it's not needed.

api/src/repositories/observation-repository.ts
```diff
@@ -0,0 +1,164 @@
...
+        survey_observation_id,
+        survey_id,
+        wldtaxonomic_units_id,
+        count,
+        latitude,
+        longitude,
+        observation_date,
+        observation_time
+      )
+      OVERRIDING SYSTEM VALUE
```
al-rosenthal, line 111: what is this line for?

api/src/repositories/observation-repository.ts
```diff
@@ -0,0 +1,164 @@
...
+            observation.longitude,
+            `'${moment(observation.observation_date).format('YYYY-MM-DD')}'`,
+            `'${observation.observation_time}'`
+          ].join(', ')})`;
+        })
+        .join(', ')
+    );
+
+    sqlStatement.append(`
+      ON CONFLICT
```
al-rosenthal, line 133: 🙏

al-rosenthal (dismissed):
ran this locally and the add/ edit and delete all work as expected

api/src/repositories/observation-repository.ts
curtisupshall: Basially, when we do an `INSERT ON CONFLICT DO UPDATE`, we key on `survey_observation_id`, but to accomodate this, you have to pass `survey_observation_id` to the insert wuery. Postgres will complain that you're trying to peform an insert on a column that is supposed to be auto-generated. Thus, `OVERRIDING SYSTEM VALUE` allows you to bypass this, and the `DEFAULT` above refers to said auto-generated value.

api/src/repositories/observation-repository.ts
curtisupshall: ⚔️

api/src/repositories/observation-repository.ts
al-rosenthal: oooh I see

al-rosenthal (approved):
🌞

database/src/migrations/20230925103600_create_survey_observations.ts
GrahamS-Quartech: I meant to add check constraints in the form of something like CHECK(latitude >= -90 AND latitude <= 90). Specifying the level of precision also makes sense but I don't think it accomplishes the same thing.

curtisupshall: I don't think this kind of audit is needed. The `latitude` and `longitude` fields really just reflect the raw data entered in the form, with enough precision to reasonably accept any valid lat/long. Without any requirements associated with how these fields are going to be used in the future, adding constraints on them now would probably be overkill. 

To that end, if we do choose to query these points or need to otherwise validate them in the future, we will probably retain the raw `latitude` and `longitude`, and make an additional column that stores them as a `geography`-like datatype that is design specifically for lat/long points.

GrahamS-Quartech: Well the form should probably be catching this way before it gets to the database, but 99 latitude is just not a correct value and can easily be rejected here, the CHECK constraint can even be added inline with the column declaration. But if you think the requirements on how these columns are going to be used are that uncertain then it's fine to go without.

curtisupshall: I think the issue is not that it's too unweildy to validate these by the database, but that it's not appropriate for the database to throw an exception for recieving a bogus latitude/longitude, since these columns by design do not need to be valid.

NickPhura (approved):

PR #1101, opened 2023-09-25
SIMSBIOHUB 217 - Edit Animal

GrahamS-Quartech:
## Links to Jira Tickets

[SIMSBIOHUB-217](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-217)

## Description of Changes

- The button for Edit Animal has been restored in the three dots action menu within rows of the survey animal table. 
- Selecting this button will bring up an edit version of the animal form. All previously recorded values should be auto filled.
- You can edit existing records, add new records, and delete records all from the same form. 
- Critter ID has been moved out of the animal table to instead be displayed at the top of this form.

## Testing Notes

- ~~A patch to Critterbase is required to support some of the new delete / update operations, will edit this comment once it's in.~~


api/src/repositories/survey-critter-repository.ts
```diff
@@ -43,6 +43,16 @@ export class SurveyCritterRepository extends BaseRepository {
     return response.rowCount;
   }
 
+  async updateCritter(surveyId: number, critterId: string): Promise<number> {
+    defaultLog.debug({ label: 'updateCritter', surveyId });
+    const queryBuilder = getKnex()
+      .table('critter')
+      .update({ survey_id: surveyId, critterbase_critter_id: critterId })
+      .where({ survey_id: surveyId, critterbase_critter_id: critterId });
```
curtisupshall: This query doesn't make a lot of sense to me. The `WHERE` predicate matches the body of the update, so nothing is actually being modified... Unless I'm missing something?

app/src/constants/i18n.ts
```diff
@@ -331,11 +331,17 @@ export const SurveyAnimalsI18N = {
   animalMortalityHelp:
     "Mortality Events describe an individual's death, including the suspected location, date, and cause of death. An individual can only have one Mortality Event.",
   animalMortalityAddBtn: 'Add Mortality',
-
+  animalCollectionUnitTitle: 'Ecological Units',
+  animalCollectionUnitTitle2: 'Ecological Unit',
+  animalCollectionUnitHelp:
+    'Ecological units are groups like population units, herds, and packs. Different species may different units and unit names.',
```
curtisupshall: Nit: first sentence, I would use "such as " instead of "like"; Second sentence contains a typo

app/src/constants/i18n.ts
```diff
@@ -331,11 +331,17 @@ export const SurveyAnimalsI18N = {
...
+  animalCollectionUnitAddBtn: 'Add Unit',
   // Input help strings
   taxonHelp:
     'The species or taxon of the animal. If the species is unknown, select the lowest-ranking known taxon, such as the genus or family',
-  taxonLabelHelp: 'A custom and unique name for you to recognize this individual'
+  taxonLabelHelp: 'A custom and unique name for you to recognize this individual',
+  wlhIdHelp: 'The wildlife health ID associated with this critter.',
```
curtisupshall: Wildlife Health needs capitalization

app/src/constants/i18n.ts
```diff
@@ -331,11 +331,17 @@ export const SurveyAnimalsI18N = {
...
+  sexHelp: 'The sex of this critter. Leave as default Unknown if unsure.'
```
curtisupshall: Nit: "Leave as Unknown if unsure."

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -99,6 +107,22 @@ const SurveyAnimals: React.FC = () => {
...
+  const obtainAnimalFormInitialvalues = (mode: ANIMAL_FORM_MODE): IAnimal => {
+    switch (mode) {
+      case ANIMAL_FORM_MODE.ADD:
+        return AnimalFormValues;
+      case ANIMAL_FORM_MODE.EDIT: {
+        const existingCritter = critterData?.find(
+          (critter: IDetailedCritterWithInternalId) => currentCritterbaseCritterId === critter.critter_id
+        );
+        if (!existingCritter) {
+          throw Error('This should not be reachable.');
```
curtisupshall: I don't think I'd find this error very helpful if I saw this in prod. Can we just `break` instead? Or is there an otherwise more graceful way to recover?

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -128,24 +152,33 @@ const SurveyAnimals: React.FC = () => {
...
     try {
-      await postCritterPayload();
+      if (animalFormMode === ANIMAL_FORM_MODE.ADD) {
+        await postCritterPayload();
+      } else {
+        await patchCritterPayload();
+      }
     } catch (err) {
+      setPopup('Submission failed.');
       console.log(`Critter submission error ${JSON.stringify(err)}`);
```
curtisupshall: Nit: remove console log

app/src/features/surveys/view/survey-animals/IndividualAnimalForm.tsx
```diff
@@ -35,15 +43,21 @@ const IndividualAnimalForm = ({ getAnimalCount }: IndividualAnimalFormProps) =>
...
       <GeneralAnimalForm />
+      <CollectionUnitAnimalForm />
       <CaptureAnimalForm />
       <MortalityAnimalForm />
       <MarkingAnimalForm />
       <MeasurementAnimalForm />
       <FamilyAnimalForm />
 
-      <FormikDevDebugger custom_payload={new Critter(values)} />
+      <FormikDevDebugger custom_payload={undefined /*new Critter(values)*/} />
```
curtisupshall: Should we leave this dev debugger in?

app/src/features/surveys/view/survey-animals/SurveyAnimalsTable.tsx
```diff
@@ -47,17 +50,17 @@ export const SurveyAnimalsTable = ({
...
       field: 'animal_id',
-      headerName: 'Animal ID',
+      headerName: 'Alias',
       flex: 1
     },
+    {
+      field: 'wlh_id',
+      headerName: 'WLH ID',
+      flex: 1,
+      renderCell: (params) => <Typography>{params.value ? params.value : 'None'}</Typography>
```
curtisupshall: Can just coalesce this: `params.value ?? 'None'`

al-rosenthal:
I can add a critter just fine but when I click the edit button it throws and error 'Cannot read properties of undefined' when trying to map something.

app/src/features/surveys/view/survey-animals/animal-form-helpers.test.ts
```diff
@@ -0,0 +1,139 @@
...
+import { v4 } from 'uuid';
+import { IAnimal, IAnimalCapture, IAnimalMarking } from './animal';
+import { arrDiff, createCritterUpdatePayload, transformCritterbaseAPIResponseToForm } from './animal-form-helpers';
+
+describe('animal form helpers', () => {
+  describe('transformCritterbaseAPIResponseToForm', () => {
+    it('should return an object matching the IAnimal interface', () => {
+      const detailedResponse: IDetailedCritterWithInternalId = {
+        survey_critter_id: 1,
+        critter_id: v4(),
```
curtisupshall: I think it makes more sense to hardcode test values like UUIDs, so that they can be directly compared with a string literal.
```javascript
expect(result.general.critter_id).toEqual('3347d57e-033a-4a6f-9b39-893ffb590510');
```

In this case, you'd know for certain that `taxon_id` didn't get mutated inside `transformCritterbaseAPIResponseToForm`

curtisupshall (commented):
Review part 1/2

api/src/repositories/survey-critter-repository.ts
GrahamS-Quartech: Yeah, it's effectively just a placeholder, since the critter table on SIMS end doesn't really hold much else of note. Doing the update operation will fill in the update audit columns at least.

app/src/constants/i18n.ts
GrahamS-Quartech: Fixed in merge conflict

app/src/features/surveys/view/survey-animals/animal.ts
```diff
@@ -256,6 +283,8 @@ export class Critter {
   critter_id: string;
   taxon_id: string;
   animal_id: string;
+  wlh_id?: string;
+  sex?: string;
```
curtisupshall: Didn't we make this an enum?

curtisupshall (commented):
Review part 2/2

api/src/repositories/survey-critter-repository.ts
curtisupshall: OK gotcha. Might make sense to leave a comment if we're gonna leave it like this for now

app/src/features/surveys/view/survey-animals/SurveyAnimalsTable.tsx
GrahamS-Quartech: Doing ternary will also display None if this value happens to be empty string. I think that's desirable.

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
```diff
@@ -163,6 +182,73 @@ POST.apiDoc = {
...
+  ],
+  requestBody: {
+    description: 'Critterbase bulk creation request object',
+    content: {
+      'application/json': {
+        schema: {
+          type: 'object',
+          properties: {
+            create: {
+              type: 'object'
```
KjartanE: Can this object params be declared? Same on `update:`

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
```diff
@@ -163,6 +182,73 @@ POST.apiDoc = {
...
+      }
+    }
+  },
+  responses: {
+    200: {
+      description: 'Responds with counts of objects created in critterbase.',
+      content: {
+        'application/json': {
+          schema: {
+            title: 'Bulk creation response object',
```
KjartanE: Same as above, declaring input output objects

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
```diff
@@ -163,6 +182,73 @@ POST.apiDoc = {
   }
 };
 
+PATCH.apiDoc = {
```
KjartanE: You added this update "PATCH" to the index file. To follow endpoints done in SIMS we have been putting Update and Delete functions in the `_id` file. Example of this is `api\src\paths\project\{projectId}\participants` with the `index.ts` file and `{projectParticipationId}.ts`

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
KjartanE: Is it proper to have two separate objects for this as a PATCH endpoint? 
A similar problem has been approached in SIMS around updating, our solution is to implore `upsert` functions.  See `survey-service.ts` @ `upsertSurveyFundingSourceData()`
Where a single update object is given and each item is checked to be updated/added/removed

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
```diff
@@ -235,3 +321,38 @@ export function addCritterToSurvey(): RequestHandler {
...
+      keycloak_guid: req['system_user']?.user_guid,
+      username: req['system_user']?.user_identifier
+    };
+    const surveyId = Number(req.params.surveyId);
+    const connection = getDBConnection(req['keycloak_token']);
+    const surveyService = new SurveyCritterService(connection);
+    const cb = new CritterbaseService(user);
+    try {
+      await connection.open();
+      if (!req.body.update.critter_id) {
```
KjartanE: This check can be removed if using the `{critter_id}.ts` file setup. As the endpoint cannot be reached without the param

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
```diff
@@ -235,3 +321,38 @@ export function addCritterToSurvey(): RequestHandler {
...
+        throw Error('No critter id found.');
+      }
+      await surveyService.updateCritter(surveyId, req.body.update.critter_id);
```
KjartanE: You dont use the response of the service. So the returns on the serivce and repo layer are unused and can be removed

api/src/repositories/survey-critter-repository.ts
```diff
@@ -43,6 +43,16 @@ export class SurveyCritterRepository extends BaseRepository {
...
+    const response = await this.connection.knex(queryBuilder);
+    return response.rowCount;
```
KjartanE: Not needed

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
GrahamS-Quartech: Maybe so, but I think that would involve too significant of a change to critterbase for the scope of this ticket.

app/src/features/surveys/view/survey-animals/SurveyAnimalsTable.tsx
curtisupshall: Ahh you're right. I think I meant to say `params.value || 'None'`, which is like a falsy coalesce rather than `??` which is a null coalesce... I think?

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
curtisupshall: https://stackoverflow.com/questions/28459418/use-of-put-vs-patch-methods-in-rest-api-real-life-scenarios

`PATCH` is for partial record updates, so the upsert pattern is a little more on-brand for a `PUT` request.

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
curtisupshall: Agreed, this should be moved as a URL parameter. That is, unless we're deleting several records at once

api/src/repositories/survey-critter-repository.ts
curtisupshall: Returning row count can be helpful, but I agree; If we aren't directly reading this value, we should just return a void promise and keep our return types as simple as possible.

api/src/paths/project/{projectId}/survey/{surveyId}/critters/index.ts
GrahamS-Quartech: I consciously did not do this to keep it more in line with the usage in Critterbase, where you are actually allowed to update multiple critters at once. But I dunno if it will ever see that usage here, so the argument to structure it that way could make sense.

app/src/features/surveys/view/survey-animals/SurveyAnimalsTable.tsx
NickPhura: ^ correct

app/src/components/fields/CbSelectField.tsx
```diff
@@ -50,12 +50,16 @@ const CbSelectField: React.FC<ICbSelectField> = (props) => {
   useEffect(() => {
     if (hasLoaded) {
       // Only refresh when the query or param changes
-      isMounted() && refresh({ route, param, query });
+      refresh({ route, param, query });
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [query, param]);
 
   const isValueInRange = useMemo(() => {
```
NickPhura: It seems like this function is doing more than it should be, based on the name.

It looks like it should just a basic 'is' function (is value in range or not), but it is also calling 'setFieldValue'. Possibly that should be pulled out, so that this function can just be a pure 'is' function? And the setFieldvalue calls should be made separately after this call is made

NickPhura: You wouldn't need the isMounted check either then, as this function wouldn't have any set state calls. Would be simpler overal.

app/src/features/surveys/view/survey-animals/IndividualAnimalForm.tsx
GrahamS-Quartech: Yeah, it's useful. Removed it since calling the constructor could affect one of the form values. That's unwanted behaviour of course, corrected it now.

app/src/features/surveys/view/SurveyAnimals.tsx
GrahamS-Quartech: It also crashes the app which is extra unhelpful. I think throwing here is appropriate but I implemented a way to handle it, let me know what you think.

api/src/openapi/schemas/critter.ts
```diff
@@ -1,6 +1,192 @@
...
+    },
+    release_location_id: {
+      type: 'string',
+      format: 'uuid'
+    },
+    capture_location: {
+      type: 'object'
+    },
+    release_location: {
+      type: 'object'
```
curtisupshall: Nit: Add object schema if it's available

api/src/openapi/schemas/critter.ts
```diff
@@ -1,6 +1,192 @@
...
+    location_id: { type: 'string', format: 'uuid' },
+    mortality_comment: { type: 'string' },
+    proximate_cause_of_death_id: { type: 'string', format: 'uuid' },
+    proximate_cause_of_death_confidence: { type: 'string' },
+    proximate_predated_by_taxon_id: { type: 'string', format: 'uuid' },
+    ultimate_cause_of_death_id: { type: 'string', format: 'uuid' },
+    ultimate_cause_of_death_confidence: { type: 'string' },
+    ultimate_predated_by_taxon_id: { type: 'string', format: 'uuid' },
+    projection_mode: { type: 'string', enum: ['wgs', 'utm'] },
+    location: { type: 'object' }
```
curtisupshall: Nit: Add object schema if it's available

api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}.ts
```diff
@@ -101,3 +167,40 @@ export function removeCritterFromSurvey(): RequestHandler {
...
+
+    const critterId = Number(req.params.critterId);
+    const connection = getDBConnection(req['keycloak_token']);
+    const surveyService = new SurveyCritterService(connection);
+    const cb = new CritterbaseService(user);
+    try {
+      await connection.open();
+      const critterbaseCritterId = req.body.update.critter_id;
+      if (!critterbaseCritterId) {
+        throw Error('No external critter id found.');
```
curtisupshall: Should probably throw some kind of HTTP error instead... Maybe 404?

api/src/repositories/survey-critter-repository.ts
```diff
@@ -66,16 +73,15 @@ export class SurveyCritterRepository extends BaseRepository {
    * @returns {*}
    * @memberof SurveyCritterRepository
    */
-  async upsertDeployment(critterId: number, deplyomentId: string): Promise<number> {
+  async upsertDeployment(critterId: number, deplyomentId: string): Promise<void> {
     defaultLog.debug({ label: 'addDeployment', deplyomentId });
+    //This update operation intentionally changes nothing. Only really being done to trigger update audit columns.
```
curtisupshall: Any important comments like this that convey business logic should be moved to the JS docs.

api/src/services/survey-critter-service.ts
```diff
@@ -26,17 +26,28 @@ export class SurveyCritterService extends DBService {
    * @param {string} critterId
    * @returns {*}
    */
-  async addCritterToSurvey(surveyId: number, critterBaseCritterId: string): Promise<number> {
-    return this.critterRepository.addCritterToSurvey(surveyId, critterBaseCritterId);
+  async addCritterToSurvey(surveyId: number, critterBaseCritterId: string): Promise<void> {
+    this.critterRepository.addCritterToSurvey(surveyId, critterBaseCritterId);
```
curtisupshall: Either `return` this Promise, or `await` it. In its current state, this isn't really an async method because it dispatches `critterRepository.addCritterToSurvey` and then immediately returns void.

api/src/services/survey-critter-service.ts
```diff
@@ -26,17 +26,28 @@ export class SurveyCritterService extends DBService {
...
 
   /**
    * Removes a critter from the survey. Does not affect the critter in the external system.
    * @param {string} critterId
    * @returns {*}
    */
-  async removeCritterFromSurvey(critterId: number): Promise<number> {
-    return this.critterRepository.removeCritterFromSurvey(critterId);
+  async removeCritterFromSurvey(critterId: number): Promise<void> {
+    this.critterRepository.removeCritterFromSurvey(critterId);
```
curtisupshall: Echoing `return` or `await`

api/src/services/survey-critter-service.ts
```diff
@@ -46,7 +57,7 @@ export class SurveyCritterService extends DBService {
    * @param {id} deplyomentId
    * @returns {*}
    */
-  async upsertDeployment(critterId: number, deplyomentId: string): Promise<number> {
-    return this.critterRepository.upsertDeployment(critterId, deplyomentId);
+  async upsertDeployment(critterId: number, deplyomentId: string): Promise<void> {
+    this.critterRepository.upsertDeployment(critterId, deplyomentId);
```
curtisupshall: Echoing `return` or `await`

api/src/services/survey-critter-service.ts
curtisupshall: I would prefer using `return` here, but either will achieve the same result

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -129,25 +153,91 @@ const SurveyAnimals: React.FC = () => {
...
+        />
+      );
+    }
+  };
+
+  const handleCritterSave = async (currentFormValues: IAnimal) => {
     const postCritterPayload = async () => {
+      const critter = new Critter(currentFormValues);
+      toggleDialog();
       await bhApi.survey.createCritterAndAddToSurvey(projectId, surveyId, critter);
```
curtisupshall, line 222: May want to wrap this in a try-catch

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -129,25 +153,91 @@ const SurveyAnimals: React.FC = () => {
...
-            {'Animal added to Survey'}
-          </Typography>
-        )
-      });
+      setPopup('Animal added to survey.');
+    };
+    const patchCritterPayload = async () => {
+      const initialFormValues = obtainAnimalFormInitialvalues(ANIMAL_FORM_MODE.EDIT);
+      if (!initialFormValues) {
+        throw Error('Could not obtain initial form values.');
```
curtisupshall, line 229: Throwing here won't crash the app, but it may provide a worse experience than just showing this text in a snackbar or dialog instead.

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -129,25 +153,91 @@ const SurveyAnimals: React.FC = () => {
...
+      }
+      const { create: createCritter, update: updateCritter } = createCritterUpdatePayload(
+        initialFormValues,
+        currentFormValues
+      );
       toggleDialog();
+      if (!selectedCritterId) {
+        throw Error('The internal critter id for this row was not set correctly.');
```
curtisupshall, line 237: echoing throw comment

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -129,25 +153,91 @@ const SurveyAnimals: React.FC = () => {
...
+      }
+      await bhApi.survey.updateSurveyCritter(projectId, surveyId, selectedCritterId, updateCritter, createCritter);
```
curtisupshall, line 239: Echoing try-catch comment

app/src/features/surveys/view/SurveyAnimals.tsx
GrahamS-Quartech: It is caught a little ways below this.

app/src/features/surveys/view/SurveyAnimals.tsx
GrahamS-Quartech: I do have it pop up a snack bar when catching these, would probably be better to have the specific message placed in the snack bar though.

curtisupshall (dismissed):
Code looks great!

JeremyQuartech (dismissed):
Looks good!

JeremyQuartech (approved):
Looks great!

curtisupshall (approved):

PR #1102, opened 2023-09-25
SIMSBIOHUB-286: Upload Telemetry Files on Device Creation

JeremyQuartech:
## Links to Jira Tickets

- [SIMSBIOHUB-286](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-286)

## Description of Changes

- Adds option to attach a .keyx or .cfg file during new telemetry device creation
- 'keyx' upload section conditionally renders when Vectronic manufacturer is selected
- 'cfg' upload section conditionally renders when Lotek manufacturer is selected


app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -151,15 +155,24 @@ const SurveyAnimals: React.FC = () => {
...
+          await bhApi.survey.uploadSurveyKeyx(projectId, surveyId, attachmentFile);
+        } else if (attachmentFile && attachmentType === AttachmentType.OTHER) {
+          await bhApi.survey.uploadSurveyAttachments(projectId, surveyId, attachmentFile);
+        }
         await bhApi.survey.addDeployment(projectId, surveyId, survey_critter_id, critterTelemetryDevice);
         setPopup('Successfully added deployment.');
-      } catch (e) {
-        setPopup('Failed to add deployment.');
+        surveyContext.artifactDataLoader.refresh(projectId, surveyId);
+      } catch (e: any) {
```
NickPhura: super nit: lets not use single letter variables in general. `error` is just better imo.

Also, lets avoid casting stuff to `any` whenever possible. Better to cast it to `Error` (or w/e is appropriate here) as needed

NickPhura: For line 191:

I'd prefer `item` over `a`, or `deploymentDataItem` or something. Better yet, is deploymentData of a certain known type? Using that type as part of the variable name can be nice too, just to add some more context.

Dug into it: looks like deploymentData is of type `IAnimalDeployment[]`. So I'd even call it `animalDeployments` or similar, so it is a bit more descriptive of its own type. And then in the map, it could be `animalDeployment` instead of `a`.

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -151,15 +155,24 @@ const SurveyAnimals: React.FC = () => {
     }
   };
 
-  const handleTelemetrySave = async (survey_critter_id: number, data: IAnimalTelemetryDevice[]) => {
+  const handleTelemetrySave = async (survey_critter_id: number, data: IAnimalTelemetryDeviceFile[]) => {
```
NickPhura: This function is a bit complicated, could we break it down into more smaller functions? Maybe the ADD content is its own function, same with the EDIT?

NickPhura: Also, adding comments inline is always okay. 

If we look at this or any code from the perspective of a brand new developer with no historical context on the project, are there any parts of the code that could use an explanation because what its doing, or why, is not obvious or involves some business knowledge that they just wouldn't have.

app/src/features/surveys/view/survey-animals/TelemetryDeviceForm.tsx
```diff
@@ -3,18 +3,44 @@ import Grid from '@mui/material/Grid';
...
   ADD = 'add',
   EDIT = 'edit'
 }
 
+export interface IAnimalTelemetryDeviceFile extends IAnimalTelemetryDevice {
+  attachmentFile?: File;
+  attachmentType?: AttachmentType;
+}
+
+const AttchmentFormSection = ({ index, deviceMake }: { index: number; deviceMake: string }): JSX.Element => {
```
curtisupshall: Nit: Remove `JSX.Element` and use `props` parameter instead of destructuring

curtisupshall (dismissed):
Code looks good to my eyes :^)

GrahamS-Quartech (approved):
Looking good.

api/src/services/bctw-service.ts
```diff
@@ -330,6 +346,10 @@ export class BctwService {
     };
   }
 
+  async getKeyXDetails(deviceIds: number[]): Promise<IKeyXDetails[]> {
```
curtisupshall, line 349: Needs JS docs

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -151,44 +160,91 @@ const SurveyAnimals: React.FC = () => {
...
-  const handleTelemetrySave = async (survey_critter_id: number, data: IAnimalTelemetryDevice[]) => {
+  const uploadAttachment = async (file?: File, attachmentType?: AttachmentType) => {
+    try {
+      if (file && attachmentType === AttachmentType.KEYX) {
+        await bhApi.survey.uploadSurveyKeyx(projectId, surveyId, file);
+      } else if (file && attachmentType === AttachmentType.OTHER) {
+        await bhApi.survey.uploadSurveyAttachments(projectId, surveyId, file);
+      }
+    } catch (error) {
+      throw new Error(`Failed to upload attachment ${file?.name}`);
```
curtisupshall, line 171: We should avoid throwing error in the browser. We probably need to audit the entire application for unsafe error handling in general. Ideally, error messages should be displayed in the app using React components.

curtisupshall (approved):
Code looks good

al-rosenthal (approved):
👍

PR #1104, opened 2023-09-29
SIMSBIOHUB-275: Remove Project Coordinator from SIMS

curtisupshall:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-275

## Description of Changes
 - Removes any Project Coordinator UI for creating, editing or viewing projects
 - Removes any reference to Project Coordinator in the frontend and backend codebase
 - Adds a migration that makes project coordinator fields optional, retaining any existing project coordinator data.

## Testing Notes (for local environment)

1. On branch `dev`, create one or more projects.
2. Check out the `SIMSBIOHUB-275` branch and run `make web` to run the new migration.
3. [ ] Upon compiling the project, you should be able to:
 - View an existing project + participants
 - Edit an existing project
 - Create a new project
 - View a survey + participants
 - Add participants to a project
 - Add participants to a survey

4. [ ] For the projects created before running the migration, their participant data in the project table should still exist.


database/src/migrations/20230929103700_ignore_project_coordinator.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
+    SET search_path = 'biohub';
+
+    ALTER TABLE project
+    ALTER COLUMN coordinator_first_name       DROP NOT NULL,
```
NickPhura: Would you mind also updating the comments for the coordinator fields, and just appending something like "Deprecated" to the start. Just in case.

NickPhura:
Looks good. 

A few tests on front and back end that are referencing project coordinator still.

api/src/services/eml-service.ts
```diff
@@ -666,44 +666,36 @@ export class EmlService extends DBService {
...
-    const primaryContact = projectData.coordinator;
-
-    if (JSON.parse(primaryContact.share_contact_details)) {
-      // return full details of the primary contact iff it is public.
-      return {
-        organizationName: primaryContact.coordinator_agency,
-        electronicMailAddress: primaryContact.email_address
-      };
-    }
+    const participant = projectData.participants[0];
```
NickPhura: In the few places that grab the 0th element, would you mind changing it to instead find a participant with the coordinator role? It should always exist, though I guess the find won't know that.. Might have to bring that if else structure back from the old version, just to appease the type safety.

NickPhura: (It probably doesn't really matter in the end, but we've come this far :/)

database/src/migrations/20230929103700_ignore_project_coordinator.ts
curtisupshall: Great idea, added.

JeremyQuartech:
I may be misunderstanding the changes a bit here, but just clarifying... When creating a project, under managing team members, `Coordinator` is still meant to be an option right? So do these changes simply remove the ability to add metadata associated with the coordinator?

api/src/services/eml-service.ts
curtisupshall: Good call. Changed!

curtisupshall:
> I may be misunderstanding the changes a bit here, but just clarifying... When creating a project, under managing team members, `Coordinator` is still meant to be an option right? So do these changes simply remove the ability to add metadata associated with the coordinator?

@JeremyQuartech That is essentially it, yeah. Project team members did not historically have roles like "Coordinator", so projects has specific metadata (Project Coordinator) to store basic info for the primary contact for the project. Since we now have the Coordinator team member role, this information is pretty much unneeded.

NickPhura (dismissed):

database/src/migrations/20230929103700_deprecate_project_coordinator.ts
```diff
@@ -0,0 +1,32 @@
...
+    SET search_path = 'biohub';
+
+    ALTER TABLE project
+    ALTER COLUMN coordinator_first_name       DROP NOT NULL,
+    ALTER COLUMN coordinator_last_name        DROP NOT NULL,
+    ALTER COLUMN coordinator_email_address    DROP NOT NULL,
+    ALTER COLUMN coordinator_agency_name      DROP NOT NULL,
+    ALTER COLUMN coordinator_public           DROP NOT NULL;
+
+    COMMENT ON COLUMN project.coordinator_first_name IS '(Deprecated) The first name of the person directly responsible for the project.';
```
al-rosenthal, line 22: love the depreciated tag

al-rosenthal (approved):
ran this locally and followed your instructions. It is working as expected 🐶

NickPhura (approved):

PR #1106, opened 2023-10-03
SIMSBIOHUB-223: Implement UI improvements + changes for observations table

curtisupshall:
## Description of Changes

Implements UI improvements + changes for observations table.

### DIFF
--- a/api/src/repositories/observation-repository.ts
+++ b/api/src/repositories/observation-repository.ts
@@ -164,11 +164,7 @@ export class ObservationRepository extends BaseRepository {
     const knex = getKnex();
     const sqlStatement = knex
       .queryBuilder()
-      .select(
-        '*',
-        knex.raw('latitude::double precision'),
-        knex.raw('longitude::double precision')
-      )
+      .select('*', knex.raw('latitude::double precision'), knex.raw('longitude::double precision'))
       .from('survey_observation')
       .where('survey_id', surveyId);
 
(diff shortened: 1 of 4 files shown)


api/src/repositories/observation-repository.ts
```diff
@@ -141,7 +141,12 @@ export class ObservationRepository extends BaseRepository {
         longitude = EXCLUDED.longitude
     `);
 
-    sqlStatement.append(`RETURNING *;`);
+    sqlStatement.append(`
+      RETURNING
+        *,
+        latitude::double precision,
```
al-rosenthal, line 147: working on the add/ edit for observations, I put a parse function in the `db.ts` that looks for the numeric type from postgres and converts it using `parseFloat`. It also means that whenever we need to add the lat/ long from this table we don't need to explicitly convert them, it is handled automatically.

al-rosenthal (commented):
I'm able to put in a negative count for an animal

api/src/repositories/observation-repository.ts
curtisupshall: For now, let's leave this conversion in; Currently, updating the table is broken in dev. Assuming this PR will get merged before yours, we can remove this conversion in the subsequent PR.

app/src/features/surveys/observations/ObservationsTable.tsx
```diff
@@ -14,6 +16,8 @@ const ObservationsTable = () => {
       flex: 1,
       minWidth: 250,
       disableColumnMenu: true,
+
+      // TODO: To be addressed by https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-288
```
al-rosenthal, line 20: 👍

al-rosenthal (approved):
ran this locally, it looks good 🌦️

NickPhura (approved):

PR #1107, opened 2023-10-03
SIMSBIOHUB-294: Delete Deployment

GrahamS-Quartech:
## Links to Jira Tickets

- [SIMSBIOHUB-294](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-294)

## Description of Changes

- Added a trash can icon next to each deployment in the Edit Deployment page. Does not appear on Add Deployment.
- Clicking the trash can will prompt you with a confirm dialogue. Accepting the dialogue should hit an endpoint in bctw api and remove the deployment from the list.
- If no deployments remain, then the dialogue automatically closes.
- Added verification of device make directly to the yup schema. Was sometimes missing changes to device id before with the previous set up, this seems more robust. 
- Fixed some date range checking logic that was missing some overlapping cases, which allowed you to submit the form and get a 500 back.


app/src/features/surveys/view/survey-animals/TelemetryDeviceForm.tsx
JeremyQuartech, line 1: I'm finding that for seemingly random device_ids, this form keeps getting stuck on this error: `This device make differs from the existing make for this Device ID`. It often won't allow the creation of a new device due to this error.

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -118,6 +118,22 @@ const SurveyAnimals: React.FC = () => {
...
 
+  const AnimalDeploymentSchemaAsyncValidation = AnimalTelemetryDeviceSchema.shape({
+    device_make: yup
+      .string()
+      .required()
+      .test(
+        'checkDeviceMakeIsNotChanged',
+        'This device make differs from the existing make for this Device ID',
+        async (value, context) => {
+          const deviceDetails = await telemetryApi.devices.getDeviceDetails(Number(context.parent.device_id));
```
curtisupshall: Are async calls allowed in Yup schema?

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -356,18 +373,31 @@ const SurveyAnimals: React.FC = () => {
     refreshCritters();
   };
 
+  const handleRemoveDeployment = async (deployment_id: string) => {
+    if (!selectedCritterId) {
+      throw Error('Survey critter ID was set incorrectly.');
+    }
+    await bhApi.survey.removeDeployment(projectId, surveyId, selectedCritterId, deployment_id);
+    const deployments = deploymentData?.filter((a) => a.critter_id === currentCritterbaseCritterId);
```
curtisupshall: Would coalesce any array-like variables to empty array so you can avoid nullish checks on arrays

app/src/features/surveys/view/survey-animals/TelemetryDeviceForm.tsx
```diff
@@ -22,48 +29,95 @@ export interface IAnimalTelemetryDeviceFile extends IAnimalTelemetryDevice {
...
 }): JSX.Element => {
+  const [openDeleteDialog, setOpenDeleteDialog] = useState(false);
+  const [deploymentToDelete, setDeploymentToDelete] = useState('');
   return (
     <>
-      <Grid container spacing={2}>
+      <YesNoDialog
+        dialogTitle={'Remove deployment?'}
+        dialogText={`Are you sure you want to remove this deployment? 
+          If you would like to end a deployment / unattach a device, you should instead set the attachment end date.
```
curtisupshall: Nit: revise "you should set the attachment end date instead"

app/src/hooks/api/useSurveyApi.ts
```diff
@@ -568,6 +568,18 @@ const useSurveyApi = (axios: AxiosInstance) => {
     return data;
   };
 
+  const removeDeployment = async (
```
curtisupshall, line 623: Add JS doc

curtisupshall (dismissed):

api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}/deployments/{bctwDeploymentId}.ts
```diff
@@ -0,0 +1,114 @@
...
+    },
+    {
+      in: 'path',
+      name: 'bctwDeploymentId',
+      schema: {
+        type: 'string',
+        format: 'uuid'
+      }
+    }
+  ],
```
NickPhura: These probably need to be marked as required

api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}/deployments/{bctwDeploymentId}.ts
```diff
@@ -0,0 +1,114 @@
...
+      description: 'Remove deployment successfully.'
+    },
+    400: {
+      $ref: '#/components/responses/400'
+    },
+    401: {
+      $ref: '#/components/responses/401'
+    },
+    403: {
+      $ref: '#/components/responses/401'
```
NickPhura: nit: issue from our old endpoints that we never noticed (403 doesn't match 401)

api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}/deployments/{bctwDeploymentId}.ts
```diff
@@ -0,0 +1,114 @@
...
+    const critterId = Number(req.params.critterId);
+    const connection = getDBConnection(req['keycloak_token']);
+    const surveyCritterService = new SurveyCritterService(connection);
+    const bctw = new BctwService(user);
+    try {
+      await connection.open();
+      const surveyEntry = await surveyCritterService.removeDeployment(critterId, deploymentId);
+      await bctw.deleteDeployment(deploymentId);
+      await connection.commit();
+      return res.status(200).json(surveyEntry);
```
JeremyQuartech: Survey entry is undefined because of the void return in the service. Maybe its worth just dropping the .json in the response all together. We should also probably choose a consistent way of doing this between this delete and the update endpoint which both have void returns.

app/src/features/surveys/view/SurveyAnimals.tsx
GrahamS-Quartech: Yup

app/src/features/surveys/view/survey-animals/TelemetryDeviceForm.tsx
```diff
@@ -22,48 +29,95 @@ export interface IAnimalTelemetryDeviceFile extends IAnimalTelemetryDevice {
...
                 />
               </Grid>
-              <Grid item xs={6}>
-                <SingleDateField name={`${index}.deployments.${i}.attachment_end`} label={'Attachment End'} />
+              <Grid item xs={mode === TELEMETRY_DEVICE_FORM_MODE.ADD ? 6 : 5.5}>
+                <SingleDateField name={`${index}.deployments.${i}.attachment_end`} label={'End Date'} />
               </Grid>
+              {mode === TELEMETRY_DEVICE_FORM_MODE.EDIT && (
+                <Grid item xs={1}>
+                  <IconButton
```
JeremyQuartech, line 111: nit: I think it looks a bit off, where the deployments are no longer aligned with the grid above due to this button. That being said, I don't really see a better option.

api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}/deployments/{bctwDeploymentId}.ts
GrahamS-Quartech: Updated to be in line with those other responses.

JeremyQuartech (dismissed):
Looks good, everything working as expected!

al-rosenthal (commented):
Creation and deletion of deployments works great but I did notice a few strange things with the form. I'm able to save end dates that are before the start dates. We have a custom formik rule that will check for these values. Both project and survey use these custom formik rules if you want to see an example. 

I also noticed that after creating a device that if I edit the device it flags these optional values as required and I cannot save the form.

Screenshot 2023-10-05 110304

JeremyQuartech:
> Creation and deletion of deployments works great but I did notice a few strange things with the form. I'm able to save end dates that are before the start dates. We have a custom formik rule that will check for these values. Both project and survey use these custom formik rules if you want to see an example.
> 
> I also noticed that after creating a device that if I edit the device it flags these optional values as required and I cannot save the form.

@al-rosenthal 
I can't seem to recreate the optional field errors that you're getting. Were any special steps required? The dates issue I definitely agree should be fixed.

al-rosenthal:
> > Creation and deletion of deployments works great but I did notice a few strange things with the form. I'm able to save end dates that are before the start dates. We have a custom formik rule that will check for these values. Both project and survey use these custom formik rules if you want to see an example.
> > I also noticed that after creating a device that if I edit the device it flags these optional values as required and I cannot save the form.
> 
> @al-rosenthal I can't seem to recreate the optional field errors that you're getting. Were any special steps required? The dates issue I definitely agree should be fixed.

@JeremyQuartech Steps to reproduce:
1. Add an animal a new animal
2. Add a new telemetry device without adding a frequency or a device model
3. Save
4. Then edit that same telemetry device and hit the save button

NickPhura:
Not sure if this is an issue or not, but just in case: 
image

The warning at the bottom is a bit funny in how its displayed, in particular the "from 2023-02-02T08:00:00.000Z until indefinite."

1. Created an animal
2. Went to add a device, and just picked some random values.


------------------


Also, when I enter a date, and then clear it, the error message is some default yup message.
image

Some of the yup checks probably just need some specific messages set, so they don't fallback to some default.


-----------------


Putting in valid values, I clicked save, and it maybe worked? But the list of animals doesnt show there is a device, and the menu still shows "Add Telemetry Device". But when I click it, it seems semi-broken (the list of frequency units doesn't load, for example).

Oh I see now that its because I'm getting 500's back from the requests it kicks off when that telemetry form is opened.
image

GrahamS-Quartech:
> Not sure if this is an issue or not, but just in case:
> The warning at the bottom is a bit funny in how its displayed, in particular the "from 2023-02-02T08:00:00.000Z until indefinite."
> 
>     1. Created an animal
> 
>     2. Went to add a device, and just picked some random values.
> 
The errors you are seeing are intentional, you are allowed to use device id's that have already been registered in BCTW here but these are just some checks to make sure that the way you are adding it here makes sense. 

I know it working like this is a bit odd, but we are leaving it as is for now, may come back to it once we can decide on a better design with Jeremy D.

> Also, when I enter a date, and then clear it, the error message is some default yup message. 
> Some of the yup checks probably just need some specific messages set, so they don't fallback to some default.

~~This should probably get updated yeah.~~ Have since updated this.

> Oh I see now that its because I'm getting 500's back from the requests it kicks off when that telemetry form is opened.

~~I think I know what's causing this and I should have a fix out shortly.~~ Should work now.

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -353,6 +374,24 @@ const SurveyAnimals: React.FC = () => {
     refreshCritters();
   };
 
+  const handleRemoveDeployment = async (deployment_id: string) => {
+    try {
+      if (!selectedCritterId) {
+        throw Error('Survey critter ID was set incorrectly.');
```
curtisupshall: I'm not sure if throwing an uncaught exception is appropriate on the frontend - regardless of whether we may ever expect the user to reach this exception or not. Can we use `setPopup` to display this message instead?

NickPhura:
There are a few code smells we should address as well.

For the 2 smells about the tests, I think we just need to add an expects for the error message (to ensure the error that we caught was the one we expected to be thrown).

app/src/features/surveys/view/survey-animals/TelemetryDeviceForm.tsx
```diff
@@ -59,27 +61,63 @@ const AttchmentFormSection = (props: { index: number; deviceMake: string }) => {
...
+  mode,
+  removeAction
 }: {
   index: number;
   deployments: IDeploymentTimespan[];
+  mode: TELEMETRY_DEVICE_FORM_MODE;
+  removeAction: (deployment_id: string) => void;
 }): JSX.Element => {
+  const [openDeleteDialog, setOpenDeleteDialog] = useState(false);
+  const [deploymentToDelete, setDeploymentToDelete] = useState('');
```
curtisupshall: I would use `null` instead of empty string to denote no deletion

app/src/utils/Utils.ts
```diff
@@ -337,6 +346,36 @@ export const pluralize = (quantity: number, word: string, singularSuffix = '', p
   return `${word}${quantity === 1 ? singularSuffix : pluralSuffix}`;
 };
 
+type nullableString = string | null | undefined;
```
curtisupshall: Nit: use Pascal case for types. I would argue that this type is unneeded anyway, because

1. The name doesn't accurately reflect what it is (I wouldn't expect a "nullable string" to be possibly `undefined` - rather I expect it to either be defined or to be `null`), and;
2. Reading `NullableString` is less clear than `string | null` (case in point, `NullableString` is hiding an assumption)

I would either explicitly use `string | null | undefined`, or something like

```javascript
type DateRange = [string, string | null | undefined]
```
which represents exactly two values, the latter of which can be falsy.

curtisupshall (dismissed):

app/src/utils/Utils.ts
GrahamS-Quartech: I just did this to get rid of a code smell. Kind of an unhelpful sonar rule, since I agree with everything you've said. I'll just change it back and forget about the code smell.

curtisupshall (approved):
Code looks good!

app/src/utils/Utils.ts
NickPhura: Yeah this code smell seems a bit dubious. We can just mark it as "Won't Fix"

MacDeluca (approved):
Code looks good.
Manually tested the form, working correctly.

NickPhura:
This doesn't need to hold up this PR, but we might want to address it in the near future:

Editing a telemetry device: 
throws some console warnings about out of range values for the dropdown fields. I'm guessing (correct me if this is wrong) it is a result of the select control options being loaded after the form has already loaded, and so the select controls render with non-empty values, but have no matching options (until a few moments later when they are fetched from BCTW).

image

(The form does re-render and the values get loaded/displayed correctly, but we should try and resolve these kinds of warnings anyways, in general.)

GrahamS-Quartech:
> This doesn't need to hold up this PR, but we might want to address it in future:
> 
> Editing a telemetry device: throws some console warnings about out of range values for the dropdown fields. I'm guessing its a result of the dropdown values being loaded after the form has already loaded, and so the select controls render with non-empty values, but the controls have no matching options (until a few moments later when they are fetched from BCTW).

Yeah, I think this can sometimes pop up in the critter form too. Should fix eventually but doesn't lead to any noticeably wrong behaviour or poor performance at least from my testing.

NickPhura:
nit: snackbars get shown on telemetry add and edit, but not on delete.

NickPhura:
> > This doesn't need to hold up this PR, but we might want to address it in future:
> > Editing a telemetry device: throws some console warnings about out of range values for the dropdown fields. I'm guessing its a result of the dropdown values being loaded after the form has already loaded, and so the select controls render with non-empty values, but the controls have no matching options (until a few moments later when they are fetched from BCTW).
> 
> Yeah, I think this can sometimes pop up in the critter form too. Should fix eventually but doesn't lead to any noticeably wrong behaviour or poor performance at least from my testing.

Yeah things work correctly, we just don't want warnings/console messages to pile up over time. As maintainability goes, a new developer might not know if the warning is real or not, so its best to resolve them and leave the console as clean as possible.  But, this could be saved for a techdebt ticket, or in whatever the next ticket that touches these components is.

PR #1113, opened 2023-10-06
BugFix: Fix Funding Source Migration

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

New funding source name field was shorter than old name field, causing migration to fail. 
Only caught in Test, where some funding sources had names longer than 50 characters.

## Testing Notes

Migration runs successfully.


al-rosenthal (approved):
🙈

KjartanE (approved):
🐈

PR #1115, opened 2023-10-06
SIMSBIOHUB-220, 221 - Show and Edit Sampling Sites

KjartanE:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-220
- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-221

## Description of Changes

## Testing Notes


api/src/repositories/sample-location-repository.ts
```diff
@@ -29,6 +33,24 @@ export type UpdateSampleLocationRecord = Pick<
...
+export type UpdateSampleSitesRecord = {
+  survey_id: number;
+  survey_sample_site_id: number;
+  name: string;
+  description: string;
+  survey_sample_sites: Feature[];
+  methods: UpdateSampleMethodRecord[];
+};
+
+export type UpdateSampleSiteRecord = {
```
curtisupshall, line 36: We may want to consider renaming these in the near future... They're very similarly named and they share a lot of overlap. From a functional perspective, I can't really tell them apart just from reading them.

api/src/paths/project/{projectId}/survey/{surveyId}/sample-site/{surveySampleSiteId}/index.ts
```diff
@@ -81,8 +81,39 @@ PUT.apiDoc = {
...
+                survey_sample_sites: {
+                  type: 'array',
+                  items: {
+                    ...(GeoJSONFeature as object)
+                  }
+                },
+                methods: {
+                  type: 'array',
+                  items: {
+                    type: 'object',
```
NickPhura, line 93: nit: (not just this line) Don't forget to specify which fields at each level are `required`, if any.

Same applies to arrays with `minItems`. We don't normally have minItems, but I think in this case we require at least 1 method and the method requires at least 1 period?

api/src/repositories/sample-method-repository.ts
```diff
@@ -62,6 +63,7 @@ export class SampleMethodRepository extends BaseRepository {
    * @memberof SampleMethodRepository
    */
   async updateSampleMethod(sampleMethod: UpdateSampleMethodRecord): Promise<SampleMethodRecord> {
+    console.log('updateSampleMethod', sampleMethod);
```
curtisupshall: Remove

api/src/services/sample-location-service.ts
```diff
@@ -95,14 +97,47 @@ export class SampleLocationService extends DBService {
...
-  /**
-   * updates a survey Sample Location.
-   *
-   * @param {PostSampleLocation} sampleLocation
-   * @return {*}  {Promise<SampleLocationRecord>}
-   * @memberof SampleLocationService
-   */
-  async updateSampleLocation(sampleLocation: UpdateSampleLocationRecord): Promise<SampleLocationRecord> {
-    return this.sampleLocationRepository.updateSampleLocation(sampleLocation);
+  async updateSampleLocationMethodPeriod(sampleSites: UpdateSampleSitesRecord) {
```
curtisupshall: This method will need some JS doc to explain what it does

api/src/services/sample-location-service.ts
```diff
@@ -95,14 +97,47 @@ export class SampleLocationService extends DBService {
...
+    const sampleSite: UpdateSampleSiteRecord = {
+      survey_sample_site_id: sampleSites.survey_sample_site_id,
+      survey_id: sampleSites.survey_id,
+      name: sampleSites.name,
+      description: sampleSites.description,
+      geojson: sampleSites.survey_sample_sites[0],
+      methods: sampleSites.methods
+    };
+
+    // Update the main sample location
```
curtisupshall, line 108: JS doc should also mention business logic, namely what a **main sample location** is

api/src/services/sample-method-service.ts
```diff
@@ -42,6 +42,15 @@ export class SampleMethodService extends DBService {
    * @memberof SampleMethodService
    */
   async deleteSampleMethodRecord(surveySampleMethodId: number): Promise<SampleMethodRecord> {
+    const samplePeriodService = new SamplePeriodService(this.connection);
+
+    // Delete all associated sample periods
+    const existingSamplePeriods = await samplePeriodService.getSamplePeriodsForSurveyMethodId(surveySampleMethodId);
+    const promises = existingSamplePeriods.map((item) => {
+      return samplePeriodService.deleteSamplePeriodRecord(item.survey_sample_period_id);
```
curtisupshall: We should refactor this method to delete multiple records, so we can delete all of them in one request.

api/src/services/sample-method-service.ts
```diff
@@ -71,6 +80,36 @@ export class SampleMethodService extends DBService {
...
   }
 
+  /**
+   * Checks for sample methods to delete.
+   *
+   * @param {number} surveySampleSiteId
+   * @param {UpdateSampleMethodRecord[]} newMethods
+   * @memberof SampleMethodService
+   */
+  async checkSampleMethodsToDelete(surveySampleSiteId: number, newMethods: UpdateSampleMethodRecord[]) {
```
curtisupshall: I would make this method name a little more verbose; I might think this function just returns an array of sample methods that should be deleted, rather than actually deleting those records. See: `ObservationService.insertUpdateDeleteSurveyObservations`

api/src/services/sample-method-service.ts
```diff
@@ -71,6 +80,36 @@ export class SampleMethodService extends DBService {
...
+    //Get any existing methods for the sample location
+    const existingMethods = await this.getSampleMethodsForSurveySampleSiteId(surveySampleSiteId);
+
+    //Compare input and existing for methods to delete
```
curtisupshall, line 93: I would make these comments a little more specific, or add a description of what's supposed to happen in the JS doc

api/src/services/sample-period-service.ts
```diff
@@ -65,4 +65,34 @@ export class SamplePeriodService extends DBService {
...
   }
+
+  /**
+   * Checks for sample Period to delete.
+   *
+   * @param {number} surveySampleMethodId
+   * @param {UpdateSampleMethodRecord[]} newPeriod
+   * @memberof SamplePeriodService
+   */
+  async checkSamplePeriodToDelete(surveySampleMethodId: number, newPeriod: UpdateSamplePeriodRecord[]) {
```
curtisupshall: Same comments from earlier apply to this class

app/src/utils/mapBoundaryUploadHelpers.ts
```diff
@@ -70,6 +81,16 @@ export const parseShapeFile = async (file: File): Promise<Feature[]> => {
   });
 };
 
+/**
+ * Generates a unique ID for a geoJSON Feature.
+ *
+ * @param {Feature} feature
+ * @return {*}  {string}
+ */
+const generateFeatureId = (feature: Feature): string => {
```
curtisupshall: The JS doc for this util claims it will produce a unique hash, but `object-hash` [defaults to SHA1](https://github.com/puleos/object-hash), which is not unique. Two like objects will hash to the same value:
```javascript
const features = [
  {
    type: 'Feature',
    properties: {},
    geometry: {
      type: 'Point',
      coordinates: [-146.95401365536304, 44.62175409623327]
    }
  },
  {
    type: 'Feature',
    properties: {},
    geometry: {
      type: 'Point',
      coordinates: [-146.95401365536304, 44.62175409623327]
    }
  }
];

console.log(features.map(hash)); // => ['440f53c212ea1...', '440f53c212ea1...']
```

I don't think `object-hash` supports any unique hashing algorithms. You could just `JSON.stringify` the object and then run that through `crypto-js` (the Javascript version of the `crypto` package: https://www.npmjs.com/package/crypto-js). I would either reconsider the relevance of unique IDs for features, or if possibe, just use a UUID.

curtisupshall (commented):
Review Part 1/2

app/src/features/surveys/components/MethodForm.tsx
```diff
@@ -41,17 +41,17 @@ export interface IEditSurveySampleMethodData extends ISurveySampleMethodData {
 }
 
 export const SurveySampleMethodPeriodArrayItemInitialValues = {
-  method_lookup_id: null,
-  survey_sample_period_id: null,
-  survey_sample_method_id: null,
+  method_lookup_id: '' as unknown as null,
```
curtisupshall, line 44: Any reason these can't be `null`?

app/src/features/surveys/components/SamplingMethodForm.tsx
```diff
@@ -122,7 +121,8 @@ const SamplingMethodForm = () => {
               mb: 2,
               maxWidth: '92ch'
             }}>
-            Methods added here will be applied to ALL sampling locations. These can be modified later if required.
+            Methods added here will be applied to ALL sampling locations. These can be modifialValues: initialData ||
```
curtisupshall: Looks like a find/replace gaffe

app/src/features/surveys/observations/sampling-sites/SamplingSiteList.tsx
```diff
@@ -1,181 +1,195 @@
...
-          overflowY: 'scroll',
-          background: grey[50],
-          '& .MuiAccordion-root + .MuiAccordion-root': {
-            borderTopStyle: 'solid',
-            borderTopWidth: '1px',
-            borderTopColor: grey[300]
-          }
-        }}>
-        <Box
+            history.push(`sampling/${selectedSampleSiteId}/edit`);
```
curtisupshall: Nit: avoid `history.push` in `onClick` event handlers and prefer using `component={RouterLink}`.

```javascript
import MenuButton from '@mui/...';
import { Link as RouterLink } from 'react-router-dom';

export default Example = () => {
  return (
    <MenuItem component={RouterLink} to={`sampling/${selectedSampleSiteId}/edit`}>
      ...
    </MenuItem>
  );
}
```

This isn't a major issue, but exposing the link to the browser is a generally a good idea because
1. a11y/semantics;
2. users can preview where the link will take them before they go, and;
3. users get access to the appropriate context menu, so they can do things like _Open in new tab_, etc.

app/src/features/surveys/observations/sampling-sites/SamplingSiteList.tsx
```diff
@@ -1,181 +1,195 @@
...
-              </IconButton>
+            <strong>Sampling Sites</strong>
+          </Typography>
+          <Button
+            sx={{
+              mr: -1
+            }}
+            variant="contained"
+            color="primary"
+            onClick={() => {
```
curtisupshall: see comment re: `history.push`

app/src/features/surveys/observations/sampling-sites/components/SamplingSiteMapControl.tsx
```diff
@@ -120,8 +120,8 @@ const SamplingSiteMapControl = (props: ISamplingSiteMapControlProps) => {
           <Typography component="legend" data-testid="funding-source-list-found">
             Site Location Preview &zwnj;
             <Typography component="span" color="textSecondary" fontWeight="400">
```
curtisupshall: I would conditionally render this `<Typography>` rather than render it with an empty string

app/src/features/surveys/observations/sampling-sites/components/SamplingSiteMapControl.tsx
```diff
@@ -120,8 +120,8 @@ const SamplingSiteMapControl = (props: ISamplingSiteMapControlProps) => {
...
-              {formikProps.values.survey_sample_sites.length > 0
-                ? `(${formikProps.values.survey_sample_sites.length} locations detected)`
```
curtisupshall, line 124: We should probably use the pluralize util function to avoid rendering "1 locations detected"

app/src/features/surveys/observations/sampling-sites/edit/components/SampleMethodEditForm.tsx
```diff
@@ -0,0 +1,224 @@
...
+  }, [codesContext.codesDataLoader]);
+
+  const handleMenuClick = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>, index: number) => {
+    setAnchorEl(event.currentTarget);
+    setEditData({ ...values.sampleSite.methods[index], index });
+  };
+
+  const handleDelete = () => {
+    if (editData) {
+      const data = values.sampleSite.methods;
```
curtisupshall: I would avoid using variable names like `data`. Maybe `const methods`?

app/src/features/surveys/observations/sampling-sites/edit/components/SampleSiteEditForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+                  props.formikRef.current?.submitForm();
+                }}
+                className={classes.actionButton}>
+                Save and Exit
+              </LoadingButton>
+              <Button
+                variant="outlined"
+                color="primary"
+                onClick={() => {
+                  history.push(
```
curtisupshall: Same comment re: `history.push`

app/src/features/surveys/observations/sampling-sites/edit/components/SampleSiteGeneralInformationForm.tsx
```diff
@@ -0,0 +1,31 @@
+import Grid from '@mui/material/Grid';
+import CustomTextField from 'components/fields/CustomTextField';
+import React from 'react';
```
curtisupshall: Nit: remove

app/src/features/surveys/observations/sampling-sites/edit/components/SampleSiteGeneralInformationForm.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Create survey - general information fields
+ *
+ * @return {*}
+ */
+const SampleSiteGeneralInformationForm: React.FC = (props) => {
```
curtisupshall: Nit: remove `React.FC` and remove `props`

app/src/features/surveys/observations/sampling-sites/edit/components/SamplingSiteEditMapControl.tsx
```diff
@@ -0,0 +1,173 @@
...
+ * Sampling site map component.
+ *
+ * @param {ISamplingSiteEditMapControlProps} props
+ * @return {*}
+ */
+const SamplingSiteEditMapControl = (props: ISamplingSiteEditMapControlProps) => {
+  const classes = useStyles();
+  const surveyContext = useContext(SurveyContext);
+  const urlParams: Record<string, string | number | undefined> = useParams();
+  const surveySampleSiteId = Number(urlParams['survey_sample_site_id']);
```
curtisupshall: `Number(undefined)` will produce `NaN`. We could make `surveySampleSiteId` nullable to handle this case?
```javascript
const surveySampleSiteId: number | null = Number(urlParams['survey_sample_site_id']) || null;

File: app/src/features/surveys/observations/sampling-sites/edit/components/SamplingSiteEditMapControl.tsx [hunk 22]
```diff
@@ -0,0 +1,173 @@
...
+            ActionButtonComponent: SampleSiteFileUploadItemActionButton,
+            ProgressBarComponent: SampleSiteFileUploadItemProgressBar
+          }}
+        />
+      </Box>
+      <Box component="fieldset">
+        <Typography component="legend" data-testid="funding-source-list-found">
+          Site Location Preview &zwnj;
+          <Typography component="span" color="textSecondary" fontWeight="400">
+            {samplingSiteGeoJsonFeatures.length > 0 ? `(${samplingSiteGeoJsonFeatures.length} locations detected)` : ''}
```
curtisupshall: Same comment re: conditionally rendering this `<span>`, and using the pluralize util.

curtisupshall (commented):
Review Part 2/2 🐬

api/src/services/sample-method-service.ts
NickPhura: +1

(Applies to all of us) Doesn't cost us anything to be more verbose in our jsdoc/comments, especially if the code does something non-obvious or specific to some business knowledge that some new developer won't necessarily have.

app/src/features/surveys/components/MethodForm.tsx
NickPhura: (Not sure if this is the issue in this specific case) Lots of values, if they are used in form controls, don't like having a null default. I think the best example is in dropdown controls, they throw a warning/error in the console often if the value is null to start with (rather than empty string).

api/src/services/sample-method-service.ts
NickPhura: Yeah, might be good to have 2 functions, one to get the methods to delete and one to do the deletions. The delete could presumably be generic then, and used anywhere records of this type need deleting?

app/src/features/surveys/observations/sampling-sites/SamplingSiteHeader.tsx
curtisupshall, line 1: Any `<Link />` components that use the `href={...}` prop should be replaced with `<Link component={RouterLink} to={...} />`

app/src/utils/mapBoundaryUploadHelpers.ts
NickPhura: This was something Alfred and I were playing with.  

The overall problem is that we need a unique key for each feature when rendering them on the map. 

Setting the ID field in the feature is ideal, since then it will be available for all future rendering needs.

We could certainly use a UUID. The only reasoning for the hash was that it would cover the case where the same exact feature was introduced to the UI again (presumably allowing react to optimize that a bit and not re-render it).  But, maybe that scenario isn't super critical and the simplicity/speed of a UUID is preferred.

Thoughts?

Edit: Actually thinking of it more, the UUID might cover most of the relevant scenarios, and be a lot faster than the hash.

NickPhura: (Technically `JSON.stringify` is not safe to run on an unknown object, as it doesn't correctly handle circular objects (just crashes eventually when it runs out of memory or something). That being said, its not really something people are to concerned with coding around since its such a rare issue, and if your code isn't critical/sensitive, it probably wouldn't matter anyways)

Edit: I might be confusing stringify with some other JSON related function that isn't safe for the above reasons.. Can't remember now... Might be printing the object or something.

api/src/services/sample-method-service.ts
al-rosenthal: on it

app/src/utils/mapBoundaryUploadHelpers.ts
curtisupshall: > The only reasoning for the hash was that it would cover the case where the same exact feature was introduced to the UI again (presumably allowing react to optimize that a bit and not re-render it).

If we want the hash to be the same for two like objects, then `object-hash` would probably be appropriate. The JS doc said "creates a unique hash" which is why I raised the issue of the hashes not being unique.

curtisupshall: Come to think of it, if two records contain the exact same lat/long, then we probably aren't concerned if one of them doesn't render (due to a matching hash), because you'd only really be able to see one of them anyway. Maybe this is a non-issue?

api/src/services/sample-method-service.ts
al-rosenthal: updated

api/src/services/sample-method-service.ts
al-rosenthal: updated the function name and added a bit more to the comments

api/src/services/sample-period-service.ts
al-rosenthal: updated

api/src/services/sample-method-service.ts
al-rosenthal: updated to hopfully make more sense

app/src/utils/mapBoundaryUploadHelpers.ts
curtisupshall: In summary, from discussion: will use UUID instead of object-hash, because the rendering efficiency gains from hashing aren't efficacious enough to justify the cost of computing the hash

al-rosenthal (approved):
running this locally, things look good 🍖

NickPhura (approved):

PR #1118, opened 2023-10-12
SIMSBIOHUB 295 - View Telemetry from Animal Table

GrahamS-Quartech:
## Links to Jira Tickets

- [SIMSBIOHUB-295](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-295)

## Description of Changes

- Added a View Telemetry button to the actions menu for rows in the animal table. Only visible if animal has deployments.
- When you click this, it will either show you a fullscreen map view of the telemetry points and tracks for this animal's deployments or show you a small dialog indicating that there are no points to show yet (most likely scenario for newly added critters).
- Points are grouped by deployment id. The deployment id is used to calculate the colour so it is consistent between re-renders.

## Testing Notes

- When you add a new critter and a new deployment, there isn't going to be any telemetry to show (since the device hasn't started getting new points since 2 seconds ago). If you want to test this locally, just add critterbase_critter_id `826aa7b4-31cf-44d5-80cd-fc159e42d7e7` in your local biohub db instances to see one with a couple active deployments.


api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}/telemetry.ts
```diff
@@ -0,0 +1,308 @@
...
+      schema: {
+        type: 'number'
+      },
+      required: true
+    },
+    {
+      in: 'path',
+      name: 'critterId',
+      schema: {
+        type: 'number'
```
curtisupshall: use `type: 'integer'`

api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}/telemetry.ts
```diff
@@ -0,0 +1,308 @@
...
+    const endDate = String(req.query.endDate);
+    const connection = getDBConnection(req['keycloak_token']);
+    const surveyCritterService = new SurveyCritterService(connection);
+    const bctw = new BctwService(user);
+    try {
+      await connection.open();
+      const surveyCritters = await surveyCritterService.getCrittersInSurvey(surveyId);
+      const thisCritter = surveyCritters.find((a) => a.critter_id === critterId);
+      if (!thisCritter) {
+        throw Error('Specified critter was not part of this survey.');
```
curtisupshall: Should we be throwing HTTP errors here instead? I know the BCTW endpoints work a little differently

api/src/services/bctw-service.ts
```diff
@@ -206,6 +211,7 @@ export class BctwService {
       const params = new URLSearchParams(queryParams);
       url += `?${params.toString()}`;
     }
+    console.log('bctw makeGetRequest to ' + this.axiosInstance.defaults.baseURL + url);
```
curtisupshall: Nit: remove

app/src/features/surveys/view/survey-animals/SurveyAnimalsTable.tsx
```diff
@@ -21,6 +21,7 @@ interface ISurveyAnimalsTableProps {
   onAddDevice: (critter_id: number) => void;
   onEditDevice: (device_id: number) => void;
   onEditCritter: (critter_id: number) => void;
+  onMapOpen: () => void;
```
curtisupshall, line 24: Could make some of these callbacks optional it seems

app/src/features/surveys/view/survey-animals/TelemetryMap.tsx
```diff
@@ -0,0 +1,121 @@
...
+  colourMap: ColourDeployment[];
+}
+
+const Legend = ({ hasData, colourMap }: ILegend) => {
+  return (
+    <div className="leaflet-bottom leaflet-left">
+      <div className="leaflet-control leaflet-bar">
+        <Paper sx={{ padding: 2 }}>
+          {hasData ? (
+            colourMap.map((a, i) => (
```
curtisupshall: Nit: avoid single-letter variables

app/src/utils/Utils.ts
```diff
@@ -397,3 +397,74 @@ export const getCodesName = (
...
+    else [r, g, b] = [c, 0, x];
+
+    return [(r + m) * 255, (g + m) * 255, (b + m) * 255].map((val) => Math.round(val));
+  }
+
+  function RGBToHex(rgb: number[]) {
+    return rgb.map((val) => val.toString(16).padStart(2, '0')).join('');
+  }
+
+  function generateOutlineColor(hsl: HSL) {
```
JeremyQuartech, line 453: Might be worth dropping the outline-color stuff, unless its going to get used at some point.

app/src/features/surveys/view/survey-animals/TelemetryMap.tsx
```diff
@@ -0,0 +1,131 @@
...
+      }
+    }
+    return result;
+  }, [telemetryData, deploymentData]);
+
+  const mapBounds = useMemo(() => {
+    const bounds = new L.LatLngBounds([]);
+    telemetryData?.points.features.forEach((feature) => {
+      if (feature.geometry.type === 'Point') {
+        const lat = feature.geometry.coordinates[1];
```
curtisupshall: Optional for next time:
```javascript
const [lon, lat] = feature.geometry.coordinates;
```

app/src/utils/Utils.ts
GrahamS-Quartech: Yeah good point, I'm not really sure it will ever get used. Can always grab it from BCTW again if we do need it.

app/src/utils/Utils.ts
```diff
@@ -397,3 +397,63 @@ export const getCodesName = (
...
 };
+
+/**
+ * Convert a UUID into an arbitrary color within a constrained color space.
+ *
+ * @param id uuid
+ * @returns {*} {fillColor: string, outlineColor: string}
+ */
+export const uuidToColor = (id: string): string => {
+  const uuidToInt = (uuid: string): number => {
```
curtisupshall, line 408: Could avoid using regex:
```javascript
const intUuid = parseInt(id.split('-')[4]);
```

GrahamS-Quartech: This gives different results which required adjusting the algorithm so not really worth it imo.

app/src/utils/Utils.ts
GrahamS-Quartech: Actually I decided to instead just add back that value and implement an outline for the map points.

app/src/features/surveys/view/survey-animals/TelemetryMap.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <div className="leaflet-control leaflet-bar">
+        <Paper sx={{ padding: 2 }}>
+          {hasData ? (
+            colourMap.map((deploymentAndColour, idx) => (
+              <Box
+                key={deploymentAndColour.deployment_id}
+                display={'flex'}
+                flexDirection={'row'}
+                alignItems={'center'}
+                marginTop={idx > 0 ? 1 : 0} /*<- Any better way to do this? */
```
curtisupshall: There is; Give `sx={{ gap: 1 }}` to the parent element, and it will put a gap between all the child elements (note this is a Flexbox property)

api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}/telemetry.ts
```diff
@@ -0,0 +1,310 @@
...
+            title: 'Telemetry response',
+            type: 'object',
+            properties: {
+              points: {
+                type: 'object',
+                properties: {
+                  type: {
+                    type: 'string'
+                  },
+                  features: {
```
NickPhura: We have GeoJSON openapi definitions in `geoJson.ts` that we should use. Just to ensure the geojson types are consistent across all endpoints. 

The 2 Feature types don't define anything specific in the properties field, so you'd have to assign it manually in this file before adding it to the responses section here.

NickPhura:
I noticed that the call to fetch telemetry map data gets made when the animal context menu is opened, rather then when the map component is loaded. We might want to move that call down a level just to avoid fetching map data when its not necessarily needed.

api/src/paths/project/{projectId}/survey/{surveyId}/critters/{critterId}/telemetry.ts
```diff
@@ -0,0 +1,310 @@
...
+  responses: {
+    200: {
+      description: 'Responds with count of rows created in SIMS DB Deployments.',
+      content: {
+        'application/json': {
+          schema: {
+            title: 'Telemetry response',
+            type: 'object',
+            properties: {
+              points: {
```
NickPhura: If the 200 response will always have `points` and `tracks`, we should mark them as required. Same with any other fields.

GrahamS-Quartech:
> I noticed that the call to fetch telemetry map data gets made when the animal context menu is opened, rather then when the map component is loaded. We might want to move that call down a level just to avoid fetching map data when its not necessarily needed.

It was originally made inside that component but I moved it up because Macgregor and Jeremy wanted to show a dialogue indicating there is no telemetry collected yet if you hit the button and there isn't any.

curtisupshall (dismissed):
Code looks good!

NickPhura:
> > I noticed that the call to fetch telemetry map data gets made when the animal context menu is opened, rather then when the map component is loaded. We might want to move that call down a level just to avoid fetching map data when its not necessarily needed.
> 
> It was originally made inside that component but I moved it up because Macgregor and Jeremy wanted to show a dialogue indicating there is no telemetry collected yet if you hit the button and there isn't any.

Gotcha. I think ideally then we would have an extra layer between the menu and the map, that does the data fetch and dictates if the map or the dialog is rendered. But, I'm not sure how important that is just yet, so can probably leave it as is for now.

JeremyQuartech (approved):
Code looks good. Following the test instructions, everything was working as expected.

NickPhura (approved):

PR #1119, opened 2023-10-13
SIMSBIOHUB-307: Telemetry Device Validation Fixes

JeremyQuartech:
## Links to Jira Tickets

- [SIMSBIOHUB-307](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-307)

## Description of Changes
The BCTW API expects all device_ids to be integer values. These changes help validate the device_ids before any requests are sent off:
- Added new CustomNumberField component
- Added integer restrictions to telemetry device yup schema
- Added conditional logic to date overlap checks to prevent api calls with bad device_id

## Testing Notes
- Go to survey animals section and open the 'Add Device' dialogue on any animal
- Device ID field should be constrained to only positive integers, and will not send requests otherwise (no more random 500 errors from BCTW API).


app/src/components/fields/IntegerField.tsx
```diff
@@ -0,0 +1,72 @@
...
+  name: string;
+  min?: number;
+  max?: number;
+}
+
+const NumberFormatCustom = React.forwardRef<NumberFormatProps, NumberFormatCustomProps>(function NumericFormatCustom(
+  props,
+  ref
+) {
+  const { onChange, min = -2147483648, max = 2147483647, ...other } = props;
```
GrahamS-Quartech: Was gonna comment that these should use the JS max int constant, but I see these values actually align with the Postgres integer type limits. Good call.

app/src/features/surveys/view/survey-animals/device.ts
```diff
@@ -12,7 +12,11 @@ export type ITelemetryPointCollection = { points: FeatureCollection; tracks: Fea
 
 const req = 'Required.';
 const mustBeNum = 'Must be a number';
-const numSchema = yup.number().typeError(mustBeNum);
+const mustBePos = 'Must be positive';
+const mustBeInt = 'Must be an integer';
+const maxInt = 2147483647;
+const numSchema = yup.number().typeError(mustBeNum).min(0, mustBePos);
+export const intSchema = numSchema.max(maxInt, `Must be less than ${maxInt}`).integer(mustBeInt).required(req);
```
GrahamS-Quartech: Nit: if you are going to export this it should probably be made slightly more generic. But it doesn't look like it's used elsewhere so it probably doesn't need to be exported anyways.

GrahamS-Quartech:
I notice that this IntegerField component seems pretty similar to the existing FloatingPointField. Is it possible to refactor this so that they could just be one component configured to constrain either a float or int value? You could probably just have an additional prop like 'numericType' that determines which isAllowed and onValueChange it uses internally. You could even have it default to float if not provided so you don't have to change the usage of this component anywhere else.

app/src/components/fields/IntegerField.tsx
JeremyQuartech: Thats a good point, maybe it still makes sense to use the JS max int as the default. Not really sure which use case would be more common.

GrahamS-Quartech: Personally I think the current value is a fine default. It's already such a high number that there shouldn't really be any reason for our application to go over it and it avoids randomly raising exceptions from the DB.

app/src/components/fields/CustomNumberField.tsx
```diff
@@ -0,0 +1,77 @@
...
+  min?: number;
+  max?: number;
+  float?: boolean;
+}
+
+const NumberFormatCustom = React.forwardRef<NumberFormatProps, NumberFormatCustomProps>(function NumericFormatCustom(
+  props,
+  ref
+) {
+  const { onChange, min, max, float = false, ...other } = props;
```
GrahamS-Quartech: This always overrides the float to false. It's probably better to just spread as is, undefined will be falsy anyways.

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -213,12 +219,14 @@ const SurveyAnimals: React.FC = () => {
           .test('checkDeploymentRangeEnd', '', async (value, context) => {
```
MacDeluca, line 219: move this schema to survey-animals/device.ts.
Might be reused in the future.

app/src/components/fields/CustomNumberField.tsx
```diff
@@ -0,0 +1,77 @@
...
+export type INumberFieldProps = TextFieldProps & {
+  required?: boolean;
+  label: string;
+  name: string;
+  min?: number;
+  max?: number;
+  float?: boolean;
+};
+
+interface NumberFormatCustomProps {
```
MacDeluca: Nit: Can extend off of INumberFieldProps

app/src/features/surveys/view/SurveyAnimals.tsx
GrahamS-Quartech: I don't think it's possible to easily insert this into device.ts, as it is not a react component. The call to deploymentOverlapTest uses the telemetryApi hook, and hooks cannot be used outside of components.

app/src/features/surveys/view/survey-animals/TelemetryDeviceForm.tsx
```diff
@@ -156,7 +157,13 @@ const DeviceFormSection = ({ values, index, mode, removeAction }: IDeviceFormSec
         </Typography>
         <Grid container spacing={3}>
           <Grid item xs={6}>
-            <CustomTextField label="Device ID" name={`${index}.device_id`} other={{ disabled: mode === 'edit' }} />
+            <CustomNumberField
+              label="Device ID"
+              name={`${index}.device_id`}
+              disabled={mode === 'edit'}
+              min={1}
+              max={2147483647}
```
GrahamS-Quartech: nit: I think it could make sense to make this a constant somewhere, something like PG_MAX_INT. Makes it more clear that this max value was chosen for a particular reason.

app/src/features/surveys/view/SurveyAnimals.tsx
JeremyQuartech: I did look at moving these schema into device.ts, and yeah I don't think it would be very straightforward given that it relies on those API hooks

GrahamS-Quartech (approved):

MacDeluca (approved):
Looks gud

PR #1121, opened 2023-10-16
SIMSBIOHUB-293: Create Project Bug

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-293

## Description of Changes

- Add missing authorization check for project permissions.
- Fix funding source endpoint only returning data for system admins (this was causing the funding source dropdown on the survey pages to be empty)
  - Now returns data for any system user, but strips out the 2 survey count/amount fields if the user is a non-admin.
  - Update tests
- Misc cleanup

## Testing Notes

- A non-admin user with the CREATOR system role should be able to create/view/edit a project/survey and every field under them, including funding sources.
- The Funding Sources tab (for admins) should work as expected, as normal.


curtisupshall (dismissed):

curtisupshall (approved):
Code looks good 🌟

al-rosenthal (approved):
ran this locally and it worked as expected

PR #1124, opened 2023-10-17
BugFixes: Misc Fixes

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

- Fix bug in survey view response dates being cast to strings (causing a null end_date to return as the string `'null'`)
  - Steps to reproduce (pre-fix):
    -  Create a project with no end date
    - Create a survey with no end date
    - See that the survey timeline on the view page is empty, because the response contains an end_date of `'null'` (as a string) rather than a proper `null` value.

## Testing Notes

n/a

### DIFF
--- a/api/src/models/survey-view.test.ts
+++ b/api/src/models/survey-view.test.ts
@@ -26,11 +26,11 @@ describe('GetSurveyData', () => {
     });
 
     it('sets end_date', () => {
-      expect(data.end_date).to.equal('undefined');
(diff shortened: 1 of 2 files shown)


curtisupshall (dismissed):
🐞

al-rosenthal (dismissed):
🔨

KjartanE (approved):
🐈

al-rosenthal (approved):
🧈

PR #1127, opened 2023-10-18
SIMSBIOHUB-308: System API User Add Bug

KjartanE:
## Links to Jira Tickets

- https://github.com/bcgov/biohubbc/commit/ae70b16ef5af984c2703fcd62fb4691f771f547f#diff-65716179d0c5e7ab674f79fee978dadb0ce326953bf9535858f42e47ebf26a8a
- https://github.com/bcgov/biohubbc/commit/0106937b88aaec7fddca0065e7e21ad4e71b9d2f#diff-c000a43b79dd47eda57857567a3e3945c163ebebbacf9f170a3da2ffc67c68ef

## Description of Changes

- On PR 194, some work surrounding user auth and adding was accidentally overwritten. This PR is to fix that mistake and update users properly. The overwritten PR is 154

## Testing Notes


api/src/database/db.ts
```diff
@@ -51,11 +60,6 @@ pg.types.setTypeParser(pg.types.builtins.TIMESTAMP, (stringValue: string) => {
 pg.types.setTypeParser(pg.types.builtins.TIMESTAMPTZ, (stringValue: string) => {
   return stringValue; // 1082 for `DATE` type
 });
-// NUMERIC column types return as strings to maintain precision. Converting this to a float so it is usable by the system
-// Explanation of why Numeric returns as a string: https://github.com/brianc/node-postgres/issues/811
-pg.types.setTypeParser(pg.types.builtins.NUMERIC, (stringValue: string) => {
-  return parseFloat(stringValue);
-});
```
NickPhura: This needs to be added back.

NickPhura: (This was added after both of our PRs)

api/src/database/db-utils.ts
```diff
@@ -81,3 +102,59 @@ export const getZodQueryResult = <T extends z.Schema>(zodQueryResultRow: T) =>
...
+      user_guid: keycloakUserInformation.bceid_user_guid,
+      user_identifier: keycloakUserInformation.bceid_username,
+      user_identity_source: SYSTEM_IDENTITY_SOURCE.BCEID_BUSINESS,
+      display_name: keycloakUserInformation.display_name,
+      email: keycloakUserInformation.email,
+      given_name: keycloakUserInformation.given_name,
+      family_name: keycloakUserInformation.family_name,
+      agency: keycloakUserInformation.bceid_business_name
+    };
+  } else {
```
curtisupshall, line 147: Maybe we should make this an else-if branch and chec kthat it's actually a BCeID Basic account? Because if it's not (for whatever reason), then we probably want to just return `null`, right?

curtisupshall: Discussed: Zod schema check asserts that the account is one of four types, so this check is not neccessary.

curtisupshall (approved):
Code looks good. 🛠

NickPhura (approved):

PR #1128, opened 2023-10-19
SIMSBIOHUB-296/313: Import Observations

curtisupshall:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-296
  - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-313
  - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-314

## Description of Changes

- Add's UIs for uploading CSV files to survey observations
- Adds a new database table to track survey observation CSV submissions, `survey_observation_submission`
- Alters the `survey_observation` table to make optional the sampling site, method, and period columns
- Adds service and repo methods for processing observation CSVs

## Testing Notes

You should be able to click the **Import** button on the Observations page and import a CSV containing observations.


api/src/repositories/observation-repository.ts
```diff
@@ -190,4 +209,70 @@ export class ObservationRepository extends BaseRepository {
...
+    const knex = getKnex();
+    const sqlStatement = knex
+      .queryBuilder()
+      .select('*')
+      .from('survey_observation_submission')
+      .where('submission_id', submissionId);
+
+    const response = await this.connection.knex(sqlStatement, ObservationSubmissionRecord);
+
+    // TODO add null check here.
```
KjartanE: Complete this TODO.

api/src/services/observation-service.ts
```diff
@@ -1,12 +1,21 @@
...
   UpdateObservation
 } from '../repositories/observation-repository';
+import { generateS3FileKey, getFileFromS3 } from '../utils/file-utils';
+import { getLogger } from '../utils/logger';
+import { CSVWorkBook } from '../utils/media/csv/csv-file';
+import { MediaFile } from '../utils/media/media-file';
+import { parseS3File } from '../utils/media/media-utils';
 import { DBService } from './db-service';
 
+const defaultLog = getLogger('services/observation-queries');
```
KjartanE: `services/observation-service`

api/src/services/observation-service.ts
```diff
@@ -50,4 +77,97 @@ export class ObservationService extends DBService {
...
+    });
+
+    const insertResult = await this.observationRepository.insertSurveyObservationSubmission(
+      submissionId,
+      key,
+      surveyId,
+      file.originalname
+    );
+
+    return { submission_id: insertResult.submission_id, key };
```
KjartanE, line 128: This may be correct in shorthand for the `key` param just wanted to double check the return type is correct

api/src/repositories/observation-repository.ts
curtisupshall: Done

api/src/services/observation-service.ts
curtisupshall: Good catch! Fixed

api/src/services/observation-service.ts
curtisupshall: This is correct

app/src/features/surveys/observations/ObservationsTable.tsx
JeremyQuartech, line 1: One thing I noticed in this table is that it allows you to leave the species blank, which leads to a 500 response from the API. Maybe this could be prevented here rather than sending a bad request.

curtisupshall: Table validation is outside of the scope of this ticket, I think - but it's definitely something we've talked about implementing in the near future.

JeremyQuartech (approved):
Code looks good. Manually tested with provided csv files, and everything worked as expected.

api/src/services/observation-service.ts
```diff
@@ -50,4 +94,107 @@ export class ObservationService extends DBService {
...
+   * @memberof ObservationService
+   */
+  async getObservationSubmissionById(submissionId: number): Promise<ObservationSubmissionRecord> {
+    return this.observationRepository.getObservationSubmissionById(submissionId);
+  }
+
+  /**
+   * Processes a observation upload submission. This method recieves an ID belonging to an
+   * observation submission, gets the CSV file associated with the submisison, and appends
+   * all of the records in the CSV file to the observations for the survey. If the CSV
```
al-rosenthal, line 145: super nit: *received *submission

al-rosenthal (approved):
Ran this locally and the import works. Not blockers but a few things I noticed:
- If you have over 100 records (page limit) in the table and manually add a new one, it isn't obvious it was added to the second page
- If you have unsaved changes and import it wipes out the unsaved changes. This might be ok, just something I noticed
- If an invalid CSV file it uploaded there isn't any feedback to say it failed

api/src/services/observation-service.ts
curtisupshall: Good catch, fixed.

PR #1136, opened 2023-10-23
SIMSBIOHUB-177: Survey Study Area enhancements 

al-rosenthal:
## Links to Jira Tickets

[SIMSBIOHUB-177](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-177)

## Description of Changes
- Survey Study area can now import files with multiple features
- Drawn features can be created and saved to a survey
- New card for displaying and editing map features
- Region selector updated

## Testing Notes

- Create/ Edit a survey
- Use the import feature to upload a list of features
- Use the draw controls to add new geometry to the map
- Modify the new geometry and save it (using the draw controls)
- Update any of the metadata (below the map)
- Save the survey
- There is no front end component that shows these so you'll need to edit the survey you just modified to verify the changes


api/src/services/survey-service.ts
```diff
@@ -750,7 +750,30 @@ export class SurveyService extends DBService {
...
+    );
+    const deletePromises = deletes.map((item) => this.deleteSurveyLocation(item.survey_location_id));
+
+    const inserts = data.filter((item) => !item.survey_location_id);
+    const insertPromises = inserts.map((item) => this.insertSurveyLocations(surveyId, item));
+
+    const updates = data.filter((item) => item.survey_location_id);
+    const updatePromises = updates.map((item) => this.updateSurveyLocation(item));
+
+    return [...insertPromises, ...updatePromises, ...deletePromises];
```
curtisupshall: I'd argue that wrapping this in `Promise.all` would make a more valuable return type

api/src/services/survey-service.ts
```diff
@@ -750,7 +750,30 @@ export class SurveyService extends DBService {
     await Promise.all(promises);
   }
 
-  async updateSurveyLocation(data: PutSurveyLocationData): Promise<void> {
+  async insertUpdateDeleteSurveyLocation(surveyId: number, data: PostSurveyLocationData[]) {
```
curtisupshall: Needs JSdocs

app/src/components/map/MapContainer.tsx
```diff
@@ -137,23 +137,27 @@ const MapContainer = (props: IMapContainerProps) => {
...
+              env: getFeatureDetails.regions
+                .filter((item) => item.sourceLayer === 'WHSE_ADMIN_BOUNDARIES.EADM_WLAP_REGION_BND_AREA_SVW')
+                .map((item) => item.regionName),
+              wmu: getFeatureDetails.regions
+                .filter((item) => item.sourceLayer === 'WHSE_WILDLIFE_MANAGEMENT.WAA_WILDLIFE_MGMT_UNITS_SVW')
+                .map((item) => item.regionName)
+            });
+          }
+        } catch (error) {
+          console.log(error);
```
curtisupshall: Nit: use `console.error`

app/src/components/map/components/DrawControls2.tsx
curtisupshall, line 1: I might consider giving this a different name. Maybe ~~`DrawControls2ElectricBoogaloo`~~ `EnhancedDrawControls`?

app/src/components/map/components/DrawControls.tsx
```diff
@@ -78,6 +78,9 @@ export interface IDrawControlsProps {
   confirmDeletion?: boolean;
 }
 
+/**
+ * @deprecated Prefer using `DrawControls2` moving forward.
```
curtisupshall: Can we elaborate here on why someone may prefer using one over the other?

app/src/components/map/components/DrawControls2.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * Adds a GeoJson feature to a new layer in the draw controls layer group.
+   *
+   * @memberof IDrawControlsRef
+   */
+  addLayer: (feature: Feature, layerId: (id: number) => void) => void;
+
+  deleteLayer: (layerId: number) => void;
+}
+
+const DrawControls2 = forwardRef<IDrawControlsRef | undefined, IDrawControlsProps>((props, ref) => {
```
curtisupshall, line 66: JSdoc? At least maybe an explanation on why this is needed instead of classic `DrawControl.tsx`

app/src/components/map/components/ImportBoundaryDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { boundaryUploadHelper } from 'utils/mapBoundaryUploadHelpers';
+
+export interface IImportBoundaryDialogProps {
+  isOpen: boolean;
+  onClose: () => void;
+  onSuccess: (features: Feature[]) => void;
+  onFailure: (message: string) => void;
+}
+
+export const ImportBoundaryDialog = (props: IImportBoundaryDialogProps) => {
```
curtisupshall: Nit: prefer default exports for React components

app/src/features/surveys/components/StudyAreaForm.tsx
```diff
@@ -46,25 +49,88 @@ export const SurveyLocationYupSchema = yup.object({
...
+      if (item.leaflet_id) {
+        drawRef.current?.deleteLayer(item.leaflet_id);
+      }
+    });
+
+    // set values
+    setFieldValue('locations', data);
+  };
+
+  const getDialogData = () => {
```
curtisupshall: This should probably be a constant, rather than a callback
```javascript
const dialogInitialValues = useMemo(() => {
  if (currentIndex === undefined) {
    return { name: '', description: '' };
  }

  return {
    name: values.locations[currentIndex]?.name,
    description: values.locations[currentIndex]?.description
  };
}, [values.locations, currentIndex]);
```

app/src/features/surveys/components/locations/StudyAreaMap.tsx
```diff
@@ -0,0 +1,126 @@
...
+            {...drawControls}
+            ref={drawRef}
+            options={{
+              ...drawControls?.options,
+              // Always disable circle, circlemarker and line
+              draw: { ...drawControls?.options?.draw, circle: false, circlemarker: false, polyline: false }
+              // edit: { remove: false }
+            }}
+            onLayerAdd={(event) => {
+              console.log('onLayerAdd', event);
```
curtisupshall: Remove placeholders?

app/src/features/surveys/components/locations/SurveyLocationPage.tsx
```diff
@@ -0,0 +1,72 @@
...
+          width={400}
+          sx={{
+            borderRightStyle: 'solid',
+            borderRightWidth: '1px',
+            borderRightColor: grey[300]
+          }}></Box>
+
+        {/* Map Component */}
+        <Box flex="1 1 auto" overflow="hidden">
+          {/* <SurveyAreaMapControl
```
curtisupshall: Nit: remove

app/src/utils/mapBoundaryUploadHelpers.ts
```diff
@@ -82,44 +82,24 @@ export const parseShapeFile = async (file: File): Promise<Feature[]> => {
...
- * @template T Type of the formikProps (should be auto-determined if the incoming formikProps are properly typed)
- * @param {File} file The file to upload
- * @param {string} name The name of the formik field that the parsed geometry will be saved to
- * @param {FormikContextType<T>} formikProps The formik props
- * @return {*}
- */
-export const handleGPXUpload = async <T>(file: File, name: string, formikProps: FormikContextType<T>) => {
-  const { setFieldValue, setFieldError } = formikProps;
-
+export const handleGPXUpload = async (file: File) => {
```
curtisupshall, line 108: No more JSdoc?

app/src/components/map/MapContainer.tsx
al-rosenthal: updated

api/src/services/survey-service.ts
al-rosenthal: added

api/src/services/survey-service.ts
al-rosenthal: good call

app/src/components/map/components/DrawControls2.tsx
al-rosenthal: the plan is to leave this as is for now and write a tech debt ticket to clean up the old draw controls

app/src/components/map/components/ImportBoundaryDialog.tsx
al-rosenthal: updated

app/src/features/surveys/components/locations/SurveyLocationPage.tsx
al-rosenthal: removed

app/src/utils/mapBoundaryUploadHelpers.ts
al-rosenthal: added

app/src/components/map/components/DrawControls.tsx
al-rosenthal: I've added to the comment on this component, hopefully if makes sense

app/src/features/surveys/components/locations/StudyAreaMap.tsx
al-rosenthal: cleaned up

app/src/features/surveys/components/StudyAreaForm.tsx
al-rosenthal: the useMemo here is a callback? The value also isn't so large that caching will provide much value, I put this into a function to avoid having the logic inline for the dialog.

curtisupshall: (The `useMemo` isn't a callback, but it uses a callback to produce a constant)

Agreeing to your point of not inlining this logic in the React component, what I really meant is that having a function `getDialogData()` that always returns the same object should really just be replaced with a constant object (ideally with a more specific name than `getDialogData`). Here, I think it makes more sense to just instantiate the constant. To that end, it's really best practice to memoize **_any_** variable that is instantiated during component render (not just large objects), hence the `useMemo`.

al-rosenthal: Ooh I see. Ok I've updated this chunk to be inline with best practices

curtisupshall (approved):
Code looks good!

NickPhura (approved):

PR #1139, opened 2023-10-24
TechDebt: Pipeline Improvements

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Fix issue where the git action skip was not working as expected when a PR's source branch was dev, test, or prod.

Removed prod from the list of branches the PR Deploy action runs against.
- We already do PR deploys in the PR to Dev, and Dev to Test

Adjust cpu/memory limits.
- Reduced the request size (especially for memory, which was way too high)
- Increased the limit size in test/prod

Fix bug that was causing the static deploy cleanup step to not delete old image stream tags

## Testing Notes

Build/deployments work as expected.

### DIFF
--- a/.github/workflows/addComments.yml
+++ b/.github/workflows/addComments.yml
@@ -4,6 +4,8 @@ name: Add Comments
 on:
   pull_request:
     types: [opened, ready_for_review]
+    branches-ignore:
+      - prod
 
 jobs:
   addOpenshiftURLComment:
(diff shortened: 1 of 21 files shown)


curtisupshall (dismissed):

curtisupshall (dismissed):

al-rosenthal (dismissed):
🔧

curtisupshall (approved):

al-rosenthal (approved):
🦮

PR #1141, opened 2023-10-25
SIMSBIOHUB-188 Manage Animals

MacDeluca:
## Links to Jira Tickets

-[SIMSBIOHUB-188](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-188)

## Description of Changes

- Overhauled the Edit Animal modal to be a full screen page instead. Accessed through the Manage Animals button on a survey.
- Animals in this view appear listed in the sidebar by their alias. Clicking on one of these items opens an accordion to the various form sections.
- Each form section has been adapted into a list of static cards. You can edit or delete cards from the relevant button icons. You can also add another card from the top of the page. Clicking the add or edit buttons brings up a form modal that will send a request on submission of the form.
- Deleting any card should prompt a confirm delete modal.
- Telemetry devices still work slightly differently than the rest of this page in that device and deployments get added together and deployments can be deleted but devices cannot. It's fairly consistently integrated otherwise.
- Currently all sections hit the bulk endpoint in Critterbase (allows multiple sections to be added at once). This can be modified to single endpoints per each form section if this is the final version of this UI.

## Testing Notes

- A small patch to critterbase was required to fix a bug in the capture delete function. This should be in live dev already.
- Currently some tests are broken but waiting till design / flow is more finalized before fixing


app/src/utils/UtilsJSX.tsx
curtisupshall, line 1: I would move the contents of this file to `Utils.ts` and just rename the file to `Utils.tsx`.

app/src/utils/UtilsJSX.tsx
```diff
@@ -0,0 +1,19 @@
+import { Typography } from '@mui/material';
+import { IDialogContext } from '../contexts/dialogContext';
+
+/**
+ * Simple reusable method to make a snackbar appear with a string of your choice.
+ *
+ * @param message string to show
+ * @param context reference to current DialogContext
+ */
+export const setPopup = (message: string, context: IDialogContext) => {
```
curtisupshall: I would probably rename this to `setSnackbar(message, dialogContext)`, or something to that effect; `setPopup` makes me think of a dialog rather than a snackbar

app/src/features/surveys/view/survey-animals/form-sections/MeasurementAnimalForm.tsx
```diff
@@ -98,23 +86,31 @@ interface MeasurementFormContentProps {
...
+  useEffect(() => {
+    setMeasurement(measurements?.find((m) => m.taxon_measurement_id === taxonMeasurementId));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [measurements]); //Sometimes will not display the correct fields without this useEffect but could have side effects, may need to revisit.
 
   const handleMeasurementTypeChange = (event: SelectChangeEvent<unknown>) => {
     handleChange(event);
     setFieldValue(valueName, '');
     setFieldValue(optionName, '');
     const m = measurements?.find((m) => m.taxon_measurement_id === event.target.value);
```
curtisupshall, line 54: We should avoid single-letter variable names; I would call this `const measurement`

app/src/features/surveys/view/survey-animals/form-sections/MeasurementAnimalForm.tsx
```diff
@@ -98,23 +86,31 @@ interface MeasurementFormContentProps {
...
   const taxonMeasurementId = values.measurements[index].taxon_measurement_id;
+  const [measurement, setMeasurement] = useState<IMeasurementStub | undefined>(
+    measurements?.find((m) => m.taxon_measurement_id === taxonMeasurementId)
+  );
   const isQuantMeasurement = has(measurement, 'unit');
 
-  const tMeasurementIDName = getAnimalFieldName<IAnimalMeasurement>(NAME, 'taxon_measurement_id', index);
-  const valueName = getAnimalFieldName<IAnimalMeasurement>(NAME, 'value', index);
-  const optionName = getAnimalFieldName<IAnimalMeasurement>(NAME, 'qualitative_option_id', index);
+  const tMeasurementIDName = getAnimalFieldName<IAnimalMeasurement>(animalKeyName, 'taxon_measurement_id', index);
```
curtisupshall: avoid abbreviations that obfuscate the purpose of a variable (`tMeasurement` could be "time measurement", "telemetry measurement", ...)

app/src/features/surveys/view/survey-animals/form-sections/MeasurementAnimalForm.tsx
curtisupshall: Since `measurement` is already taken, you could rename that to `const [currentMeasurement, setCurrentMeasurement]` or maybe `const [formMeasurement, setFormMeasurement]`

app/src/features/surveys/components/EditDeleteStubCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const EditDeleteStubCard = ({ header, subHeaderData, onClickEdit, onClickDelete }: EditDeleteStubCardProps) => {
+  const formatSubHeaderString = (subHeaderData: ISubHeaderData) => {
+    const formatArr: string[] = [];
+    const entries = Object.entries(subHeaderData);
+    entries.forEach(([key, value]) => {
+      if (value == null || value === '') {
+        return;
+      }
+      formatArr.push(`${key} • ${value}`);
```
curtisupshall: Nit: prefer [HTML entities](https://www.w3schools.com/html/html_entities.asp) over Unicode.

app/src/features/surveys/components/SurveySectionHeader.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { Link as RouterLink } from 'react-router-dom';
+
+export interface SurveySectionHeaderProps {
+  project_id: number;
+  survey_id: number;
+  survey_name: string;
+  title: string;
+}
+
+const SurveySectionHeader: React.FC<SurveySectionHeaderProps> = (props) => {
```
curtisupshall: Remove `React.FC`

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -510,7 +511,7 @@ const SurveyAnimals: React.FC = () => {
         open={openDeviceDialog}
         dialogLoading={isSubmittingTelemetry}
         component={{
-          element: <TelemetryDeviceForm removeAction={handleRemoveDeployment} mode={telemetryFormMode} />,
+          element: <TelemetryDeviceForm removeAction={handleRemoveDeployment} survey_critter_id={-1} />, // This section will be removed
```
curtisupshall: If there is an existing ticket to track the work of removing this section, please link it :)

app/src/features/surveys/view/survey-animals/animal.ts
```diff
@@ -6,6 +6,8 @@ import { v4 } from 'uuid';
 import { AnyObjectSchema, InferType, reach } from 'yup';
 import { AnimalTelemetryDeviceSchema } from './device';
 
+export const MANAGE_ANIMALS_DEFAULT_URL_PARAM = '0';
```
curtisupshall: Is there a lot of value in tracking this as a constant?

app/src/features/surveys/view/SurveyAnimals.tsx
```diff
@@ -535,14 +536,25 @@ const SurveyAnimals: React.FC = () => {
...
+          component="h2"
+          variant="h4"
+          sx={{
+            flex: '1 1 auto'
+          }}>
+          Marked and Known Animals
+        </Typography>
+        <Button
+          component={RouterLink}
+          to={`animals/${MANAGE_ANIMALS_DEFAULT_URL_PARAM}`}
```
curtisupshall: I think I understand what this code is for... On the `/animals` page, if it receives a selected critter ID of exactly `0`, then the page renders in some default state; and if instead the page receives some actual critter ID, then the user is able to manage details for that particular critter?

I think passing an out-of-bounds critter ID for this purpose is an anti-pattern. Instead of using a path param, e.g. `/animals/{id}`, I would use a query param instead, which would allow it to be optional. E.g., `/animals` and `/animals?cid=12`

app/src/features/surveys/view/survey-animals/AnimalList.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+interface AnimalListProps {
+  isLoading?: boolean;
+  critterData?: IDetailedCritterWithInternalId[];
+  selectedSection: IAnimalSections;
+  onSelectSection: (section: IAnimalSections) => void;
+  onAddButton: () => void;
+}
+
+const AnimalList = ({ selectedSection, onSelectSection, critterData, isLoading, onAddButton }: AnimalListProps) => {
```
curtisupshall: Nit: Avoid destructuring large props lists

MacDeluca: Fixed :)

app/src/features/surveys/components/SurveySectionHeader.tsx
MacDeluca: Fixed 💯

app/src/features/surveys/view/SurveyAnimals.tsx
MacDeluca: I prefer this stratgey, implementing this change now.

app/src/features/surveys/view/survey-animals/AnimalList.tsx
MacDeluca: Fixed

app/src/features/surveys/view/SurveyAnimals.tsx
MacDeluca: @GrahamS-Quartech?

app/src/features/surveys/components/SurveySectionHeader.tsx
MacDeluca: Fixed

app/src/features/surveys/view/SurveyAnimals.tsx
MacDeluca: Fixed

app/src/features/surveys/view/survey-animals/animal.ts
MacDeluca: Removed not needed anymore

app/src/features/surveys/view/survey-animals/form-sections/MeasurementAnimalForm.tsx
MacDeluca: fixed

app/src/features/surveys/view/survey-animals/form-sections/MeasurementAnimalForm.tsx
MacDeluca: fixed

app/src/utils/UtilsJSX.tsx
MacDeluca: Fixedarooni

app/src/utils/UtilsJSX.tsx
MacDeluca: Done!

app/src/features/surveys/view/SurveyAnimals.tsx
GrahamS-Quartech: I'm not sure what the ghost of Graham past meant by this but it's probably appropriate to remove all / most of SurveyAnimals.tsx considering much of it is now copied into the various new components that make up this ticket.

app/src/features/surveys/components/EditDeleteStubCard.tsx
curtisupshall: (If you can't get this to work very easily, just ignore this comment)

MacDeluca: Yea the solution for this is a bit more complex than I think its worth. Just gonna keep it like this for now

app/src/features/surveys/components/EditDeleteStubCard.tsx
```diff
@@ -0,0 +1,68 @@
+import { mdiPencilOutline, mdiTrashCanOutline } from '@mdi/js';
+import Icon from '@mdi/react';
+import { Card, CardHeader, IconButton } from '@mui/material';
+import { grey } from '@mui/material/colors';
+import React from 'react';
+
+export type ISubHeaderData = Record<string, string | number | undefined> | string;
```
curtisupshall: Nit: this type is not an interface, so you can just call this `SubHeaderData`.

I might reconsider accepting an `object | string`-like React prop, since the resulting UI behvaiour varies quite a bit. Two things to consider:

1. Use `type SubHeaderData = { key: string; value?: string | number }`. This way, if I want my header to just display a single string, I can pass `header={{ key: 'My Header }}`.
2. Inline your header rendering method:
```javascript
<CardHeader
  subheader={
    props.header.map((header) => [header.key, header.value].filter(Boolean).join(' • ')).join(' | ')
  }
  ...
/>

Comment (curtisupshall) [reply app/src/features/surveys/components/EditDeleteStubCard.tsx]: If you find this strategy compelling, you may reconsider the property names for `SubHeaderData`, e.g. `title` instead of `key`

Comment (MacDeluca) [reply app/src/features/surveys/components/EditDeleteStubCard.tsx]: Fixed the type.

It might almost be beneficial to only accept strings for the subHeaderData. 
Allow the subheader format to be handled by the parent component for more flexibility.

Just from exploring this solution, the objects passed to subHeaderData become very verbose.

File: app/src/features/surveys/view/survey-animals/TelemetryDeviceForm.tsx [hunk 11]
```diff
@@ -158,12 +175,10 @@ const DeviceFormSection = ({ values, index, mode, removeAction }: IDeviceFormSec
         </Typography>
         <Grid container spacing={3}>
           <Grid item xs={6}>
-            <CustomNumberField
+            <CustomTextField
```
JeremyQuartech: It seems that this field isn't always validating properly. Sometimes when you enter a new value, it doesn't if the device already exists and will allow a submission resulting in 500 error. Another thing that could be fixed is that when editing a device / deployment, this field is enabled, but maybe it should be disabled like with device id.

app/src/features/surveys/view/survey-animals/TelemetryDeviceForm.tsx
```diff
@@ -204,13 +219,13 @@ const DeviceFormSection = ({ values, index, mode, removeAction }: IDeviceFormSec
           <Grid item xs={6}>
             <TelemetrySelectField
```
JeremyQuartech: The keyx / cfg upload functionality for Vectronic and Lotek devices has disappeared when adding a new device, but still shows up on edit

app/src/features/surveys/view/survey-animals/form-sections/MortalityAnimalForm.tsx
JeremyQuartech, line 1: Save button is unresponsive

JeremyQuartech: Seems this is an issue on all but the telemetry form... presumably this is a known issue already

GrahamS-Quartech:
I understand this is still a WIP so apologies if these are well known but I did a quick run through and noticed a few things:

- Formatting / spacing on family modal seems incorrect compared to the other sections.
- Dragging the map around will place the marker, I don't believe it did this previously.
- Doesn't seem like there's any way to place the release marker on the map anymore? Maybe this still needs design work.
- Triggering error messages on the Add Telemetry modal will cause the dialog box to resize horizontally.

MacDeluca:
> * Formatting / spacing on family modal seems incorrect compared to the other sections.

Will look into this 

> * Dragging the map around will place the marker, I don't believe it did this previously.

This has been corrected in the latest changes

> * Doesn't seem like there's any way to place the release marker on the map anymore? Maybe this still needs design work.

This has been updated in the new UI

> * Triggering error messages on the Add Telemetry modal will cause the dialog box to resize horizontally.

I'll look into this, will be part of the overall telemetry section changes coming

GrahamS-Quartech:
Figured I would do another quick run through of this. 
- Seems like clicking on the map to place the pin and resizing the uncertainty does not work. Perhaps this is intentional?
- Clicking save on the animal subsection modals does not appear to fire a POST request right now.
- Mortality has a header for release location.
- Selecting Predation as reason allows you to select a taxon, then if you select a taxon and change the reason, it will grey out the taxon dropdown but not clear it. This behaviour might be fine though, I'm pretty sure the API will ignore the taxon in this case.
- May want to check in with Jeremy about integrating the 'View Family Structure' modal better.
- I think the lack of spacing between the critter ID and the general info card looks a bit odd.
- Telemetry validation seems a bit bugged but I probably just don't have latest fixes for that right now.

app/src/constants/i18n.ts
```diff
@@ -341,8 +341,10 @@ export const SurveyAnimalsI18N = {
     'The species or taxon of the animal. If the species is unknown, select the lowest-ranking known taxon, such as the genus or family.',
```
JeremyQuartech, line 341: Typo: 'Different species may different units and unit names' should read 'Different species *may* have...'

JeremyQuartech (commented):
A couple things I noticed...

- The mortality workflow saving was unreliable. I tried to make a new mortality twice, and didn't get any confirmation from either. Had to refresh page and try again. The first one I added eventually showed up.
- Deleting a mortality event led to a runtime error, however, the deletion does succeed: `ERROR
Cannot read properties of undefined (reading 'projection_mode')`
- The capture workflow took a very long time to validate, but did succeed ultimately
- Deleting a capture event also is also causing the same runtime error as above
- Deleting a measurement causes the measurement dialog to briefly open and ten disappear

GrahamS-Quartech:
- Editing a qualitative measurement and changing it to a different qualitative measurement type works correctly, but changing it to a quantitative measurement does not. I think this is because the resulting quantitative measurement object is placed in the update payload, but it probably needs to go in the create payload instead.
- I'm not sure if we should allow creation of multiple mortality events. This used to be disallowed when it was one big form.
- It seems like the validation for telemetry devices may still be firing unreliably. See screenshot, it's verifying against device id 1 even though I entered 456.
image
- I was able to save and edit all other sections successfully.

app/src/constants/i18n.ts
MacDeluca: fixed

MacDeluca:
> A couple things I noticed...
> 
> * The mortality workflow saving was unreliable. I tried to make a new mortality twice, and didn't get any confirmation from either. Had to refresh page and try again. The first one I added eventually showed up.
> * Deleting a mortality event led to a runtime error, however, the deletion does succeed: `ERROR Cannot read properties of undefined (reading 'projection_mode')`
> * The capture workflow took a very long time to validate, but did succeed ultimately
> * Deleting a capture event also is also causing the same runtime error as above
> * Deleting a measurement causes the measurement dialog to briefly open and ten disappear

Found the issue for this, was a bad piece of state inside the dialog modal. It was keeping it open while the form was submitting, and subsequently while the form was deleting entries.

I think this fix solves all these issues.

MacDeluca:
> * Editing a qualitative measurement and changing it to a different qualitative measurement type works correctly, but changing it to a quantitative measurement does not. I think this is because the resulting quantitative measurement object is placed in the update payload, but it probably needs to go in the create payload instead.
> * I'm not sure if we should allow creation of multiple mortality events. This used to be disallowed when it was one big form.
> * It seems like the validation for telemetry devices may still be firing unreliably. See screenshot, it's verifying against device id 1 even though I entered 456.
>   image
> * I was able to save and edit all other sections successfully.

Fixed :)

curtisupshall (commented):
I added a **Do Not Merge** flag to this PR, since you need to change your base branch from `dev-staging` to `dev` (and resolve the subsequent merge conflicts).

PR #1146, opened 2023-10-26
SIMSBIOHUB 347/341 - Names from shapefiles

GrahamS-Quartech:
## Links to Jira Tickets

- [SIMSBIOHUB-347](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-347)
- [SIMSBIOHUB-341](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-341)

## Description of Changes

- Added some additional logic to the API to extract a name and description from the provided shape file when creating sampling sites.
- It will search the keys of the geometry's properties for "name" or "label" and "desc" or "description" (exact match, should we allow substring?), then use the value for the sampling site name and description respectively. 
- Added a character limit to the sampling site name field in the edit form. 

## Testing Notes

- `OminecaShapefile.zip`'s properties are an example of something that won't be picked up by this approach.
- `62_Features_Shapefile.zip` properties will be used by this approach.

### DIFF
--- a/api/src/repositories/sample-location-repository.ts
+++ b/api/src/repositories/sample-location-repository.ts
@@ -152,6 +152,10 @@ export class SampleLocationRepository extends BaseRepository {
    * @memberof SampleLocationRepository
    */
   async insertSampleLocation(sample: InsertSampleLocationRecord): Promise<SampleLocationRecord> {
+    const name = sample.name
+      ? SQL`${sample.name}`
+      : SQL`(SELECT concat('Sample Site ', (SELECT count(survey_sample_site_id) + 1 FROM survey_sample_site sss WHERE survey_id = ${sample.survey_id})))`;
+
     const sqlStatement = SQL`
     INSERT INTO survey_sample_site (
       survey_id,
@@ -160,11 +164,10 @@ export class SampleLocationRepository extends BaseRepository {
       geojson,
       geography
     ) VALUES (
-      ${sample.survey_id},
-      (SELECT concat('Sample Site ', (SELECT count(survey_sample_site_id) + 1 FROM survey_sample_site sss WHERE survey_id = ${sample.survey_id}))),
-      ${sample.description},
-      ${sample.geojson},
-      `;
+      ${sample.survey_id},`.append(name).append(SQL`,
+        ${sample.description},
+        ${sample.geojson},
+        `);
     const geometryCollectionSQL = generateGeometryCollectionSQL(sample.geojson);
 
     sqlStatement.append(SQL`
--- a/api/src/repositories/sample-location-repository.ts
+++ b/api/src/repositories/sample-location-repository.ts
@@ -25,7 +25,9 @@ export const SampleLocationRecord = z.object({
 export type SampleLocationRecord = z.infer<typeof SampleLocationRecord>;
 
 // Insert Object for Sample Locations
-export type InsertSampleLocationRecord = Pick<SampleLocationRecord, 'survey_id' | 'name' | 'description' | 'geojson'>;
+export type InsertSampleLocationRecord = Pick<SampleLocationRecord, 'survey_id' | 'description' | 'geojson'> & {
+  name: string | undefined;
+};
 
 // Update Object for Sample Locations
 export type UpdateSampleLocationRecord = Pick<
@@ -152,7 +154,7 @@ export class SampleLocationRepository extends BaseRepository {
    * @memberof SampleLocationRepository
    */
   async insertSampleLocation(sample: InsertSampleLocationRecord): Promise<SampleLocationRecord> {
-    const name = sample.name
+    const shapeNameOrQuery = sample.name
       ? SQL`${sample.name}`
       : SQL`(SELECT concat('Sample Site ', (SELECT count(survey_sample_site_id) + 1 FROM survey_sample_site sss WHERE survey_id = ${sample.survey_id})))`;
 
@@ -164,7 +166,7 @@ export class SampleLocationRepository extends BaseRepository {
       geojson,
       geography
     ) VALUES (
-      ${sample.survey_id},`.append(name).append(SQL`,
+      ${sample.survey_id},`.append(shapeNameOrQuery).append(SQL`,
         ${sample.description},
         ${sample.geojson},
         `);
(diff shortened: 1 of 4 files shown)


api/src/repositories/sample-location-repository.ts
```diff
@@ -152,6 +152,10 @@ export class SampleLocationRepository extends BaseRepository {
    * @memberof SampleLocationRepository
    */
   async insertSampleLocation(sample: InsertSampleLocationRecord): Promise<SampleLocationRecord> {
+    const name = sample.name
```
curtisupshall: I'd maybe name this `const nameQuery`

api/src/services/sample-location-service.ts
```diff
@@ -64,13 +64,24 @@ export class SampleLocationService extends DBService {
    */
   async insertSampleLocations(sampleLocations: PostSampleLocations): Promise<SampleLocationRecord[]> {
     const methodService = new SampleMethodService(this.connection);
-
+    const determineName = (geometry: Feature<Geometry | GeometryCollection, Properties>) => {
```
curtisupshall: Nit: Add return type and JSdoc (or add it to the parent method) explaining business logic/how the name is determined

api/src/services/sample-location-service.ts
```diff
@@ -64,13 +64,24 @@ export class SampleLocationService extends DBService {
...
+      const nameKey = Object.keys(geometry.properties ?? {}).find(
+        (key) => key.toLowerCase() === 'name' || key.toLowerCase() === 'label'
+      );
+      return nameKey && geometry.properties ? geometry.properties[nameKey].substring(0, 50) : '';
+    };
+    const determineDesc = (geometry: Feature<Geometry | GeometryCollection, Properties>) => {
+      const descKey = Object.keys(geometry.properties ?? {}).find(
+        (key) => key.toLowerCase() === 'desc' || key.toLowerCase() === 'description'
+      );
+      return descKey && geometry.properties ? geometry.properties[descKey].substring(0, 250) : '';
```
curtisupshall: If I understand this right: `determineName` and `determineDesc` will try to get the names and descriptions from the features themselves. If these properties don't exist, then empty string is returned. Then when they are being inserted in the database, Postgres will coerce to the default value (Sample Site #n). Do I have that right?

GrahamS-Quartech: Yeah, that's right. Perhaps all that logic should be in the same place, since I wound up getting pretty confused at first about why the name field was being overridden from what I had already set here. I could make a separate query function in the repository file to call in here, I imagine it's not especially inefficient compared to leaving that coercion as a subquery.

al-rosenthal (approved):
ran this locally and it works as expected

api/src/services/sample-location-service.ts
```diff
@@ -64,13 +64,24 @@ export class SampleLocationService extends DBService {
...
+    };
     // Create a sample location for each feature found
-    const promises = sampleLocations.survey_sample_sites.map((item, index) => {
+    const promises = sampleLocations.survey_sample_sites.map((item) => {
       const sampleLocation = {
         survey_id: sampleLocations.survey_id,
-        name: `Sample Site ${index + 1}`, // Business requirement to default the names to Sample Site # on creation
-        description: sampleLocations.description,
+        name: determineName(item), // Please note that additional logic for this also happens in insertSampleLocation below
```
curtisupshall: This comment isn't very helpful. What is the additional logic? And I don't think `insertSampleLocation` is below? (Could be wrong)

api/src/services/sample-location-service.ts
curtisupshall: Your current approach isn't bad. But I would remove the `determineName` function (since it's just called immedately after definition) to something like
```javascript
const shapeFileFeatureName: string | undefined = ...
```

This way, we're explicitly declaring that the shapefile feature might not have a name, so when we coerce it later, it makes more sense.

app/src/features/surveys/observations/sampling-sites/edit/components/SampleSiteEditForm.tsx
```diff
@@ -33,7 +33,7 @@ export interface ISampleSiteEditFormProps {
 
 export const samplingSiteYupSchema = yup.object({
   sampleSite: yup.object({
-    name: yup.string().default(''),
+    name: yup.string().default('').max(50),
```
NickPhura: Can you specify a message for the max check (not sure what the default message is, but probably something gross) `'Maximum 50 characters'`

api/src/services/sample-location-service.ts
```diff
@@ -64,13 +64,24 @@ export class SampleLocationService extends DBService {
...
-
+    const determineName = (geometry: Feature<Geometry | GeometryCollection, Properties>) => {
+      const nameKey = Object.keys(geometry.properties ?? {}).find(
+        (key) => key.toLowerCase() === 'name' || key.toLowerCase() === 'label'
+      );
+      return nameKey && geometry.properties ? geometry.properties[nameKey].substring(0, 50) : '';
+    };
+    const determineDesc = (geometry: Feature<Geometry | GeometryCollection, Properties>) => {
+      const descKey = Object.keys(geometry.properties ?? {}).find(
+        (key) => key.toLowerCase() === 'desc' || key.toLowerCase() === 'description'
```
NickPhura: Shapefiles can't have property names longer than 10 characters, so it will never be `'description'`.

Some random internet sites seem to think `'des'` and `'descr'` are other common shorthand versions, so could add those if you want. Ultimately the bios will just have to be told what to use tho, so... :/

api/src/services/sample-location-service.ts
GrahamS-Quartech: I can make this comment more precise but insertSampleLocation is called on line 88.

api/src/services/sample-location-service.ts
GrahamS-Quartech: I added those for good measure. Though it's as you say, ultimately we can only provide so many options.

JeremyQuartech (approved):
Code looks good, tested changes locally and everything worked as described 👍

PR #1149, opened 2023-10-27
SIMSBIOHUB-353/325: Survey form updates

al-rosenthal:
## Links to Jira Tickets

[SIMSBIOHUB-325](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-325)
[SIMSBIOHUB-353](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-353)

## Description of Changes

- Site selection strategies is now required for surveys 
- Removed field method from surveys

## Testing Notes

- You should be able to create and edit new and existing surveys without issue


api/src/paths/project/{projectId}/survey/create.ts
```diff
@@ -202,7 +202,8 @@ POST.apiDoc = {
                   type: 'string'
                 },
                 field_method_id: {
```
curtisupshall: It's probably in our best interest to remove and field mthod code entirely, and not just allow it to be null. Thoughts?

JeremyQuartech: Yeah I agree, I think it could cause some confusing leaving this in here.

JeremyQuartech (commented):
It might make sense to update the db seed file, adding a value for site selection strategy, seeing as it should be required, and removing the value for field method id. A deprecation comment in the db column for field_method_id could be useful too.

api/src/paths/project/{projectId}/survey/create.ts
al-rosenthal: good call, removed field method from the open api specs

al-rosenthal:
> It might make sense to update the db seed file, adding a value for site selection strategy, seeing as it should be required, and removing the value for field method id. A deprecation comment in the db column for field_method_id could be useful too.

Good idea. I've removed the field method from the seed file and added the site strategy to the survey seed process

JeremyQuartech (approved):
Tested locally, everything looks good to me 👍

api/src/paths/project/{projectId}/survey/create.ts
curtisupshall: > good call, removed field method from the open api specs

What I mean is, we should probably remove any and all mention of `field_method` from the codebase. Example:

image

al-rosenthal: I have since removed field method from the project 😅

curtisupshall (approved):
Code looks good to me. Tested this locally and everything works as expected. 👍

PR #1151, opened 2023-10-27
SIMSBIOHUB-331: Keycloak Library Updates

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-331

## Description of Changes

Swap out old keycloak libraries for newer more up to date libraries. 

Remove login and logout pages.

Refactor the keycloakWrapper, and update code/tests as necessary.

## Testing Notes

Anything to do with logging in/out works as before.
- Logging in as an IDIR/BCEID user
- Logging in without access, requesting access, being granted access, etc
- Logging in with a non-admin system role and only being able to access projects you have a project level role on


app/src/components/security/RouteGuards.tsx
```diff
@@ -50,240 +52,145 @@ export interface IProjectRoleRouteGuardProps extends RouteProps {
...
       // The user has a pending access request, restrict them to the request-submitted or logout pages
-      if (location.pathname !== '/request-submitted' && location.pathname !== '/logout') {
+      if (location.pathname !== '/request-submitted') {
         return <Redirect to="/request-submitted" />;
       }
     } else {
-      // The user does not have a pending access request, restrict them to the access-request, request-submitted or logout pages
-      if (!['/access-request', '/request-submitted', '/logout'].includes(location.pathname)) {
+      // The user does not have a pending access request, restrict them to the access-request or request-submitted pages
+      if (!['/access-request', '/request-submitted'].includes(location.pathname)) {
```
curtisupshall: Might be nice to factor these into something like
```javascript
const UNAUTHENTICATED_ROUTES = [
  '/access-request',
  '/request-submitted'
]
```

app/src/contexts/useAuthStateContext.tsx
curtisupshall, line 1: Shouldn't this live in `app/src/hooks` (alongside other hooks that provide contexts, similar to this function)?

curtisupshall (commented):
Great refactor! 👏

app/src/contexts/useAuthStateContext.tsx
NickPhura: All of the contexts we have live in `app/src/contexts` currently.

NickPhura: Ah sorry, was reading this as `authStateContext.tsxt`, not the hook that provides it. Yeah, this should probably move.

NickPhura: Well, now that I say that, I don't know if we have other hooks like this one. We've always just called `useContext(<name>)` before. So this might be the first one we have. So we can probably decide where it should live. I don't have any strong feelings either way.

app/src/components/security/RouteGuards.tsx
NickPhura: My only thought with this is that there are other components where the unauth routes are mentioned, so we'd have to come up with a more generalized solution. But then it would feel weird, I think, to have some routes in constants, and others not. But then that gets messy as some of the routes have path variables. So I'm not sure what the best course of action is :/

I guess I'd be inclined to leave it for now, pending some general overall routing enhancement.

app/src/contexts/useAuthStateContext.tsx
NickPhura: I guess I'll move it just to be consistent with other hooks.

curtisupshall: I realize it's a bit of a unique case, yeah

al-rosenthal (commented):
running this locally I was not able to approve an access request

al-rosenthal (approved):
🚜

JeremyQuartech (commented):
Something I noticed while testing out permissions is that updating my own permission from admin to creator caused the app to crash as well as an infinite refresh loop. I think this is due to not having permission to the endpoint after the request has been processed, and maybe a non-issue.

Update: looks like without these new changes it still causes an app crash, but not a infinite reload loop.

curtisupshall (commented):
Code looks good; Testing locally...

curtisupshall (commented):
Logging out as an IDIR and then logging back in as a BCeID doesn't appear to work:

image

curtisupshall (commented):
Posting to `administrative-activity` fails, ostensibly due to the email notification failing. Maybe I'm just missing some GC Notify config in my .env? (Running this on localhost)

image

NickPhura:
> Logging out as an IDIR and then logging back in as a BCeID doesn't appear to work:
> 
> image

As discussed, Pushed up a commit to re-add the siteminder logout redirect. Seems to fix the problem.

JeremyQuartech (approved):
Tested locally and everything looks good!

PR #1152, opened 2023-10-27
SIMSBIOHUB 324 - Multiple Ecological Variables (Intended Outcomes)

GrahamS-Quartech:
## Links to Jira Tickets

- [SIMSBIOHUB-324](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-324)

## Description of Changes

- Added a migration that removes the intended_outcome_id column from the survey table and shifts the responsibility onto a join table called survey_intended_outcome. This makes it possible to have multiple intended outcomes associated with a single survey.
- Modified the frontend and API to support this change. The dropdown in the add and edit forms for the survey have been replaced with a multiselect. Multiselect will still display the subtext associated with these elements. 
- In various cases where a single intended outcome was displayed before, they display as a comma deliminated list instead.
- These items now display with the label of "Ecological Variables" in the frontend survey components.


app/src/features/surveys/components/PurposeAndMethodologyForm.tsx
```diff
@@ -33,7 +34,7 @@ export const PurposeAndMethodologyYupSchema = yup.object().shape({
   purpose_and_methodology: yup.object().shape({
     field_method_id: yup.number().required('Field Method is Required'),
     additional_details: yup.string(),
-    intended_outcome_id: yup.number().required('Intended Outcome is Required'),
+    intended_outcome_ids: yup.array().required('Intended Outcome is Required'),
```
JeremyQuartech: It might make sense to add a minimum of at least one array element here. Currently it appears that the field is required, but it will validate and create the survey just fine with out any selected.

api/src/paths/project/{projectId}/survey/create.ts
```diff
@@ -195,8 +195,11 @@ POST.apiDoc = {
             purpose_and_methodology: {
               type: 'object',
               properties: {
-                intended_outcome_id: {
-                  type: 'number'
+                intended_outcome_ids: {
```
al-rosenthal, line 198: this should probably have a `minItems` like this:
```js
{
  type: 'array',
  minItems: 1,
  items: {
    type: 'integer'
  }
}
```

api/src/paths/project/{projectId}/survey/{surveyId}/update.ts
```diff
@@ -211,16 +211,19 @@ PUT.apiDoc = {
...
                 'additional_details',
                 'field_method_id',
                 'vantage_code_ids',
                 'ecological_season_id',
                 'revision_count'
               ],
               properties: {
-                intended_outcome_id: {
-                  type: 'number'
+                intended_outcome_ids: {
```
al-rosenthal, line 215: this probably needs `minItems` like the create endpoint

api/src/repositories/survey-repository.ts
```diff
@@ -739,6 +739,63 @@ export class SurveyRepository extends BaseRepository {
...
+    return;
+  }
+
+  /**
+   * Insert new rows that weren't found in the existing list. Delete rows that existed before but weren't found in the new list.
+   * @param {number[]} intendedOutcomeIds
+   * @param {number[]} existingOutcomeIds
+   * @param {number} surveyId
+   */
+  async updateSurveyIntendedOutcomes(
```
al-rosenthal: I think the decision making aspect of this code should live in the survey service. We try and keep the repos as simple as we can. A good example of this pattern is in the `insertUpdateDeleteSurveyLocation` function in the survey service

app/src/features/surveys/components/PurposeAndMethodologyForm.tsx
```diff
@@ -60,10 +61,10 @@ const PurposeAndMethodologyForm: React.FC<IPurposeAndMethodologyFormProps> = (pr
         </Typography>
         <Grid container spacing={3}>
           <Grid item xs={12}>
-            <SelectWithSubtextField
-              id="intended_outcome_id"
-              name="purpose_and_methodology.intended_outcome_id"
-              label="Intended Outcome"
+            <MultiAutocompleteField
+              //id="intended_outcome_id"
```
al-rosenthal: nit: could remove this

al-rosenthal (commented):
Running this locally I found some issues: 
- I can create a survey with no eco variables selected
- If I edit a survey with eco variables and remove any of them and save, I get an error from the api

GrahamS-Quartech:
> Running this locally I found some issues:
> 
>     * I can create a survey with no eco variables selected
> 
>     * If I edit a survey with eco variables and remove any of them and save, I get an error from the api

These should be resolved now.

JeremyQuartech (approved):
Tested locally and everything is working great. Maybe worth bumping up some of the API test coverage first before we merge.

JeremyQuartech (approved):

al-rosenthal (approved):
tested locally and works 🎃

PR #1154, opened 2023-10-27
SIMSBIOHUB-348: Delete single or many observation records

curtisupshall:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-348

## Description of Changes

- Adds a new `ObservationTableContext`
- Adds an endpont and repository methods to delete one or more observation records
- Adds `loading` button props to the `YesNoDialog`
- Refactors the observation context to support deleting one or more observation records

## Testing Notes

You should be able to:
 - Create observation records
 - Delete single observation records (using the trash can icon)
 - Delete single observation records (by selecting the row by checkbox, opening the context menu, and clicking **Delete Observation**
 - Delete multiple observation records (by selecting the rows by checkbox, opening the context menu, and clicking **Delete Observations**


api/src/paths/project/{projectId}/survey/{surveyId}/observations/delete.ts
```diff
@@ -0,0 +1,133 @@
...
+    defaultLog.debug({ label: 'deleteSurveyObservations', surveyId });
+
+    const connection = getDBConnection(req['keycloak_token']);
+
+    try {
+      await connection.open();
+
+      const observationService = new ObservationService(connection);
+
+      const deleteObservationIds = req.body?.surveyObservationIds?.map((observationId: string | number) => Number(observationId)) ?? [];
```
MacDeluca: Should this throw an error if no observationIds provided in the payload?

curtisupshall: The OpenAPI schema already asserts that there be at least one observation ID in the request body.

GrahamS-Quartech:
Tested this locally and deletion through either method seems to work correctly. 
Only weird thing I found is that if you try to delete a record that you haven't committed through the save changes button, it gives you an error pop up.

app/src/contexts/observationsContext.tsx
```diff
@@ -135,9 +161,91 @@ export const ObservationsContextProvider = (props: PropsWithChildren<Record<neve
...
+      });
+
+      const deletingObservationIds = observationRecords
+        .filter((observationRecords) => 'survey_observation_id' in observationRecords)
+        .map((observationRecords) => (observationRecords as IObservationRecord).survey_observation_id);
+
+      return biohubApi.observation
+        .deleteObservationRecords(projectId, surveyId, deletingObservationIds)
+        .then(() => {
+          console.log('.then()');
```
GrahamS-Quartech: nit: remove

al-rosenthal:
Noticed a few things testing this:
- A strange UI issue with the controls at the top of the table. The buttons are mismatched sizes and there is a strange bar cutting through the time column if scrolling to the right of the table
Screenshot 2023-11-01 144235


- When deleting a record the observations count at the top of the table
- Ran into an issue deleting records
  -  Import records
  - Edit any record
  - Delete the same record
  - Save changes
- Not an error but it is a strange interaction to create a new record and try and delete that record before saving any changes. The UI throws an error but I think it should just remove the record

al-rosenthal (commented):
Noticed a few things testing this:
- A strange UI issue with the controls at the top of the table. The buttons are mismatched sizes and there is a strange bar cutting through the time column if scrolling to the right of the table
Screenshot 2023-11-01 144235


- When deleting a record the observations count at the top of the table
- Ran into an issue deleting records
  -  Import records
  - Edit any record
  - Delete the same record
  - Save changes
- Not an error but it is a strange interaction to create a new record and try and delete that record before saving any changes. The UI throws an error but I think it should just remove the record

app/src/contexts/observationsContext.tsx
curtisupshall: Removed

curtisupshall:
> Tested this locally and deletion through either method seems to work correctly. Only weird thing I found is that if you try to delete a record that you haven't committed through the save changes button, it gives you an error pop up.

Fixed!

GrahamS-Quartech:
Adding a record and then immediately deleting it will crash the app with the following:
```
The data grid component requires all rows to have a unique `id` property.
Alternatively, you can use the `getRowId` prop to specify a custom id for each row.
A row was provided without id when calling updateRows():
```

NickPhura (approved):

GrahamS-Quartech:
Went through this again and did not encounter any more issues 👍

GrahamS-Quartech (approved):

al-rosenthal:
the count at the top of the table still doesn't update when deleting a record. Probably out of the scope of the ticket but I also still see the line through the time column when scrolling to the table to the right

PR #1157, opened 2023-11-01
SIMSBIOHUB-354-355: Study area and Sampling sites on observations map

al-rosenthal:
## Links to Jira Tickets

[SIMSBIOHUB-354](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-354)
[SIMSBIOHUB-355](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-355)

## Description of Changes

- Added sampling site locations to the observation map
- Added study area locations to the observation map

## Testing Notes

- Create/ Edit a survey
- Add a sampling site 
- Navigate to survey details page and see features on the map
- Toggle layers with map control in the bottom right of the map

### DIFF
--- a/app/public/assets/icon/circle-medium.svg
+++ b/app/public/assets/icon/circle-medium.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="#3388ff"><path d="M12,8A4,4 0 0,0 8,12A4,4 0 0,0 12,16A4,4 0 0,0 16,12A4,4 0 0,0 12,8Z" /></svg>
--- a/app/src/components/map/components/DrawControls.tsx
+++ b/app/src/components/map/components/DrawControls.tsx
@@ -93,12 +93,12 @@ const DrawControls = forwardRef<IDrawControlsRef | undefined, IDrawControlsProps
     const CustomMarker = L.Icon.extend({
       // The preview icon rendered when you are in the process of adding a marker to the map
       options: {
-        shadowUrl: null,
-        iconSize: new L.Point(36, 36),
-        iconAnchor: new L.Point(18, 18),
         iconUrl: 'assets/icon/circle-medium.svg',
         iconRetinaUrl: 'assets/icon/circle-medium.svg',
-        popupAnchor: [20, 20]
+        iconSize: new L.Point(36, 36),
+        iconAnchor: new L.Point(18, 18),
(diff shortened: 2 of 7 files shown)


al-rosenthal:
Noticed a few things testing this:
- A strange UI issue with the controls at the top of the table. The buttons are mismatched sizes and there is a strange bar cutting through the time column if scrolling to the right of the table
Screenshot 2023-11-01 144235


- When deleting a record the observations count at the top of the table
- Ran into an issue deleting records
  -  Import records
  - Edit any record
  - Delete the same record
  - Save changes
- Not an error but it is a strange interaction to create a new record and try and delete that record before saving any changes. The UI throws an error but I think it should just remove the record

NickPhura (approved):

GrahamS-Quartech:
Tested this on the deployed pod. Seems to work as expected.
Perhaps outside the scope of this ticket, but I think it would be good if there was some way to see a label of which feature is which when clicking on one. It's not really clear which is a study area and which is a sampling site since everything is blue and they can overlap. Similarly, might be nice if you could toggle on/off individual elements rather than all sampling sites and all study areas.

GrahamS-Quartech (approved):

al-rosenthal:
> Tested this on the deployed pod. Seems to work as expected. Perhaps outside the scope of this ticket, but I think it would be good if there was some way to see a label of which feature is which when clicking on one. It's not really clear which is a study area and which is a sampling site since everything is blue and they can overlap. Similarly, might be nice if you could toggle on/off individual elements rather than all sampling sites and all study areas.

I think we'll be keeping this pretty simple for now and make this interaction a little more complex as we get a better idea of how users want to interact with the map. I did add an on over for the study areas and sample sites so hopefully it is a little less confusing which is which when toggling. The observations are also now an orange color so they stand out a bit more

GrahamS-Quartech:
> I think we'll be keeping this pretty simple for now and make this interaction a little more complex as we get a better idea of how users want to interact with the map. I did add an on over for the study areas and sample sites so hopefully it is a little less confusing which is which when toggling. The observations are also now an orange color so they stand out a bit more

Tested this again locally. I think the onOver and color changes are good, definitely sufficient for now.

One thing I noticed: It seems that when you refresh the page, the sample site can load in after the study area, so the map bounds will not be fit to both shapes.

curtisupshall (approved):
Code looks good

NickPhura (approved):

PR #1158, opened 2023-11-01
SIMSBIOHUB-356: Update survey list endpoint to return a subset of survey fields

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-356

#### Replace existing survey/list endpoint with survey/index.
- Add service/repo function to fetch a smaller subset of survey fields for a given project id.
- Call new service function

#### Update useSurveyApi
- Calls survey/index endpoint
- Update related interface to match new endpoint response
#### Update survey list component
- No longer renders a column for survey purpose. There is some work being done to modify this data (remove fields, make others arrays, etc, so for now I just opted to just ignore it entirely until it stabilizes)

## Testing Notes

Loading the project page (specifically the list of surveys) works. It should also load much faster now, since it isn't loading every survey field or the survey spatial (which could cause major slowdowns when loading multiple surveys).


api/src/paths/project/{projectId}/survey/index.ts
```diff
@@ -0,0 +1,170 @@
...
+    {
+      Bearer: []
+    }
+  ],
+  parameters: [
+    {
+      in: 'path',
+      name: 'projectId',
+      schema: {
+        type: 'number'
```
GrahamS-Quartech: nit: integer

api/src/paths/project/{projectId}/survey/index.ts
```diff
@@ -0,0 +1,170 @@
...
+            items: {
+              type: 'object',
+              required: ['surveyData', 'surveySupplementaryData'],
+              properties: {
+                surveyData: {
+                  type: 'object',
+                  required: ['survey_id', 'name', 'start_date', 'end_date', 'focal_species', 'focal_species_names'],
+                  properties: {
+                    survey_id: {
+                      type: 'number',
```
GrahamS-Quartech: nit: integer

api/src/repositories/survey-repository.ts
```diff
@@ -119,6 +119,17 @@ export type StakeholderPartnershipRecord = z.infer<typeof StakeholderPartnership
 
 export type IndigenousPartnershipRecord = z.infer<typeof IndigenousPartnershipRecord>;
 
+export const SurveyBasicFields = z.object({
```
curtisupshall, line 122: +1 for more verbose naming for survey fields. 👍

curtisupshall (commented):
Code looks good! Just needs tests fixed

GrahamS-Quartech:
Tested in the deployed pod and added a bunch of surveys. Very snappy. 👍

api/src/paths/project/{projectId}/survey/index.ts
NickPhura: Fixed

api/src/paths/project/{projectId}/survey/index.ts
NickPhura: Fixed

curtisupshall (approved):
tested, looks good to me!

al-rosenthal (approved):
💎

PR #1159, opened 2023-11-02
SIMSBIOHUB 370: Investigate backend caching for taxonomy endpoint

GrahamS-Quartech:
## Links to Jira Tickets

- [SIMSBIOHUB-370](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-370)

## Description of Changes

- Sets the Cache-Control header in the search and list taxonomy endpoints to have a max-age. This overrides the default no-cache behaviour elsewhere in the app. 

## Testing Notes

- Note that when listing many observations using the same species the network tab may display a request for each row, however the response will be fetched from cache. In Firefox, the request is greyed out a bit to indicate this.

### DIFF
--- a/api/src/paths/taxonomy/species/list.ts
+++ b/api/src/paths/taxonomy/species/list.ts
@@ -77,7 +77,7 @@ export function getSpeciesFromIds(): RequestHandler {
     try {
       const taxonomyService = new TaxonomyService();
       const response = await taxonomyService.getSpeciesFromIds(ids as string[]);
-
+      res.setHeader('Cache-Control', 'max-age=604800')
       res.status(200).json({ searchResponse: response });
     } catch (error) {
       defaultLog.error({ label: 'getSearchResults', message: 'error', error });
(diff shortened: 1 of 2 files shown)


api/src/paths/taxonomy/species/list.ts
```diff
@@ -77,7 +77,7 @@ export function getSpeciesFromIds(): RequestHandler {
     try {
       const taxonomyService = new TaxonomyService();
       const response = await taxonomyService.getSpeciesFromIds(ids as string[]);
-
+      res.setHeader('Cache-Control', 'max-age=604800')
```
curtisupshall: A comment explaining how long this is would be nice (e.g. `// 60 minutes`)

curtisupshall (approved):

api/src/paths/taxonomy/species/list.ts
NickPhura: +1. 

To add additional context, could say something like: `// Overwrite default cache-control header, allow caching up to 60 minutes`

curtisupshall: Done

PR #1163, opened 2023-11-03
SIMSBIOHUB-361: Remove all study areas button

al-rosenthal:
## Links to Jira Tickets

[SIMSBIOHUB-361](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-361)

## Description of Changes

- Added 'Remove All Study Area' button to Survey create/edit page
- Updated error message for study areas
- Updated verbiage for site selection formik error
- Fixed an issue on the map when selecting features on the map (strange black border)

## Testing Notes

- Create or modify a survey to add multiple features to the study area
- Click the remove all button 
- See that all drawn or imported features are removed from the list and map


app/src/features/surveys/components/StudyAreaForm.tsx
```diff
@@ -93,11 +113,26 @@ const StudyAreaForm = () => {
...
+        yesButtonProps={{ color: 'error' }}
+        yesButtonLabel={'Delete'}
+        noButtonProps={{ color: 'primary', variant: 'outlined' }}
+        noButtonLabel={'Cancel'}
+        open={isDeleteOpen}
+        onYes={() => {
+          setIsDeleteOpen(false);
+          onDeleteAll();
+        }}
+        onClose={() => {}}
```
NickPhura: (Jeremy might have an opinion on this, and we might not be consistent on it in the app) When a user clicks outside of the dialog, it calls the onClose. So right now it would do nothing, which is maybe what we want anyways? Or possibly we want it to close?

NickPhura: I think historically we always made the `onClose` do the same thing as the `onNo`, which is probably why its not an optional prop. If thats not the case anymore, we could consider making `onClose` optional, and it could do nothing by default.

al-rosenthal: I've updated it to close the dialog when clicking off of the dialog. I think in most places in the app it will close the dialog and take no action.

GrahamS-Quartech:
Tested this locally. The delete all functionality itself all works as intended, just noticed a couple other things that may be out of scope for this ticket.
- It's not really evident that Study Area is required until you try to save the form without including it.
- There is an odd bit of whitespace on the add version of the study area form, probably in part due to it not having the Define Survey Study Area blurb.
- Sometimes when saving the form, it will show a little error snackbar but still correctly submit the form anyways.

NickPhura:
Found 1 issue with the draw controls delete all button:

1. Draw multiple shapes
2. Click the draw controls delete all button
3. All drawn shapes on the map are removed as expected, but only the newest of the cards is removed from the list (so if you draw 4 shapes, and click delete all, 3 of the cards still remain).

NickPhura:
@GrahamS-Quartech 
>     * Sometimes when saving the form, it will show a little error snackbar but still correctly submit the form anyways.

Yeah its a bug with the scroll to error component. It is still tracking the errors from the previous submit attempt. TBH, I'd be in favour of just removing it (in some other ticket), as the way it works isn't the best, and we don't use it anywhere else. Could also see if we can fix it, if it gets prioritized.

app/src/features/surveys/components/StudyAreaForm.tsx
```diff
@@ -93,11 +111,26 @@ const StudyAreaForm = () => {
...
 
     // set values
-    setFieldValue('locations', data);
+    setFieldValue('locations', data, true);
   };
 
   return (
     <form onSubmit={handleSubmit}>
+      <YesNoDialog
+        dialogTitle={`Delete all ${values.locations.length} Study Areas?`}
```
curtisupshall: If there is only one location, the wording here might get a bit awkward (Delete all 1 Study Areas?)

app/src/features/surveys/components/StudyAreaForm.tsx
```diff
@@ -93,11 +111,26 @@ const StudyAreaForm = () => {
...
+        dialogText="Are you sure you want to remove these study areas? This action can be undone by exiting the survey page without saving"
```
curtisupshall: Kind unsure about the wording here. If the action can be undone, why should the user be prompted?

app/src/features/surveys/components/StudyAreaForm.tsx
```diff
@@ -124,6 +157,31 @@ const StudyAreaForm = () => {
         draw_controls_ref={drawRef}
       />
 
+      {errors.locations && !Array.isArray(errors?.locations) && (
+        <Box pt={2}>
+          <Typography style={{ fontSize: '12px', color: '#f44336' }}>{errors.locations}</Typography>
```
curtisupshall: Is it possible to use `sx={{ color: 'error' }}`?

app/src/features/surveys/observations/ObservationsMap.tsx
```diff
@@ -139,7 +139,10 @@ const ObservationsMap = () => {
               },
               {
                 layerName: 'Sample Sites',
-                features: sampleSiteFeatures.map((feature) => ({ geoJSON: feature, tooltip: <>Sample Site</> }))
+                features: sampleSiteFeatures.map((feature) => ({
+                  geoJSON: feature,
+                  tooltip: <>Sample Site</>
```
curtisupshall: Nit: I would probably use a `<span>` to contain this, instead of a Fragment

curtisupshall (commented):
Code looks good! Just a handful of comments

app/src/features/surveys/components/StudyAreaForm.tsx
al-rosenthal: yeah it does feel a little weird but this is just following our delete pattern of prompting the user before anything gets removed.

app/src/features/surveys/observations/ObservationsMap.tsx
al-rosenthal: updated

app/src/features/surveys/components/StudyAreaForm.tsx
al-rosenthal: this has since been updated to an alert with the severity set to 'error'

al-rosenthal:
> Found 1 issue with the draw controls delete all button:
> 
> 1. Draw multiple shapes
> 2. Click the draw controls delete all button
> 3. All drawn shapes on the map are removed as expected, but only the newest of the cards is removed from the list (so if you draw 4 shapes, and click delete all, 3 of the cards still remain).

I've updated the clear code to account for multiple drawn features

al-rosenthal:
> Tested this locally. The delete all functionality itself all works as intended, just noticed a couple other things that may be out of scope for this ticket.
> 
> * It's not really evident that Study Area is required until you try to save the form without including it.
> * There is an odd bit of whitespace on the add version of the study area form, probably in part due to it not having the Define Survey Study Area blurb.
> * Sometimes when saving the form, it will show a little error snackbar but still correctly submit the form anyways.

I've spoken with @jeznorth and he's gone through and updated a lot of the UI for that component

app/src/features/surveys/components/StudyAreaForm.tsx
NickPhura: One thought: the original file they uploaded might not be around anymore, so deleting the spatial is maybe slightly more dangerous than changing other survey meta. Would be the same for documents too.

NickPhura:
@al-rosenthal Seems to be working.

Just noticed one thing: When drawing items, the map is not smoothly zooming to the new bounds, it just jumps suddenly. Similarly, the new card below the map doesn't slide in smoothly, it kind of just appears abruptly.

I poked around a bit, and it seems to be caused by the call to `validateField('locations');` in `SurveyAreaMapControls.tsx`. I guess because its triggering a full formik re-render or something. Removing it seems to fix the issue.. Do we need that call? None of the other fields in the survey form are validated manually.

al-rosenthal:
> @al-rosenthal Seems to be working.
> 
> Just noticed one thing: When drawing items, the map is not smoothly zooming to the new bounds, it just jumps suddenly. Similarly, the new card below the map doesn't slide in smoothly, it kind of just appears abruptly.
> 
> I poked around a bit, and it seems to be caused by the call to `validateField('locations');` in `SurveyAreaMapControls.tsx`. I guess because its triggering a full formik re-render or something. Removing it seems to fix the issue.. Do we need that call? None of the other fields in the survey form are validated manually.

Good catch, I've removed the offending line.

NickPhura (approved):

MacDeluca (approved):
Looks gud

PR #1165, opened 2023-11-06
SIMSBIOHUB-371: Update Seed to Create Sampling Site and Observation Records

GrahamS-Quartech:
## Links to Jira Tickets

- [SIMSBIOHUB-371](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-371)

## Description of Changes

- Added SQL to insert a sampling site, a sampling method, a sampling period, and 3 observations. Observations should be generated with a valid taxonomy id and lat lon values that will put them somewhere around BC.


al-rosenthal (approved):
ran this locally, looks good 🥼

database/src/seeds/03_basic_project_survey_setup.ts
```diff
@@ -449,6 +453,176 @@ const insertSurveyStakeholderData = (surveyId: number) => `
...
+      "type": "Feature",
+      "geometry": {
+        "type": "Polygon",
+        "coordinates": [
+          [
+            [-121.904297, 50.930738],
+            [-121.904297, 51.971346],
+            [-120.19043, 51.971346],
+            [-120.19043, 50.930738],
+            [-121.904297, 50.930738]
```
NickPhura: nit: Can we define a different polygon, just so this one and the survey boundary aren't perfectly overlapping.
Actually, if this could be smaller, and inside the survey boundary, that would be more accurate to real life even.

If you haven't used it before, this site is helpful for plotting/generating geojson: http://geojson.io

JeremyQuartech: It would be neat if these values were also generated using faker, but this might be a lot more work than it's worth.

database/src/seeds/03_basic_project_survey_setup.ts
```diff
@@ -449,6 +453,176 @@ const insertSurveyStakeholderData = (surveyId: number) => `
...
+          public.ST_Force2D(public.ST_GeomFromGeoJSON('
+            {
+                "type": "Polygon",
+                "coordinates": [
+                  [
+                    [-121.904297, 50.930738],
+                    [-121.904297, 51.971346],
+                    [-120.19043, 51.971346],
+                    [-120.19043, 50.930738],
+                    [-121.904297, 50.930738]
```
JeremyQuartech: Here too

database/src/seeds/03_basic_project_survey_setup.ts
GrahamS-Quartech: I updated the polygon, it's a smaller rectangle within the study area now.

database/src/seeds/03_basic_project_survey_setup.ts
```diff
@@ -495,11 +495,11 @@ const insertSurveySamplingSiteData = (surveyId: number) =>
...
-                    [-121.904297, 50.930738],
-                    [-121.904297, 51.971346],
-                    [-120.19043, 51.971346],
-                    [-120.19043, 50.930738],
-                    [-121.904297, 50.930738]
+                    [-121,    51],
+                    [-121,    51.7],
+                    [-120.5,  51.7],
+                    [-120.5,  51],
+                    [-121,    51]
```
NickPhura, line 502: Can we make this one different from the one above?

NickPhura: ohh, read this more closely, ignore this!

NickPhura (approved):

PR #1166, opened 2023-11-07
SIMSBIOHUB-297: Keycloak Value Case Sensitivity

JeremyQuartech:
## Links to Jira Tickets

- [SIMSBIOHUB-297](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-297?)

## Description of Changes

- Removes toLowerCase casting on user_identifier fields from keycloak and stores raw string value in the database regardless of case
- Adds LOWER() sql statements on any relation to user_identifiers and user guids as to make them case-insensitive

## Testing Notes

- Test with IDIR and BCeID, as well as access request and approval workflows
- Any other functionality which relates to user_guids or user_identifiers

- Noticed that given_name and family_name do not get added to the database when a new user is added. Not sure if this is intentional or not.


NickPhura (approved):
🐐

al-rosenthal (approved):
looks good 🛩️

PR #1167, opened 2023-11-07
SIMSBIOHUB-345: Display Time in Sampling Periods

KjartanE:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-345

## Description of Changes

- Added Time picker for start and end time on sample site periods.
- Database changes include 2 more columns for start,end time. 

## Testing Notes


app/src/components/fields/DateTimeFields.tsx
```diff
@@ -0,0 +1,147 @@
...
+  };
+  time: {
+    timeLabel: string;
+    timeName: string;
+    timeId: string;
+    timeRequired: boolean;
+    timeHelperText?: string;
+    timeIcon: string;
+  };
+  formikProps: any;
```
GrahamS-Quartech: I think you can probably use FormikContextType here.

app/src/components/fields/DateTimeFields.tsx
```diff
@@ -0,0 +1,147 @@
...
+                fullWidth: true
+              }
+            }}
+            label={dateLabel}
+            format={DATE_FORMAT.ShortDateFormat}
+            minDate={moment(DATE_LIMIT.min)}
+            maxDate={moment(DATE_LIMIT.max)}
+            value={formattedDateValue}
+            onChange={(value) => {
+              if (!value || String(value.creationData().input) === 'Invalid Date') {
```
GrahamS-Quartech, line 93: Could use moment().isValid() here instead.

GrahamS-Quartech:
Couple things I noticed:
- The modal will allow you to set an end time occurring before the start time when the dates are the same.
- If you enter the edit sampling site page, then edit either time period time field and hit Update in that modal, it will update the card with that new time. However if you re-enter that edit modal, the time will have reverted to the previous value.
- I think the layout of how the fields and trash can are arranged could be better but I don't really know what would look best. Maybe check in with Jeremy?

KjartanE:
> Couple things I noticed:
> 
>     * The modal will allow you to set an end time occurring before the start time when the dates are the same.
> 
>     * If you enter the edit sampling site page, then edit either time period time field and hit Update in that modal, it will update the card with that new time. However if you re-enter that edit modal, the time will have reverted to the previous value.
> 
>     * I think the layout of how the fields and trash can are arranged could be better but I don't really know what would look best. Maybe check in with Jeremy?

Good points, I have handled those situations accordingly!

app/src/features/surveys/components/MethodForm.tsx
```diff
@@ -72,7 +84,25 @@ export const SamplingSiteMethodYupSchema = yup.object({
...
           .required('End Date is required')
-          .isEndDateSameOrAfterStartDate('start_date')
+          .isEndDateSameOrAfterStartDate('start_date'),
+        start_time: yup.string().nullable(),
+        end_time: yup
+          .string()
+          .nullable()
+          .test(
+            'checkDatesAreSameAndEndTimeIsAfterStart',
+            'End Date and Time must be after Start Date and Time',
```
al-rosenthal: nit: this should say End time and Start time

al-rosenthal (commented):
adding and editing works as expected

app/src/features/surveys/components/MethodForm.tsx
curtisupshall: I actually think "End date and time must be after start date and time" is better than "End date and end time must be...", since the latter suggests that the end time must always be after the start time, which isn't necessarily true if the end date is after the start date.

al-rosenthal (approved):
🕐

app/src/features/surveys/components/SamplingMethodForm.tsx
```diff
@@ -190,6 +194,16 @@ const SamplingMethodForm = () => {
                             <Icon path={mdiCalendarRangeOutline} size={1} />
                           </ListItemIcon>
                           <ListItemText primary={`${period.start_date} to ${period.end_date}`} />
+                          {period.start_time && (
+                            <>
+                              <ListItemIcon>
+                                <Icon path={mdiClockOutline} size={1} />
+                              </ListItemIcon>
+                              <ListItemText
+                                primary={`${period.start_time} ${(period.end_time && `to ${period.end_time}`) || ''}`}
```
curtisupshall, line 203: Could avoid the trailing whitespace:

```javascript
primary={[period.start_time, period.end_time].filter(Boolean).join(' to ')}
```

JeremyQuartech (approved):
Tested locally, everything working correctly

PR #1168, opened 2023-11-09
SIMSBIOHUB-332 - Observation Table Validation

curtisupshall:
## Links to Jira Tickets
 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-332

## Description of Changes
 - Adds a validation state object to the `observationTableContext`
 - Adds error styling to Observations Table controls that have failed validation
 - Adds an alert to the Observations Table, allowing users to page through validation errors
 - Prevents saving to the Observations Table unless all column values are valid

## Testing Notes

 - When saving changes to the observation table, if any observation table row contains an invalid value, that cell should enter an error state (red outline/color) and prevent saving until it is fixed. Validation errors occur when:
   - Any of the rows contain a `null` or `undefined` value (unless all three Sampling columns are empty - in which case, the observation is incidental, and none of the Sampling columns should be validated)
   - An observation date is defined but invalid
   - An observation time is defiend by invalid
 - Users should be able to page through the errors using the alert bar
 - When a validation error is fixed and a follow-up attempt to save the observation records is made, the errored cell should no longer have an error state


app/src/components/map/components/MapBaseCss.tsx
al-rosenthal, line 1: this looks like it might be a duplicate of the file at `app/src/components/map/styles`

al-rosenthal (commented):
We've talked about these but just writing them down here:
- when creating a record and selecting a date, the date defaults to today regardless of what I picked
- date is invalid with todays date
- it is not obvious when imported records have validation errors when saving a new record

NickPhura:
@curtisupshall 
Re validation: Noticed that if you don't select a sample site, none of the sample_* fields get highlighted red nor does the error status bar show up, upon clicking save.

al-rosenthal:
I ran into an issue with the date validation. I can only add a record manually when I type the date in, I can't use the picker, it flags the date when I try and save it

curtisupshall:
> @curtisupshall 
> 
> Re validation: Noticed that if you don't select a sample site, none of the sample_* fields get highlighted red nor does the error status bar show up, upon clicking save.

@NickPhura This is actually expected behaviour (unless I misunderstood the business decision); a record can have all of its sampling fields left undefined, thus making it an "incidental" observation. But if one of the fields is filled in, then all of them need to be non-empty. We came to this decision after discussing the issue around importing observations in the absence of sampling data.

al-rosenthal (dismissed):
ran this locally and it looks good. The banner is nice and snappy 🐇

MacDeluca (dismissed):
Looks super good, and it's very responsive.
Few small things I noticed:

1. On row error the colouring of the default text for the date / time is different. Attached screenshot
2. Nit: Does it make more sense to have the Prev / Next buttons horizontal icons ie: < or >. Since it traverses the errors in the row horizontally?
3. This might be out of the scope of this ticket, but it appears there is a rendering issue with the header. When you scroll the hidden columns into view it leaves a line fragment. Attached screenshot
4. Should the cell validation errors go away as you update them ie: onChange?
image
image

NickPhura:
@curtisupshall 
Looks good, the navigation between errors is super cool.

On the sample site form
- I was getting some duplicate key warnings. I do have a fix for it that I can push up (just adjusting the keys for the sample method/period fields).

On the observations table
- I did notice a couple warnings in the console about uncontrolled values in the observations data grid. Probably not a blocker for this PR, but something we should try and resolve if we can.

curtisupshall:
> Looks super good, and it's very responsive. Few small things I noticed:
> 
> 1. On row error the colouring of the default text for the date / time is different. Attached screenshot
> 2. Nit: Does it make more sense to have the Prev / Next buttons horizontal icons ie: < or >. Since it traverses the errors in the row horizontally?
> 3. This might be out of the scope of this ticket, but it appears there is a rendering issue with the header. When you scroll the hidden columns into view it leaves a line fragment. Attached screenshot
> 4. Should the cell validation errors go away as you update them ie: onChange?
>    image
>    image

I think we'll address these changes in Jer's upcoming `UI-Fixes` branch (although I did fix the cell coloring).

curtisupshall:
> @curtisupshall Looks good, the navigation between errors is super cool.
> 
> On the sample site form
> 
> * I was getting some duplicate key warnings. I do have a fix for it that I can push up (just adjusting the keys for the sample method/period fields).
> 
> On the observations table
> 
> * I did notice a couple warnings in the console about uncontrolled values in the observations data grid. Probably not a blocker for this PR, but something we should try and resolve if we can.

Resolved!

al-rosenthal (approved):
Ran this locally and it looks great! 🍔

MacDeluca (approved):
Love it!

PR #1174, opened 2023-11-21
BugFixes: Misc Performance Updates and Related Cleanup

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

* Adjust API memory settings, set node env properly, adjust log levels
  * Added a NODE_OPTIONS pipeline param so that the max heap size is getting set explicitly. (I noticed in the pods in dev that despite the deploy config specifying 2Gig max memory, the max heap size was only 1Gig).

* Some misc pipeline cleanup around `env` vs `nodeEnv`.
  * Some of the same keys were being referenced from different objects for different purposes, and it was getting a bit hard to track. So now `env` is the pipeline phase, and `nodeEnv` is the node environment variable.
  * Also tried to cleanup our node_env vars in general. We had a mixture of 'dev', 'test', 'prod', etc, which aren't even the ones that node/react look for. So now it should just be 'development' and 'production'.

* Tweak parameters of trace command

* Update tsconfigs

* Pin zod package version, Remove some unused packages
  * Was some talk on the zod repo about possible issues with newer versions of typescript. Nothing concrete or certain if it was due to zod or typescript, but just to be safe, pinning the version to what we have now.

* Simplify zod geojson schema
  * Only used in 1 spot in one of the bctw services. But, in an effort to prevent zod from taking a long time to process on large geometries, I simplified them to just check the top level structure, and not dig into every sub level.

* Remove zod from keycloak token type definitions (hotspot).
  * Was flagged as a hotspot, and we never really used them that much. Also, since we get those tokens from the gov keycloak, parsing them shouldn't be needed, as if they change unexpected the app is going to break regardless.

* Disable zod database response parsing in production.

* __New:__ Add query and zod timers in development.
  * In non-production, if you set `LOG_LEVEL=silly` it will log timers for the database query execution and subsequent zod schema parsing.

* BugFix: Allow surveys to be created with $0 amounts.
  * Just had a bad falsy check that turned '0' into 'null'

## Testing Notes

Everything works as before.


api/src/database/db.ts
```diff
@@ -335,14 +334,32 @@ export const getDBConnection = function (keycloakToken: object): IDBConnection {
     sqlStatement: SQLStatement,
     zodSchema?: z.Schema<T, any, any>
   ): Promise<pg.QueryResult<T>> => {
+    if (process.env.NODE_ENV === 'production') {
+      // Don't run timers or zod schemas in production
```
curtisupshall, line 338: Good call 👍

curtisupshall (dismissed):
Code looks good to me!

curtisupshall (approved):
Tested locally; everything works as expected. ✅

api/src/utils/keycloak-utils.ts
```diff
@@ -1,93 +1,84 @@
...
-  email_verified: z.boolean(),
-  name: z.string(),
-  preferred_username: z.string(),
-  display_name: z.string(),
-  given_name: z.string(),
-  family_name: z.string(),
-  email: z.string()
-});
-export type IdirUserInformation = z.infer<typeof IdirUserInformation>;
+export type IdirUserInformation = {
```
al-rosenthal, line 6: was the zod schema causing some issues?

al-rosenthal (approved):
🦆

al-rosenthal (approved):
🍖

KjartanE (approved):
🐈

api/src/utils/keycloak-utils.ts
NickPhura: The way I was parsing the tokens before was being flagged as a performance hotspot by those trace commands. So I was trying to simplify some of the keycloak stuff in general, and remove unused code, etc.

PR #1178, opened 2023-12-06
SIMSBIOHUB-396: Added aliases to column validator

curtisupshall:
## Links to Jira Tickets

 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-396
## Description of Changes

 - Updates the Observations import pipeline to support alternative spellings and cases for observation latitude and longitude columns:
   - Lat and Latitude should both be recognized as Latitude.
   - Lon, Long, Lng, and Longitude should both be recognized as Longitude
   - Upper and lowercase should both work (case insensitive)


api/src/utils/xlsx-utils/worksheet-utils.test.ts
```diff
@@ -0,0 +1,33 @@
+import xlsx from 'xlsx';
+import { expect } from 'chai';
+import { describe } from 'mocha';
+import sinon from 'sinon'
+
+import * as worksheet_utils from "./worksheet-utils";
+
+
+describe.only('worksheet utils', () => {
```
NickPhura: nit: dont forget to remove the `.only`

NickPhura: (Just noticed it also showed up as a SonarCloud bug, didn't know it did that, but thats cool)

al-rosenthal (approved):
I ran this locally, changing the lat/ long columns and everything worked. 🏖️

NickPhura (approved):

PR #1179, opened 2023-12-07
SIMSBIOHUB-379: Dataset object generation fix for geometry

KjartanE:
## Description of Changes
- Validation quick fix for dataset geometry

## Testing Notes


al-rosenthal (commented):
I tried running this locally and ran into an issue once the data hit BioHub. The error in the biohub API was "publish_timestamp" of relation "submission" does not exist. My Biohub app was running off of the feature branch `dataset_security_feature`

api/src/services/platform-service.ts
```diff
@@ -324,8 +325,19 @@ export class PlatformService extends DBService {
...
     const survey = await surveyService.getSurveyData(surveyId);
     const { surveyObservations } = await observationService.getSurveyObservationsWithSupplementaryData(surveyId);
-
-    const surveyDataPackage = new PostSurveyToBiohubObject(survey, surveyObservations, additionalInformation);
+    const surveyLocation = await surveyService.getSurveyLocationsData(surveyId);
+
+    const geometryFeatureCollection = {
+      type: 'FeatureCollection',
+      features: surveyLocation.flatMap((location) => (location.geojson as unknown) as Feature)
+    } as GeoJSON.FeatureCollection;
```
NickPhura: You should be able to remove this cast, and just make it the type of `geometryFeatureCollection`?

Also, is `location.geojson` not already typed?

In general, we should see if we can get all of the typings happy, so we can avoid any `as` casting.

NickPhura (approved):

PR #1180, opened 2023-12-07
BugFix: Fix api_set_context Function

NickPhura:
## Links to Jira Tickets

Fix api_set_context function

## Description of Changes

Dev is currently failing because the api_set_context function is failing (not returning any rows).
- Had added a call to LOWER() in the db.ts, not sure if it is the issue breaking dev, but its the only recent change involving the context. Moved LOWER() into into the api_set_context function, wrapping the string where clauses so everything is checked as lowercase.

## Testing Notes

- App works as normal.


KjartanE (approved):
🐈

al-rosenthal (approved):
🦆

curtisupshall (approved):

PR #1202, opened 2024-01-23
SIMSBIOHUB-445: Feature Branch - Consolidate Observations/Telemetry/Animals

curtisupshall:
## Links to Jira Tickets

 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-445

## Description of Changes

- Combined the Telemetry and Observations map/table view on the Survey page, allowing users to toggle between them;
- Added server-side support for paginating the observations fetching endpoint;
- Modified the Manage Observations page to use server-side pagination;
- Added a new endpoint to fetch all the spatial data relating to survey observations;
- Deleted some old code related to Survey Summary Results.

## Testing Notes

You should be able to:
- [ ] View observation and telemetry points on the survey page
- [ ] On the survey page, sort on various columns and page through the observations table (except for Species; see [465](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-465))
- [ ] On the Manage Observations page, sort on various columns and page through the observations

PR #1206, opened 2024-02-01
SIMSBIOHUB-414: Show Feature Popup on Survey Page

curtisupshall:
## Links to Jira Tickets

- [Figma](https://www.figma.com/file/SstgKUBlafIk5PZbi8LEua/Map-Popups?type=design&node-id=0-1&mode=design)
- [SIMSBIOHUB-414](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-414)

## Description of Changes

- Adds survey sampling sites and study areas back to the Survey Map (since being removed by https://github.com/bcgov/biohubbc/pull/1202);
- Adds various popups to survey map features, showing various point/polygon metadata;
- Removes extraneous observation table hook call;
- Updates method names and documentation for telemetry API hooks.

## Testing Notes

- See Figma for appropriate popup styling/metadata.


app/src/features/surveys/view/SurveyMap.tsx
```diff
@@ -41,15 +82,71 @@ const SurveyMap = (props: ISurveyMapProps) => {
...
               <GeoJSON
                 key={key}
-                data={nonEditableGeo.feature}
-                pointToLayer={(_, latlng) => coloredPoint({ latlng, fillColor: '#1f7dff', borderColor: '#ffffff' })}>
-                {nonEditableGeo.popupComponent}
+                data={mapPoint.feature}
+                pointToLayer={(_, latlng) => coloredPoint({ latlng, fillColor: '#1f7dff', borderColor: '#ffffff' })}
+                onEachFeature={(feature, layer) => {
+                  layer.on({
+                    popupopen: () => {
```
NickPhura: Does this need to check `mapPointMetadata` for previously loaded metadata for the point, and only call onLoad if it is missing?

app/src/features/surveys/view/SurveyMap.tsx
```diff
@@ -41,15 +82,71 @@ const SurveyMap = (props: ISurveyMapProps) => {
           <FullScreenScrollingEventHandler bounds={bounds} scrollWheelZoom={false} />
           <SetMapBounds bounds={bounds} />
 
-          {props.mapPoints?.map((nonEditableGeo: INonEditableGeometries, index: number) => {
-            const key = nonEditableGeo.feature.id ?? index;
+          {props.mapPoints?.map((mapPoint: ISurveyMapPoint, index: number) => {
+            const { key } = mapPoint;
```
NickPhura: probably subjective: But I'd lean towards either de-referencing all values from mapPoint or none of them. Rather than having a mixture.

NickPhura: (I tend to like de-referencing none of the values if the code is more than 1 page long, just because its easier to know what prop the values came from when reading farther than the page, but that is definitely just a personal preference. Though, I don't think I am even super consistent, because when it comes to Formik props, for example, I always de-reference them. So I dunno :/)

curtisupshall: Agreed! I'm definitely a bit more cautious about dereferencing for this reason

app/src/features/surveys/view/SurveyMap.tsx
curtisupshall: Good catch! It should not make repeated fetches.

app/src/features/surveys/view/SurveyMap.tsx
curtisupshall: I think I did it so I could cash in on the syntactic sugar :P
```javascript
return {
  key,
  ...
```

al-rosenthal (dismissed):
Ran this locally, looks and runs well! 📈

app/src/features/surveys/view/SurveyMap.tsx
```diff
@@ -1,29 +1,177 @@
...
+              overflow: 'hidden',
+              textOverflow: 'ellipsis',
+              textTransform: 'uppercase'
+            }}>
+            {props.title}
+          </Typography>
+          <Box component="dl" mt={1} mb={0}>
+            {props.metadata.map((metadata, index) => (
+              <Stack
+                key={`${metadata.label}-${index}`}
```
MacDeluca: Is there a better key to use instead of the index?

curtisupshall: I guess the only thing better than using the index here would be to use `metadata.value`. But `metadata.label` is already pretty much guarenteed to be unique, so the index is just used as redundancy

MacDeluca: I think if we can get away from using the index it would be prefered. Just so we dont have to re-render the whole list if the order of the items change.

curtisupshall: Done

MacDeluca (approved):
LGTM

al-rosenthal (approved):
🍿

PR #1207, opened 2024-02-02
SIMSBIOHUB-462: Observation Subcount Schema

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-462

## Description of Changes

New migration to add 3 new tables, for tracking observation subcount records, an associated critterbase event id, and associates to critters.

## Testing Notes

Migration runs successfully. No existing functionality touches these tables, so they shouldn't have any impact on existing code.


database/src/migrations/20240202000001_observation_subcount_attribute_tables.ts
```diff
@@ -0,0 +1,193 @@
...
+  await knex.raw(`--sql
+    -------------------------------------------------------------------------
+    -- Create tables
+    -------------------------------------------------------------------------
+    SET search_path = biohub;
+
+    CREATE TABLE observation_subcount(
+      observation_subcount_id    integer            GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+      survey_observation_id      integer            NOT NULL,
+      count                      integer            NOT NULL,
```
curtisupshall: Would you consider renaming this to `subcount`? It could make joins on `survey_observation` a bit easier to reason about.

KjartanE (dismissed):
Ran locally, tables built properly

database/src/migrations/20240202000001_observation_subcount_attribute_tables.ts
NickPhura: Yeah fair point. Updated.

curtisupshall (approved):

database/src/migrations/20240202000001_observation_subcount_attribute_tables.ts
```diff
@@ -0,0 +1,193 @@
...
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
+    -------------------------------------------------------------------------
+    -- Create tables
+    -------------------------------------------------------------------------
+    SET search_path = biohub;
+
+    CREATE TABLE observation_subcount(
```
al-rosenthal, line 23: might want to make this table name consistent with other sub count tables `subcount_observation`

NickPhura: Yeah this was semi-intentional, though I'm not strongly for or against either name.

I was sort of following a cascading naming pattern:
- survey_observation
- observation_subcount
- subcount_attribute
- attribute_type (if this existed)

al-rosenthal: yeah that's fair, if you've got reason for it I'm fine with the table name as is.

al-rosenthal (approved):
🌮

PR #1208, opened 2024-02-06
421 - BCTW Multiple Device Fix

MacDeluca:
## Links to Jira Tickets

- [SIMSBIOHUB-421](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-421)

## Description of Changes

- Multiple devices can be created using the same device_id for critters (with different device makes)

## Testing Notes

- Add a critter
- Add a deployment using a device id and a device make
- Add another deployment with same device id and different device make (should succeed)
- Add another deployment with same device id and same device make (should fail)


app/src/features/surveys/view/survey-animals/telemetry-device/DeploymentForm.tsx
```diff
@@ -51,6 +51,7 @@ export const DeploymentForm = (props: DeploymentFormSectionProps): JSX.Element =
   };
 
   const deploymentOverlapTest = (deployment: IDeploymentTimespan) => {
+    console.log(deviceDetails);
```
al-rosenthal: nit

app/src/features/surveys/view/survey-animals/telemetry-device/TelemetryDeviceForm.tsx
```diff
@@ -20,27 +20,27 @@ const TelemetryDeviceForm = (props: ITelemetryDeviceFormProps) => {
...
   );
 
-  const validateDeviceMake = async (value: number | '') => {
-    const deviceMake = deviceDetails?.device?.device_make;
-    if (device.device_id && deviceMake && deviceMake !== value && mode === ANIMAL_FORM_MODE.ADD) {
-      return `The current make for this device is ${deviceMake}`;
-    }
-  };
-
+  //TODO remove?
```
al-rosenthal: we can probably remove this, looks like the old variable isn't being used anymore and if we really need this code it is in the repo already

al-rosenthal (commented):
I ran this locally and was able to add multiple device makes as you'd expect. When I tried with data that should 'fail' I noticed that the date inputs gave warnings of the same device make within the date range and could not save. Is this what you meant by 'fail'?

api/src/services/bctw-service.ts
```diff
@@ -273,23 +273,28 @@ export class BctwService {
...
    *
    * @param {number} deviceId
+   * @param {string} deviceMake
    * @returns {*} {Promise<IDeploymentRecord[]>}
    * @memberof BctwService
    */
-  async getDeviceDeployments(deviceId: number): Promise<IDeploymentRecord[]> {
-    return await this._makeGetRequest(GET_DEPLOYMENTS_BY_DEVICE_ENDPOINT, { device_id: String(deviceId) });
+  async getDeviceDeployments(deviceId: number, deviceMake: string): Promise<IDeploymentRecord[]> {
+    return await this._makeGetRequest(GET_DEPLOYMENTS_BY_DEVICE_ENDPOINT, {
```
curtisupshall: Nit: prefer `return Promise` instead of `return await Promise`

api/src/services/bctw-service.ts
```diff
@@ -273,23 +273,28 @@ export class BctwService {
...
    * @memberof BctwService
    */
-  async getDeviceDetails(deviceId: number): Promise<IDevice[]> {
-    return this._makeGetRequest(`${GET_DEVICE_DETAILS}${deviceId}`);
+  async getDeviceDetails(deviceId: number, deviceMake: string): Promise<IDevice[]> {
+    return this._makeGetRequest(`${GET_DEVICE_DETAILS}${deviceId}`, { make: deviceMake });
   }
 
   /**
    * Get deployments by device id, may return results for multiple critters.
```
curtisupshall: This JS doc should be amended.

curtisupshall (approved):
LGTM

al-rosenthal (approved):
tried this out and works as expected 🍰

PR #1209, opened 2024-02-06
SIMSBIOHUB-460, 475 - Study Area + Observations Import bugs

curtisupshall:
## Links to Jira Tickets

- [SIMSBIOHUB-460](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-460)
- [SIMSBIOHUB-475](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-475)

## Description of Changes

- Fixed a bug where newly added Study Areas were not shown until page reload
- Fixed a bug where users could not import survey observations

### Test 1
1. Edit a survey.
2. Delete its existing study area, then add a new one.
3. Save the survey. You should:
 - [ ] See that the new study area is shown on the survey page

### Test 2
1. Delete all observations for a particular survey.
2. Import a survey observation CSV. You should:
 - [ ] Immediately see the new observation records
 - [ ] be able to navigate back to the survey page and see the same records.

### DIFF
--- a/app/src/contexts/observationsTableContext.tsx
+++ b/app/src/contexts/observationsTableContext.tsx
@@ -381,7 +381,6 @@ export const ObservationsTableContextProvider = (props: PropsWithChildren<Record
           open: true
         });
 
-        console.log('Refreshing after delete...');
         refreshObservationRecords();
       } catch {
         // Close yes-no dialog
--- a/app/src/contexts/observationsTableContext.tsx
+++ b/app/src/contexts/observationsTableContext.tsx
@@ -219,14 +219,6 @@ export const ObservationsTableContextProvider = (props: PropsWithChildren<Record
   });
   const [sortModel, setSortModel] = useState<GridSortModel>([]);
 
-  // const updatePaginationModel = (model: GridPaginationModel) => {
-  //   setPaginationModel(model);
-  // };
-
-  // const updateSortModel = (model: GridSortModel) => {
-  //   setSortModel(model);
-  // };
-
   // Fetch new rows based on sort/ pagination model changes
   useEffect(() => {
     refreshObservationRecords();
@@ -389,7 +381,7 @@ export const ObservationsTableContextProvider = (props: PropsWithChildren<Record
           open: true
         });
 
-        console.log('Refreshing after delete...')
+        console.log('Refreshing after delete...');
         refreshObservationRecords();
       } catch {
         // Close yes-no dialog
--- a/app/src/contexts/observationsTableContext.tsx
+++ b/app/src/contexts/observationsTableContext.tsx
@@ -219,35 +219,17 @@ export const ObservationsTableContextProvider = (props: PropsWithChildren<Record
   });
   const [sortModel, setSortModel] = useState<GridSortModel>([]);
 
-  const updatePaginationModel = (model: GridPaginationModel) => {
-    setPaginationModel(model);
-  };
+  // const updatePaginationModel = (model: GridPaginationModel) => {
+  //   setPaginationModel(model);
+  // };
 
-  const updateSortModel = (model: GridSortModel) => {
-    setSortModel(model);
-  };
-
-  // initial load with pagination values
-  useEffect(() => {
-    observationsContext.observationsDataLoader.refresh({
-      limit: paginationModel.pageSize,
-
-      // API pagination pages begin at 1, but MUI DataGrid pagination begins at 0.
-      page: paginationModel.page + 1
-    });
-  }, []);
+  // const updateSortModel = (model: GridSortModel) => {
+  //   setSortModel(model);
+  // };
 
   // Fetch new rows based on sort/ pagination model changes
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


app/src/contexts/observationsTableContext.tsx
```diff
@@ -406,6 +380,9 @@ export const ObservationsTableContextProvider = (props: PropsWithChildren<Record
           ),
           open: true
         });
+
+        console.log('Refreshing after delete...');
```
al-rosenthal: nit

app/src/features/surveys/components/locations/SurveyAreaMapControl.tsx
```diff
@@ -152,7 +152,7 @@ export const SurveyAreaMapControl = (props: ISurveyAreMapControlProps) => {
                 geojson: [feature],
                 revision_count: 0,
                 leaflet_id: id,
-                uuid: v4()
+                uuid: v4() // TODO why does ISurveyLocation have this property?
```
al-rosenthal: the interface definition has a brief description but it is just used as a unique key when the locations cards are being rendered. It is being used in `SurveyAreaList` on line 79

app/src/features/surveys/components/locations/SurveyAreaMapControl.tsx
```diff
@@ -204,9 +204,12 @@ export const SurveyAreaMapControl = (props: ISurveyAreMapControlProps) => {
           <StaticLayers
             layers={values.locations
               .filter((item) => !item?.leaflet_id) // filter out user drawn locations
-              .map((item) => {
+              .map((item, index) => {
                 // Map geojson features into layer objects for leaflet
-                return { layerName: item.name, features: item.geojson.map((geo) => ({ geoJSON: geo })) };
+                return {
+                  layerName: item.name,
+                  features: item.geojson.map((geo) => ({ geoJSON: geo, key: index }))
```
al-rosenthal: might be able to use the previously mentioned uuid here as the key instead of the index

app/src/features/surveys/view/SurveyMap.tsx
```diff
@@ -173,6 +173,94 @@ const SurveyMap = (props: ISurveyMapProps) => {
...
+                metadata={mapPointMetadata[mapPoint.key]}
+              />
+            )
+          };
+        })
+      };
+    })
+  ];
+
+  // console.log('staticlayers:', staticLayers)
```
al-rosenthal, line 262: nit

al-rosenthal (commented):
ran this locally and things work as described

app/src/features/surveys/components/locations/SurveyAreaMapControl.tsx
curtisupshall: Good call.

al-rosenthal (approved):
🐛

app/src/contexts/observationsTableContext.tsx
```diff
@@ -574,11 +550,16 @@ export const ObservationsTableContextProvider = (props: PropsWithChildren<Record
...
   const refreshObservationRecords = useCallback(async () => {
+    const sort = firstOrNull(sortModel);
     return observationsContext.observationsDataLoader.refresh({
-      page: paginationModel.page,
-      limit: paginationModel.pageSize
+      limit: paginationModel.pageSize,
+      sort: sort?.field || undefined,
+      order: sort?.sort || undefined,
+
+      // API pagination pages begin at 1, but MUI DataGrid pagination begins at 0.
```
NickPhura, line 559: I notice that in the postgres queries, pagination is 0 based. Is there any reason we convert from 0 to 1 on the frontend, and then back to 0 on the backend (rather than just have it 0 based the entire time)?

curtisupshall: The motivation to have 1-indexed pagination was because exposing the first page as `page=1` to the user seemed more sensible than `page=0`. Although the page number isn't really "exposed" to the user outside of the page number we show in the frontend, I think at the time this was decided, having the pagination query persist in the browser URL was still up in the air. So it made sense that the first page should naturally be `1`. Also, the frontend and backend code were developed independantly, so the zero-to-one and one-to-zero conversions were made on both sides to satisfy this decision. Maybe if we had picked a word like `offset` to go in the query parameters... I think `offset=0` would be totally acceptable. I still lean towards page number begins at `1` but I'm open to opposing arguments 😅

NickPhura: Gotcha. Not a big deal either way, really. The UI could render `page+1` to the user, even if the URL tracked `page=0` (I don't think the average user is going to notice or care if the URL says page 0). I think making the rest of the code simpler is what I would prioritize. But, I really don't think its necessary to change anything at this point. As long as everything is well documented/typed, then it really shouldn't matter one way or the other. I was just curious if there was some other reason for avoiding 0 that I wasn't aware of.

NickPhura (approved):

PR #1210, opened 2024-02-06
SIMSBIOHUB-459 - Update SIMS To Use ITIS and BioHub Taxonomy

curtisupshall:
## Links to Jira Tickets

 - [SIMSBIOHUB-459](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-459)

## **Quickstart for the new `.env` variables**
Paste these into your `.env`, replacing the old `BACKBONE_API_HOST`:
```bash
# BACKBONE_INTERNAL_API_HOST=http://host.docker.internal:6200
BACKBONE_INTERNAL_API_HOST=https://api-dev-biohub-platform.apps.silver.devops.gov.bc.ca

# BACKBONE_PUBLIC_API_HOST=http://localhost:6200
BACKBONE_PUBLIC_API_HOST=https://api-dev-biohub-platform.apps.silver.devops.gov.bc.ca
```

## Description of Changes

- `.env` and pipeline changes to reflect new environment variables:
  - `BACKBONE_INTERNAL_API_HOST`: The internal host used by the SIMS API to talk to BioHub
  - `BACKBONE_PUBLIC_API_HOST`: The public host used by the SIMS frontend (and other services) to talk to BioHub
- Removes code associated with elasticsearch;
- Replaces taxonomy-related API calls with BioHub taxonomy endpoint calls;
- Shows error dialog whenever observation import fails;
- Adds new `Species` column alias support for observation importing:
  - `Species`
  - `Tsn`
  - `Itis_tsn`
  - `Taxon`
- Minor edits to the ErrorDialog;
- Updates database seeds to use ITIS TSN instead of elasticsearch;
- Updates the `survey_observation` and `study_species` tables to support ITIS TSN/scientific name.

## Testing Notes

You should be able to:
- [ ] Import observations, supporting allowed column aliases;
- [ ] Edit observations;
- [ ] Insert new observations;
- [ ] Create/edit a survey with focal species + ancillary species.


database/src/migrations/20240124141800_observation_scientific_name.ts
```diff
@@ -11,8 +12,34 @@ export async function up(knex: Knex): Promise<void> {
...
+    COMMENT ON COLUMN study_species.wldtaxonomic_units_id IS '(Deprecated) System generated UID for a taxon..';
+    COMMENT ON COLUMN study_species.itis_tsn IS 'The ITIS TSN identifier for the species associated with the survey.';
+
+    -- TODO Currently, we are using a placeholder ITIS species. Instead, we should map all known species used in prod from their elsaticsearch taxonomic ID to their ITIS TSN.
+
+    -- Placeholder:
+      UPDATE study_species SET itis_tsn = 202384;
+
+    -- Add not null constraint
+    ALTER TABLE study_species ALTER COLUMN itis_tsn SET NOT NULL;
```
NickPhura, line 32: We may as well do the same thing for the itis_scientific_name (set a default placeholder and mark it not null)

curtisupshall: Good call

curtisupshall: Actually it seems like this table (`study_species`) doesn't have a column `itis_scientific_name`. Or were you referring to a different table?

api/src/services/observation-service.ts
```diff
@@ -27,12 +27,11 @@ import { DBService } from './db-service';
...
 
 const observationCSVColumnValidator: IXLSXCSVValidator = {
-  columnNames: ['SPECIES', 'COUNT', 'DATE', 'TIME', 'LATITUDE', 'LONGITUDE'],
-  columnTypes: ['number', 'number', 'date', 'string', 'number', 'number'],
+  columnNames: ['ITIS_TSN', 'ITIS_SCIENTIFIC_NAME', 'COUNT', 'DATE', 'TIME', 'LATITUDE', 'LONGITUDE'],
+  columnTypes: ['number', 'string', 'number', 'date', 'string', 'number', 'number'],
   columnAliases: {
     LATITUDE: ['LAT'],
-    LONGITUDE: ['LON', 'LONG', 'LNG'],
-    SPECIES: ['TAXON']
```
curtisupshall, line 35: Given we now have two columns for tracking species, we should decide on new column aliases for `ITIS_TSN` and `ITIS_SCIENTIFIC_NAME`.

api/src/services/validation-service.ts
curtisupshall, line 1: Commented code and skipped tests for this service are being removed on another branch.

api/src/services/observation-service.ts
curtisupshall: As discussed, we will remove ITIS_SCIENTIFIC_NAME, and support the following aliases for ITIS_TSN:
```
species
taxon
tsn
itis_tsn
```

database/src/migrations/20240124141800_observation_scientific_name.ts
NickPhura: Right (remembering now). We only added scientific name to the other table to support sorting (and because it should be safe anyways). I think we opted not for this table in light of there being no strict requirement for it, yet.

Disregard my previous comment

NickPhura: Can you add a todo item to this page, so we don't forget to patch the TSNs when we release: https://apps.nrs.gov.bc.ca/int/confluence/display/TASHIS/SIMS+Releases

Edit: Apparently I created a bullet for this already in past, but I think our strategy has changed since then. So maybe just add/replace it with the latest plan, if its different.

KjartanE (dismissed):
🐱

NickPhura (approved):

KjartanE (approved):
🐈

PR #1212, opened 2024-02-14
SIMSBIOHUB-483: Paginate the Projects lists endpoint

curtisupshall:
## Overview
Paginates the projects list table, removes dead code, and optimizes code associated with projects.

## Links to Jira Tickets

- [SIMSBIOHUB-483](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-483)

## Description of Changes

- Adds new environment variables used to the set number of seed projects and surveys to generate:
  - `NUM_SEED_PROJECTS`
  - `NUM_SEED_SURVEYS_PER_PROJECT`
- Removes frontend code associated with creating, editing and deleting project drafts.
- Optimizes SQL queries used to fetch the projects list
- Removes the project submission status from the projects list response; Removes the project submission status banner
- Modifies the list projects endpoint to support server-side pagination

## Testing Notes

You should be able to:
 - [ ] Create several seed projects and seed surveys using the two new environment variables
 - [ ] Create a new project
 - [ ] Edit an existing project
 - [ ] List projects at varying page sizes
 - [ ] Sort the projects list


api/src/openapi/schemas/pagination.ts
```diff
@@ -0,0 +1,79 @@
...
+    name: 'sort',
+    required: false,
+    description: 'The column being sorted on'
+  },
+  {
+    in: 'query',
+    name: 'order',
+    required: false,
+    description: 'The order of the sort, i.e. asc or desc'
+  }
```
NickPhura: Should this include a schema and define that it must be a string? And ideally also specify an enum for 'asc' and 'desc' (if allowed in the spec).

api/src/openapi/schemas/pagination.ts
```diff
@@ -0,0 +1,79 @@
...
+ * API schema to assert pagination information for paginated data
+ * responses.
+ */
+export const paginationResponseSchema: SchemaObject = {
+  type: 'object',
+  required: ['total', 'current_page', 'last_page'],
+  properties: {
+    total: {
+      type: 'integer',
+      description: 'The total number of observation records belonging to the survey'
```
NickPhura: nit: is this description meant to be generic?

api/src/repositories/project-repository.ts
```diff
@@ -137,13 +137,77 @@ export class ProjectRepository extends BaseRepository {
...
+    sqlStatement.append(';');
+
+    const response = await this.connection.sql(sqlStatement, z.object({ project_count: z.string().transform(Number) }));
+
+    if (response?.rowCount < 1) {
+      throw new ApiExecuteSQLError('Failed to get project count', [
+        'ProjectRepository->getProjectCount',
+        'rows was null or undefined, expected rows != null'
+      ]);
+    }
```
NickPhura, line 206: We probably have to remove this, as a user could legitimately have no projects.

app/src/features/projects/create/CreateProjectPage.tsx
```diff
@@ -238,48 +131,13 @@ const CreateProjectPage: React.FC = () => {
     return true;
   };
 
-  const handleDeleteDraft = async () => {
-    await deleteDraft();
```
NickPhura, line 242: Are there draft endpoints/etc we can purge as well?

NickPhura: Though, it could obv be done in a separate cleanup ticket.

app/src/features/projects/list/ProjectsListPage.tsx
```diff
@@ -1,65 +1,191 @@
...
+  };
 
-  const draftsDataLoader = useDataLoader(() => biohubApi.draft.getDraftsList());
-  draftsDataLoader.load();
+  codesContext.codesDataLoader.load();
 
-  /**
-   * Handle filtering project results.
-   */
+  // TODO unnecesssary
```
NickPhura: nit: just in case

app/src/features/projects/create/CreateProjectPage.tsx
curtisupshall: Probably yeah. I did a non-exhaustive audit of the frontend for draft-related code, but skipped the backend. I'll address this

api/src/repositories/project-repository.ts
curtisupshall: This query is supposed to return exactly one row, which captures the project count. So if a user has no projects, the result of this query should be `[{ project_count: 0 }]`, hence the check for `response.rowCount`

api/src/openapi/schemas/pagination.ts
curtisupshall: Good catch, I've changed it.

api/src/openapi/schemas/pagination.ts
curtisupshall: Yeah good call. When I wrote this a while back, I didn't know how to add the schema. It's been added now!

api/src/repositories/project-repository.ts
NickPhura: Ah gotcha, fair

app/src/features/projects/create/CreateProjectPage.tsx
curtisupshall: Being addressed by https://github.com/bcgov/biohubbc/pull/1216.

env_config/env.docker
```diff
@@ -194,3 +194,10 @@ GCNOTIFY_SMS_URL=https://api.notification.canada.ca/v2/notifications/sms
 # project to match the given user identifier. If left blank, defaults to the
 # user with system user ID 1.
 PROJECT_SEEDER_USER_IDENTIFIER=
+
+# Sets the number of desired seed projects to generate. Defaults to 1.
+NUM_SEED_PROJECTS=500
+
+# Sets the number of desired seed surveys to generate per project.
+# Defaults to 1.
+NUM_SEED_SURVEYS_PER_PROJECT=60
```
NickPhura: I would change these defaults to something smaller, since the average person (especially if they aren't as savvy with these env files) probably doesn't want to seed this much stuff every time.

Maybe like 2 projects with 2 surveys each or something.

api/src/paths/project/list.ts
```diff
@@ -164,6 +157,13 @@ GET.apiDoc = {
  */
 export function getProjectList(): RequestHandler {
   return async (req, res) => {
+    defaultLog.debug({ label: 'getProjectList' });
+
+    const page: number | undefined = req.query.page ? Number(req.query.page) : undefined;
+    const limit: number | undefined = req.query.limit ? Number(req.query.limit) : undefined;
+    const order: 'asc' | 'desc' | undefined = req.query.order ? (String(req.query.order) as 'asc' | 'desc') : undefined;
```
NickPhura, line 165: The pagination fields aren't included in the openapi spec for this endpoint.

NickPhura:
Looks good. The project advanced filters are a bit messed up though (partially related to pagination, but also related to other changes we've made recently, like the taxon changes). But maybe we should handle that as part of a different ticket. We probably need to run it by Jeremy first, to see what filters are even relevant still, etc.

env_config/env.docker
curtisupshall: Oops, yeah this is from testing lol. Two and two sounds good

api/src/paths/project/list.ts
curtisupshall: Good catch - fixed.

curtisupshall:
> Looks good. The project advanced filters are a bit messed up though (partially related to pagination, but also related to other changes we've made recently, like the taxon changes). But maybe we should handle that as part of a different ticket. We probably need to run it by Jeremy first, to see what filters are even relevant still, etc.

Good call. I've [opened a ticket](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-492) to get the conversation started.

NickPhura (approved):
📃

al-rosenthal (approved):
ran this locally and it works as expected 🍰

PR #1213, opened 2024-02-14
SIMSBIOHUB-466: Add Observation Measurement Columns

NickPhura:
## Links to Jira Tickets

[SIMSBIOHUB-466](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-466)

#### Many updates to the Observation Table components and related components
- Support for adding, removing, hiding, and populating measurement columns.
- New database tables + related service/repo

When creating an observation record, a new subcount record is also created.

### Pending Critterbase updates

Wire up the stubbed calls to Critterbase, once its new endpoints are ready.

### TODOs for another ticket

Update the telemetry datagrid components/tables to use the latest updates from the survey observations components.

Fix the 3-4 app tests that started to fail (probably a side effect of package updates?) I tried fixed them but was making no progress, so they are skipped for now. They are all related to trying to capture select box clicks, which is historically tricky.

## Testing Notes

Observations table works as expected, when adding and removing observations with and without measurement columns.


api/src/paths/project/{projectId}/survey/{surveyId}/observations/index.ts
```diff
@@ -181,14 +91,329 @@ GET.apiDoc = {
...
+                    },
+                    itis_tsn: {
+                      type: 'integer'
+                    },
+                    itis_scientific_name: {
+                      type: 'string',
+                      nullable: true
+                    },
+                    survey_sample_site_id: {
+                      type: 'number',
```
curtisupshall: Nit: use `type: 'integer'`.

Might make sense to do a quick Ctrl+F for these.

image

api/src/paths/project/{projectId}/survey/{surveyId}/observations/index.ts
```diff
@@ -394,32 +676,15 @@ export function insertUpdateSurveyObservations(): RequestHandler {
...
-      );
+      const observationRows = req.body.surveyObservations;
 
-      const surveyObservations = await observationService.insertUpdateSurveyObservations(surveyId, records);
+      await observationService.insertUpdateSurveyObservationsWithMeasurements(surveyId, observationRows);
 
       await connection.commit();
 
-      return res.status(200).json({ surveyObservations });
+      return res.status(200).send();
```
curtisupshall: Nit: use `HTTP 204 No Content`?

api/src/repositories/observation-repository.ts
```diff
@@ -443,18 +476,20 @@ export class ObservationRepository extends BaseRepository {
...
    * @memberof ObservationRepository
    */
-  async deleteObservationsByIds(observationIds: number[]): Promise<number> {
+  async deleteObservationsByIds(surveyId: number, observationIds: number[]): Promise<number> {
     const queryBuilder = getKnex()
       .queryBuilder()
       .delete()
       .from('survey_observation')
       .whereIn('survey_observation_id', observationIds)
+      .andWhere('survey_id', surveyId)
```
curtisupshall, line 492: Any particular reason we need this constraint?

api/src/utils/media/xlsx/xlsx-file.ts
```diff
@@ -41,20 +71,20 @@ export class XLSXCSV {
   }
 
   /**
-   * Runs all content and workbook-related validation for this CSV file, based on the given validation
+   * Runs all content and _workbook-related validation for this CSV file, based on the given validation
```
curtisupshall: nit: replace `_workbook` with `workbook` in code comments

app/src/components/data-grid/taxonomy/TaxonomyDataGridViewCell.tsx
```diff
@@ -42,7 +53,7 @@ const TaxonomyDataGridViewCell = <DataGridType extends GridValidRowModel>(
         textOverflow: 'ellipsis',
         color: props.error ? 'error' : undefined
       }}>
-      {[response.commonName, `(${response.scientificName})`].filter(Boolean).join(' ')}
+      {[taxon.commonName, `(${taxon.scientificName})`].filter(Boolean).join(' ')}
```
curtisupshall: Nit: if no common name exists, do we still want to wrap the scientific name in parens? It may look out of place (or we may not care about this at all)

app/src/features/surveys/observations/measurements/dialog/MeasurementsDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+      maxWidth="xl"
+      open={open}
+      aria-labelledby="observation-measurements-dialog-title">
+      <DialogTitle id="observation-measurements-dialog-title">Add Observation Measurements</DialogTitle>
+      <DialogContent>
+        <MeasurementsSearch selectedMeasurements={selectedMeasurements} onSelectOptions={onSelectOptions} />
+        <MeasurementsList selectedMeasurements={selectedMeasurements} onRemove={onRemove} />
+      </DialogContent>
+      <DialogActions>
+        <LoadingButton
```
curtisupshall: Should this just be a plain `<Button>`, rather than a LoadingButton?

app/src/features/surveys/observations/measurements/list/MeasurementsListCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+      sx={{
+        background: grey[100]
+      }}
+      data-testid="measurements-list-item">
+      <Stack gap={0.75} mt={-0.25}>
+        <Box>
+          <Typography variant="body2">
+            <em>{props.measurement.itis_tsn}</em>
+          </Typography>
+          {/* <Typography variant="body2">
```
curtisupshall, line 53: Remove?

app/src/features/surveys/observations/measurements/search/MeasurementsSearchAutocomplete.tsx
```diff
@@ -0,0 +1,204 @@
...
+      data-testid="measurements-autocomplete"
+      noOptionsText="No matching options"
+      autoHighlight={true}
+      options={options}
+      multiple={true}
+      disableCloseOnSelect={true}
+      blurOnSelect={false}
+      clearOnBlur={false}
+      getOptionLabel={(option) => option.measurement_name}
+      //   getOptionLabel={(option) => option.commonName || option.scientificName}
```
curtisupshall: Nit: remove

app/src/features/surveys/observations/measurements/search/MeasurementsSearchAutocomplete.tsx
```diff
@@ -0,0 +1,204 @@
...
+              value={renderOption.taxon_measurement_id}
+              color="default"
+            />
+            <Stack gap={0.75} mt={-0.25}>
+              <Box>
+                <Typography variant="body2">
+                  <em>{renderOption.itis_tsn}</em>
+                </Typography>
+                {/* <Typography variant="body2">
+                  {renderOption.commonName ? (
```
curtisupshall, line 146: Remove?

app/src/features/surveys/observations/observations-table/configure-table/ConfigureColumnsContainer.tsx
```diff
@@ -0,0 +1,200 @@
...
+      return [...currentColumns, ...newColumns];
+    });
+  };
+
+  /**
+   * On first mount, load visibility state from session storage, if it exists.
+   */
+  useEffect(() => {
+    if (JSON.stringify(observationsTableContext.columnVisibilityModel) !== '{}') {
+      // TODO is there a better way to prevent this useEffect from running repeatedly after trying once to load from session storage?
```
curtisupshall: Could you elaborate on this issue?

al-rosenthal (commented):
Ran into a few issues when testing this locally:
Validation Banner doesn't show
- Hide required columns (like the sample sites)
- Add new row and fill
- Save (see error dialog)
- No banner

Hiding newly added columns
- Add new measurement
- New column cannot be toggled
- Close/ Reopen measurements panel
- Newly added measurement is still cannot be toggled

Deleting Columns from measurement popup
- Removing a single measurement from the list seems to work as expected
- Removing all added measurements seems to reload the panel and add columns back to list (might be data related)

app/src/constants/session-storage.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Get a session storage key which is unique to the provided survey id.
+ *
+ * @param {number} surveyId
+ * @param {string} sessionStorageKey
+ * @return {*}  {string}
+ */
+export const getSurveySessionStorageKey = (surveyId: number, sessionStorageKey: string): string => {
+  return `${sessionStorageKey}_${surveyId}`;
```
curtisupshall: Nit: I might prefix the survey ID with `SURVEY_`, just to distinguish it from other potential keys.

For instance, if we had a `getProjectSessionStorageKey` method:

```javascript
export const getProjectSessionStorageKey = (projectId: number, sessionStorageKey: string): string => {
  return `${sessionStorageKey}_${projectId}`;
}
const key1 = getSurveySessionStorageKey(mySurveyId, PREFERRED_COLUMN_ORDER);
const key2 = getProjectSessionStorageKey(myProjectId, PREFERRED_COLUMN_ORDER);
// key1 and key2 collide if surveyId and projectId match (plausible) and if I pass the same storage key (not smart, but also plausible)
```

api/src/repositories/observation-repository.ts
NickPhura: Its part of a larger tech debt update we need to do for security reasons. There is a ticket in the backlog for it I believe.

We authenticate our endpoints against either a project_id or survey_id. If we don't include a where clause against one of those, then someone could easily call our endpoints with a survey_id they do have access to, but pass a sample_site_id (for example) that they don't have access to, and the query would gladly delete it (or w/e the operation is).

We can consider adding more auth functions, to cover more than just project_id and survey_id, but the concept still holds the same: whatever id is used by the auth must also be used by the query.

app/src/features/surveys/observations/measurements/list/MeasurementsListCard.tsx
NickPhura: This is code that we do want, that Jer wrote, but which can't work just yet because the response from Critterbase is in flux. Ideally this, or some version of it, will be uncommented soon and the typography above will be removed.

api/src/repositories/observation-repository.ts
curtisupshall: Ahh, makes sense. Thanks for explaining!

NickPhura:
@al-rosenthal 
> Ran into a few issues when testing this locally: Validation Banner doesn't show

I think this is an existing issue. It seems that we only validate the sampling columns if at least one of them is populated, probably to do with the observation import maybe? The comments in that code talk about not validating sampling columns if its an incidental observation. 

> Hiding newly added columns

I fixed a few issues with the visibility controls. I think it should be happy now.

> Deleting Columns from measurement popup

Yeah this is a more interesting issue. The problem is that the measurement data is still part of the row data, and just with the way the useEffect is set up right now, when you remove all measurement columns it re-loads the columns from the existing data, hence those columns showing up again suddenly.  There should be a fix, but to date we only ever do row based deletes, so ill have to poke at it.

app/src/features/surveys/observations/observations-table/configure-table/ConfigureColumnsContainer.tsx
NickPhura: I just don't like the need to stringify the visibility model and compare it to `'{}'`. Can probably come up with something better, I just didnt go to far with it at the time.

app/src/features/surveys/observations/measurements/search/MeasurementsSearchAutocomplete.tsx
NickPhura: Removed.

app/src/constants/session-storage.ts
NickPhura: Fair point, added.

app/src/features/surveys/observations/measurements/dialog/MeasurementsDialog.tsx
NickPhura: Yup! Probably just copy-paste mistake. Fixed

api/src/paths/project/{projectId}/survey/{surveyId}/observations/index.ts
NickPhura: Sure, updated.

app/src/components/data-grid/taxonomy/TaxonomyDataGridViewCell.tsx
NickPhura: Yeah I just was doing this quickly and never went back. Fixed

api/src/utils/media/xlsx/xlsx-file.ts
NickPhura: Fixed

app/src/features/surveys/observations/measurements/search/MeasurementsSearchAutocomplete.tsx
NickPhura: As as above: 

This is code that we do want, that Jer wrote, but which can't work just yet because the response from Critterbase is in flux. Ideally this, or some version of it, will be uncommented soon and the typography above will be removed.

app/src/features/surveys/observations/measurements/search/MeasurementsSearchAutocomplete.tsx
NickPhura: Same as above:

This is code that we do want, that Jer wrote, but which can't work just yet because the response from Critterbase is in flux. Ideally this, or some version of it, will be uncommented soon and the typography above will be removed.

api/src/paths/project/{projectId}/survey/{surveyId}/observations/index.ts
NickPhura: Updated.

al-rosenthal (commented):
Retested this and saw something else:
- Add a new row
- Discard Changes
- Open Measurement Panel 
- Cannot add measurements (probably an issue with state)

NickPhura:
@al-rosenthal 
> Retested this and saw something else:

Should be fixed. Was not resetting the row modes model on discard

app/src/features/surveys/observations/measurements/search/MeasurementsSearchAutocomplete.tsx
curtisupshall: Understood

app/src/features/surveys/observations/measurements/dialog/MeasurementsDialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+import DialogContent from '@mui/material/DialogContent';
+import DialogTitle from '@mui/material/DialogTitle';
+import useTheme from '@mui/material/styles/useTheme';
+import useMediaQuery from '@mui/material/useMediaQuery';
+import { MeasurementsList } from 'features/surveys/observations/measurements/list/MeasurementsList';
+import { MeasurementsSearch } from 'features/surveys/observations/measurements/search/MeasurementsSearch';
+import { CBMeasurementType } from 'interfaces/useCritterApi.interface';
+import { useState } from 'react';
+
+export interface IrementsDialogProps {
```
curtisupshall: Typo!

curtisupshall: Found seven instances of 'irements'

app/src/features/surveys/observations/observations-table/configure-table/ConfigureColumnsContainer.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+      return [...currentColumns, ...newColumns];
+    });
+  };
+
+  /**
+   * On first mount, load visibility state from session storage, if it exists.
+   */
+  useEffect(() => {
+    if (JSON.stringify(observationsTableContext.columnVisibilityModel) !== '{}') {
```
curtisupshall: Still looking for an alternative here? Would this work?
```javascript
if (Object.keys(observationsTableContext.columnVisibilityModel).length) {
  // ...
}
```

app/src/features/surveys/observations/measurements/dialog/MeasurementsDialog.tsx
NickPhura: haha, fixed

app/src/features/surveys/observations/observations-table/configure-table/ConfigureColumnsContainer.tsx
NickPhura: Yeah thats better, good call

curtisupshall (dismissed):
Looks good to me! 📏

al-rosenthal (dismissed):
🍾

curtisupshall (approved):

al-rosenthal (approved):
⚒️

PR #1219, opened 2024-02-23
SIMSBIOHUB-79: Feature Flag Guard

NickPhura:
## Links to Jira Tickets

(must have been deleted or removed)

## Description of Changes

Adds a new env var FEATURE_FLAGS which is a general purpose flag variable that can be used to disable frontend or backend features as needed.

### APP
Adds a new type of guard that uses the FEATURE_FLAGS vars to disable child components.

The guard can be used to wrap components, which will be rendered as normal unless feature flags are specified and at least one matching flag is defined in the associated env var.

Replaced the `REACT_APP_BIOHUB_FEATURE_FLAG` with a new feature flag: `APP_FF_SUBMIT_BIOHUB`

### API
Included the feature flags in the api pipeline/docker-compose as well, and added some helper util functions for working with the feature flag env var. 

Replaced the `BACKBONE_INTAKE_ENABLED` with a new feature flag: `API_FF_SUBMIT_BIOHUB`

## Testing Notes

Ex: If you set the FEATURE_FLAGS var in `.env` to `FEATURE_FLAGS=APP_FF_SUBMIT_BIOHUB` the publish button should be hidden from the UI.


MacDeluca (approved):
S H I P   I T

PR #1223, opened 2024-02-28
Add draw controls to the sample site map

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Added FeatureLayer and DrawControls to the Sample Site create and edit map

## Testing Notes

- Should be able to create and replace sampling site locations using drawing controls
- Should be able to create points, lines, or polygons
- You can only create one site per create request (can't draw multiple sites at once and bulk-create)


KjartanE (approved):
Draw control function properly

curtisupshall (approved):
Tested locally, works well 🗺️

PR #1230, opened 2024-03-05
Multiple common names & including taxonomic rank on the species card

mauberti-bc:
## Links to Jira Tickets

None

## Description of Changes

- Replace all references to commonName with commonNames
- Change commonName type from a string to an array of strings
- Include taxonomic rank in the Species Card to indicate whether taxon is a species, genus, family, etc.
- Include kingdom and rank in biohub's taxonomy API response

## Testing Notes
- *Must be run locally with this BioHub PR: https://github.com/bcgov/biohubbc-platform/pull/237*
- Species can have multiple common names
- Species card should always include the rank label

Will need to be merged with https://github.com/bcgov/biohubbc-platform/pull/237


database/src/migrations/release.0.34/api_get_eml_data_package.sql
```diff
@@ -294,7 +294,7 @@ begin
         <taxonomicClassification>
 				  <taxonRankName>%1$s</taxonRankName>
 				  <taxonRankValue>%2$s</taxonRankValue>
-				  <commonName>%3$s</commonName>
+				  <commonNames>%3$s</commonNames>
```
NickPhura: Looks like this just got caught in the find/replace.

Since this is an old migration, we should undo this change.

mauberti-bc: Oops, I'll make sure to exclude migrations when find-replacing. Thanks

app/src/components/data-grid/autocomplete/AsyncAutocompleteDataGridEditCell.tsx
```diff
@@ -190,8 +196,24 @@ const AsyncAutocompleteDataGridEditCell = <DataGridType extends GridValidRowMode
       )}
       renderOption={(renderProps, renderOption) => {
         return (
-          <Box component="li" {...renderProps}>
-            {renderOption.label}
+          <Box
```
al-rosenthal, line 205: this should probably be swapped out for: `<List>` and `<ListItem>` components

app/src/components/species/components/SpeciesCard.tsx
```diff
@@ -1,45 +1,93 @@
...
-  commonName: string | null;
+  commonNames: string[];
   scientificName: string;
   tsn: number;
+  rank: string;
+  kingdom: string;
 }
 
 const SpeciesCard = (props: ISpeciesCard) => {
+  const rankColors = [
```
al-rosenthal: nit: this could probably be placed in the constants folder and imported to be used here. It makes it easier to use in other places in code while only have a single source of truth to maintain

al-rosenthal (commented):
I tried running this locally and ran into issues viewing the survey page. Looks like an issue with the open api spec on the view endpoint. It is missing the `commonNames` property

mauberti-bc:
> I tried running this locally and ran into issues viewing the survey page. Looks like an issue with the open api spec on the view endpoint. It is missing the `commonNames` property

I should have mentioned that this PR needs to be tested with this BioHub branch: https://github.com/bcgov/biohubbc-platform/pull/237. That fixes the issue for me.

NickPhura (approved):

PR #1235, opened 2024-03-08
SIMSBIOHUB-492: Fix project advanced filters

curtisupshall:
## Links to Jira Tickets

 - [SIMSBIOHUB-492](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-492)

## Description of Changes

- Fixes the projects list advanced filters to support:
  - Paginated results with advanced filters applied, and;
  - ITIS TSN searching.
- Added support to search on multiple species;
- Used MUI Collapse to show/hide advanced filters form;
- Code cleanup, minor tech debt.

## Testing Notes

You should change `NUM_SEED_PROJECTS` in your `.env` file to make testing easier.

- [ ] You should be able to filter projects on any and all of the advanced filter fields
- [ ] Pagination should work on advanced-filtered results.


api/src/models/project-view.ts
```diff
@@ -4,13 +4,12 @@ import { ProjectUser } from '../repositories/project-participation-repository';
...
 
 export interface IProjectAdvancedFilters {
-  permit_number?: string;
   project_programs?: number[];
   start_date?: string;
   end_date?: string;
   keyword?: string;
   project_name?: string;
-  species?: number[];
+  species_tsns?: number[];
```
NickPhura: nit: can we call this `itis_tsns`?

Not sure if we are 100% consistent, but for the most part we've been using `itis_*`

api/src/repositories/project-repository.ts
```diff
@@ -28,185 +29,146 @@ const defaultLog = getLogger('repositories/project-repository');
...
-        p.project_id,
-        p.name,
-        p.start_date,
-        p.end_date
-    `);
+    // Programs filter
+    if (filterFields.project_programs?.length) {
+      query.havingRaw(
+        `array_agg(DISTINCT prog.program_id) && ARRAY[${filterFields.project_programs.join(',')}]::integer[]`
+      );
```
NickPhura: I think this is more ideal, since `having` is less performant. I tried it locally and it seems to be working, maybe just double check in case:

```suggestion
      query.where('prog.program_id', 'IN', filterFields.project_programs);
```

NickPhura: (I've noticed in past as well, that knex `whereIn` doesn't seem to work due to some typing issues. But if you just use `where` + `IN` then it works fine, and still recognizes it has requiring an array for the third param)

api/src/models/project-view.ts
curtisupshall: Good call.

api/src/repositories/project-repository.ts
curtisupshall: > (I've noticed in past as well, that knex `whereIn` doesn't seem to work due to some typing issues. But if you just use `where` + `IN` then it works fine, and still recognizes it has requiring an array for the third param)

ahh that's interesting. I tried `.whereIn` and it didn't work (ostensibly for the same reason you mentioned). I will change it

NickPhura (approved):
🔍

al-rosenthal (approved):
ran this locally and things looks good

PR #1237, opened 2024-03-11
Landing page & project page UI changes

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Add new FAQ section to landing page, and update text
- Show project members on the project page

## Testing Notes

- All project members should be listed on the project page, with an icon indicating their role


api/package-lock.json
curtisupshall, line 1: Probably unintentional, but we don't want to delete this. Try:
```
git checkout dev -- api/package.json
```

app/src/components/data-grid/autocomplete/AsyncAutocompleteDataGridEditCell.tsx
```diff
@@ -1,4 +1,4 @@
-import Autocomplete from '@mui/material/Autocomplete';
```
curtisupshall: Nit: should revert this change. Absolute imports are preferred

api/package.json
```diff
@@ -63,7 +63,8 @@
     "winston": "~3.3.3",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.19.3/xlsx-0.19.3.tgz",
     "xml2js": "~0.4.23",
-    "zod": "~3.21.4"
+    "zod": "~3.21.4",
+    "logform": "2.3.2"
```
curtisupshall, line 67: This PR adds `logform` but it's not being used

curtisupshall: As discussed, installed to fix a dependancy issue.

app/src/features/projects/view/components/TeamMember.tsx
```diff
@@ -0,0 +1,97 @@
...
+    return `#${randomComponent()}${randomComponent()}${randomComponent()}`;
+  }
+
+  return (
+    <Stack spacing={1}>
+      {projectTeamMembers.map((member, index) => (
+        <Box display="flex" alignItems="center" key={member.display_name}>
+          <Box
+            sx={{ height: '35px', width: '35px', minWidth: '35px', borderRadius: '50%' }}
+            bgcolor={getRandomHexColor(index)}
```
curtisupshall: I would not use `index` as the seed for the user's color. If I remove a user from the top of the list, then all the users' avatar colors will change. Although the actual color doesn't matter, I think users would expect the color for a particular user to stay the same. 

Here is a function that could convert a user's name into a seed. This means the color they are assigned is still random, but will remain consistent across all projects.

```typescript
const usernameToSeed = (username: string): number => {
  let hash = 0;
  for (let i = 0; i < username.length; i++) {
    const char = username.charCodeAt(i);
    hash = (hash << 5) - hash + char;
    hash = hash & hash; // Convert to 32bit integer
  }
  return hash;
}
```

app/src/features/projects/view/components/TeamMember.tsx
```diff
@@ -0,0 +1,97 @@
...
+  assert(projectContext.projectDataLoader.data);
+
+  // Define a custom sorting order for roles
+  const roleOrder: { [key: string]: number } = {
+    Coordinator: 1,
+    Collaborator: 2,
+    Observer: 3
+  };
+
+  const projectTeamMembers: IProjectParticipantsRoles[] = projectContext.projectDataLoader.data.projectData.participants
```
curtisupshall: This should be memoized. To that end, we may as well also calculate the user's avatar color (see my other comment) in here as well.
```javascript
const projectTeamMembers: (IProjectParticipantsRole & { avatarColor: string })[] = useMemo(() => {
   // Logic to get their initials, sort the list, etc. goes here.
}, [projectContext.projectDataLoader.data.projectData.participants]);
```

app/src/features/projects/view/components/TeamMember.tsx
```diff
@@ -0,0 +1,97 @@
...
+    .map((member) => {
+      const initials = member.display_name
+        .split(',')
+        .map((name) => name.trim().slice(0, 1).toUpperCase())
+        .reverse()
+        .join('');
+      return { display_name: member.display_name, roles: member.project_role_names, initials };
```
curtisupshall: If we're memoizing this call, let's also return the user's avatar color in this object.

app/src/features/projects/view/components/TeamMember.tsx
```diff
@@ -0,0 +1,107 @@
...
+                path={mdiCrown}
+                size={0.75}
+                style={{ marginLeft: '8px' }}
+              />
+            )}
+            {member.roles.includes('Collaborator') && (
+              <Icon
+                title={member.roles.join(', ')}
+                color="gray"
+                path={mdiPencilOutline}
```
al-rosenthal: We use the pencil icon in a few places to indicate that we are can make edits, so having that icon next to the user names made me think I could click and edit the users. I'd suggest maybe the account-group icon? it looks like it might fit with the 'Collaborator'

curtisupshall (commented):
Code looks good! I committed a minor change. Once the app builds, I will test this in Openshift and approve.

al-rosenthal (dismissed):
⛵

curtisupshall (dismissed):
Looks good to me!

app/src/pages/landing/LandingIntro.tsx
```diff
@@ -0,0 +1,62 @@
+import { Box, Grid } from '@mui/material';
+import { QuestionAnswer } from './components/QuestionAnswer';
+
+export const LandingIntro = () => {
+  return (
+    <>
```
NickPhura: super nit: this fragment isn't needed, since it only has one child element

app/src/features/projects/view/components/TeamMember.tsx
```diff
@@ -0,0 +1,100 @@
...
+        })
+        .sort((a, b) => {
+          const roleA = a.roles[0] || '';
+          const roleB = b.roles[0] || '';
+          return roleOrder[roleA] - roleOrder[roleB];
+        }) || [],
+    [projectContext.projectDataLoader.data.projectData.participants]
+  );
+
+  function getRandomHexColor(seed: number) {
```
NickPhura: Not super necessary, but you could consider moving this into `utils/Utils.tsx`, which just has a collection of standalone functions.

Might be the tiniest bit more efficient, I think? Because it wont have to re-define the function on each re-render. Though realistically very negligible.

NickPhura (dismissed):
One small nit, though its not required.

curtisupshall (dismissed):
LGTM

app/src/features/projects/view/components/TeamMember.tsx
curtisupshall: Moved.

curtisupshall (approved):

NickPhura (approved):

PR #1242, opened 2024-03-14
Fix survey progress breaking dev

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Update any existing Survey data to have a progress_id value

## Testing Notes

- Should set progress_id in any existing survey data


al-rosenthal (approved):
🐕

curtisupshall (approved):
LGTM

NickPhura:
@mauberti-bc Looks like the static deploy still failed because the migration failed.

`column "progress_id" contains null values`

I am not entirely sure why, but the only thing left I can think of is maybe the foreign key constraint can't be added until after the progress_id is patched. It probably doesn't like the fact that the foreign table column has a not null constraint, but the source table does not.

PR #1244, opened 2024-03-15
SIMSBIOHUB-530: Unique Key Constraint on Survey Funding Source

KjartanE:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-530?

## Description of Changes

- added Unique Key Constraint on Survey Funding Source table
- create yup schema validation for no duplicate funding source

## Testing Notes

### DIFF
--- a/database/src/migrations/20240315000000_update_funding_source_table.ts
+++ b/database/src/migrations/20240315000000_update_funding_source_table.ts
@@ -1,0 +1,32 @@
+import { Knex } from 'knex';
+
+/**
+ * Added unique key constraint to survey funding source table
+ *
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
+
+    SET SEARCH_PATH=biohub;
+
+    ----------------------------------------------------------------------------------------
+    -- Add unique key constraints
+    ----------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


al-rosenthal (commented):
ran this locally and it works. It would be great if the funding sources were also highlighted in red, the error text is easy to miss if you have to scroll through the form to find the issue

database/src/migrations/20240315000000_update_funding_source_table.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
+
+    SET SEARCH_PATH=biohub;
+
+    ----------------------------------------------------------------------------------------
+    -- Add unique key constraints
+    ----------------------------------------------------------------------------------------
+    CREATE UNIQUE INDEX survey_funding_source_uk1 ON survey_funding_source(funding_source_id, survey_id);
```
NickPhura, line 18: Technically we would need to account for possible duplicate funding data in prod, or else this would fail.

That being said, I checked prod, and there doesn't seem to be any duplicate funding sources, so its probably fine for now.

NickPhura (approved):

curtisupshall (approved):

NickPhura:
> ran this locally and it works. It would be great if the funding sources were also highlighted in red, the error text is easy to miss if you have to scroll through the form to find the issue

Agreed. Though it looks like this matches the current behaviour of permits (which has the same unique constraint), so maybe something we can leave for a future ticket

PR #1245, opened 2024-03-15
SIMSBIOHUB-544: Update project/survey start/end date constraints

NickPhura:
Update associated yup schema.

## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-544

## Description of Changes

Update project and survey date triggers to remove restrictions on survey dates being between project dates.
Update associated yup schema.

Why? The triggers were not complete (there were cases they didn't handle correctly) + the project dates are going to be deprecated in future. So this resolves the trigger issue (by removing the constraints altogether), and moves a few steps closer to the eventual goal of removing the project dates.

## Testing Notes

Project start date must always come before its own end date, as before.

Survey start date must always come before its own end date, as before, but now has no constraints related to the project dates. So a survey could start before a project, or start after the project has ended, etc.

### DIFF
--- a/database/src/procedures/delete_project_procedure.ts
+++ b/database/src/procedures/delete_project_procedure.ts
@@ -1,5 +1,13 @@
 import { Knex } from 'knex';
 
+/**
+ * Inserts a procedure that makes all of the necessary deletions when a project is deleted (deleting all child records
+ * before deleting the project record itself).
+ *
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
 export async function seed(knex: Knex): Promise<void> {
   await knex.raw(`--sql
     SET search_path = 'biohub';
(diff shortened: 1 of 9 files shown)


database/src/procedures/delete_survey_procedure.ts
NickPhura, line 1: I just ran the vscode 'Trim Trailing Whitespace' command to satisfy the `--sql` template extension.

database/src/procedures/delete_project_procedure.ts
NickPhura, line 1: I just ran the vscode 'Trim Trailing Whitespace' command to satisfy the `--sql` template extension.

al-rosenthal (commented):
the new trigger works but I noticed that the survey end date component is still warning that `End Date cannot come after the Project End Date`

mauberti-bc (approved):
Created Projects and Surveys with non-overlapping dates 👍

curtisupshall (approved):
Code looks good to me

PR #1251, opened 2024-03-19
SIMSBIOHUB-529: Show Error Message If Project List Fails to Load

curtisupshall:
## Links to Jira Tickets

- [SIMSBIOHUB-529](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-529)

## Description of Changes

 - Adds an error dialog whenever the _Projects List_ or _Project Details_ page fails to load.

## Testing Notes

You should see an error dialog when the request fails while attempting to:
 - [ ] View the projects list page
 - [ ] View the project page for a single project

Multiple ways to test this locally:
1. Once you've loaded the project page/project list page, hot-reload the API, and then navitgate to the project page or project list page while the API is compiling. You should see the error dialog.
2. Throw an exception in the project view endpoint or project list endpoint. You should see the error dialog when loading either of the respective pages.

### DIFF
--- a/api/src/paths/project/{projectId}/view.ts
+++ b/api/src/paths/project/{projectId}/view.ts
@@ -191,7 +191,6 @@ GET.apiDoc = {
  */
 export function viewProject(): RequestHandler {
   return async (req, res) => {
-    throw new Error('This is a sample error for testing purposes and should not be committed.');
     const connection = getDBConnection(req['keycloak_token']);
 
     try {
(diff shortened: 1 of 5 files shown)


NickPhura:
@curtisupshall Sorry, thought I had this in the ticket: You can use the `useDataLoaderError` hook to encapsulate most of the code to handle those 2 error cases. It just ties into the dataloader `error` field and spawns a dialog for you, without needing any of the extra context setup or try/catch logic, etc.

KjartanE (dismissed):
Tested locally, works as intended

NickPhura (approved):

al-rosenthal (approved):
forced the api to throw errors and saw both dialogs! 🪘

PR #1253, opened 2024-03-19
Add Response Metric field to Sampling Method

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Added new "Response Metric" field to Sampling Methods
- Updated Method form to include Response Metric dropdown
- Updated the method lookup values
- Added description field to method lookup table

## Testing Notes

- Should be able to add and edit the Response Metric for Sampling Methods
- Should be blocked from adding a Sampling Method without a Response Metric


api/src/repositories/sample-period-repository.ts
mauberti-bc, line 1: There is no survey ID in these tables, so need to do some joins to prevent users from affecting methods that they don't have access to.

api/src/repositories/code-repository.ts
```diff
@@ -455,4 +456,27 @@ export class CodeRepository extends BaseRepository {
...
     return response.rows;
   }
+
+  /**
+   * Fetch method response metrics
+   *
+   * @return {ICodeWithDescription}
+   * @memberof CodeRepository
+   */
+  async getMethodResponseMetrics() {
```
curtisupshall: Needs typescript types

api/src/repositories/code-repository.ts
```diff
@@ -455,4 +456,27 @@ export class CodeRepository extends BaseRepository {
...
+        method_response_metric_id as id,
+        name,
+        description
+      FROM
+        method_response_metric
+      WHERE
+        record_end_date is null;
+    `;
+
+    const response = await this.connection.sql(sqlStatement);
```
curtisupshall: Could use Zod schema here.

database/src/migrations/20240310180000_lookup_options.ts
curtisupshall, line 1: Could probably just combine this migration into the other one

api/src/repositories/code-repository.ts
curtisupshall:
```javascript
async getMethodResponseMetrics(): Promise<ICodeWithDescription> {
...
}
```

mauberti-bc: I think this is resolved by adding the Zod schema

curtisupshall (dismissed):
Tested locally. Looks good to me!

api/src/repositories/code-repository.ts
NickPhura: Adding on late: Often its desirable to specify the return type at the function level as well, so that if someone were to inadvertently alter the `return`, the function would complain.

The `return` is the current return type, but not necessarily the intended return type.

curtisupshall: I'll add this

api/src/repositories/sample-period-repository.ts
```diff
@@ -183,7 +184,7 @@ export class SamplePeriodRepository extends BaseRepository {
       ON
         ssm.survey_sample_site_id = sss.survey_sample_site_id
       WHERE
-        ssp.survey_sample_period_id = ${surveySamplePeriodId}
+        ssp.survey_sample_period_id = ${surveySamplePeriodId};
```
al-rosenthal: the `;` here means the `AND sss.survey_id = ${surveyId}` won't actually be included in the sql.

al-rosenthal (commented):
I create a sampling site in dev, then switched to this branch and the migration fails. The message refers to a foreign key constraint violation on the `survey_sample_method` table.

NickPhura:
UI nit: The method card doesn't show the value of the response metric. It doesn't necessarily have to, but it does show every other field currently (method name, description, period(s))

image

Seems to work as expected though otherwise.

mauberti-bc:
> UI nit: The method card doesn't show the value of the response metric. It doesn't necessarily have to, but it does show every other field currently (method name, description, period(s))
> 
> image
> 
> Seems to work as expected though otherwise.

The method card is updated on another branch that I'll push up. I had a large branch and split it into two to make it easier to review, so that fix is coming. Good catch

api/src/repositories/sample-period-repository.ts
mauberti-bc: Oops, thanks for catching that.

curtisupshall: Fixed

curtisupshall (approved):
LGTM

NickPhura (approved):
🐟

NickPhura (approved):
😮‍💨

PR #1257, opened 2024-03-20
SIMSBIOHUB-470: Observation Table Measurement Validation

al-rosenthal:
## Links to Jira Tickets

[SIMSBIOHUB-470](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-470)

## Description of Changes

- Added front end validation for measurement columns in the observations table
- Refactored validation process and moved files into own validation utils file
- Added measurement validation at API entry level for observation measurements as well
- Fixed bugs in import/ measurement validation process

### Setup
- Run Critterbase locally from branch `SIMSBIOHUB-498`
- In SIMS, navigate to the `Manage Observations` table on a survey 
- Add records for Moose (Alces Americanus) and Black capped Chickadee (Poecile atricapillus)
- Save those records 
- Open table Cog menu and add measurements
- Search for antler and add all measurements found
- Search for nest and add all measurements found

### Testing
- Add antler values to the chickadee rows 
- Save changes 
- See validation errors for each cell that has antler data

- Add nest values to Moose rows
- Save Changes
- See validation errors for each cell that has nest data

- Add nest data to the chickadee rows
- Add antler data to the moose rows
- Save changes
- See table refresh with saved data


NickPhura:
Messing around with it a bit, seems to work so far. I do need to try it again against some better test data though, was just testing against critterbase dev which doesnt have the best measurements.

There was a todo comment about removing `colDef` from the measurements state. Maybe we can sneak that change in too, if we can in fact remove it everywhere.

al-rosenthal:
> Messing around with it a bit, seems to work so far. I do need to try it again against some better test data though, was just testing against critterbase dev which doesnt have the best measurements.
> 
> There was a todo comment about removing `colDef` from the measurements state. Maybe we can sneak that change in too, if we can in fact remove it everywhere.

I removed the colDef and got things working then took it a step further and just removed the MeasurementColumn type, it was only holding a reference the underlying measurement definition so it seemed silly to keep that as well

api/src/utils/xlsx-utils/worksheet-utils.ts
```diff
@@ -499,27 +509,39 @@ export async function getCBMeasurementsFromWorksheet(
...
   try {
-    for (const row of rows) {
-      const tsn = String(row['ITIS_TSN'] ?? row['TSN'] ?? row['TAXON'] ?? row['SPECIES']);
+    for (const tsn of tsns) {
       if (!tsnMeasurements[tsn]) {
-        // TODO: modify Critter Base to accept multiple TSN at once
         const measurements = await critterBaseService.getTaxonMeasurements(tsn);
         if (!measurements) {
-          // TODO: Any data for a TSN that has no measurements is invalid and should reject the uploaded CSV
+          throw `No measurements found for tsn: ${tsn}`;
```
curtisupshall: I would probably only throw an instance of `Error` rather than a string (even though this is technically OK), mostly because a caller that might catch this exception is most likely expecting an instance of `Error`.
```javascript
throw new Error(`No measurements found for tsn: ${tsn}`);
```

api/src/utils/xlsx-utils/worksheet-utils.ts
```diff
@@ -499,27 +509,39 @@ export async function getCBMeasurementsFromWorksheet(
...
         }
 
         tsnMeasurements[tsn] = measurements;
       }
     }
   } catch (error) {
```
curtisupshall, line 541: Assuming you accept the above comment, you can type anotate this, and refactor the throw on Line 543:
```javascript
} catch (error: Error) {
  throw new ApiGeneralError(`Error connecting to the Critterbase API: ${error.message}`);
}
```

app/src/constants/i18n.ts
```diff
@@ -315,7 +315,12 @@ export const ObservationsTableI18N = {
   // Import observation records
   importRecordsSuccessSnackbarMessage: 'Observations imported successfully.',
   importRecordsErrorDialogTitle: 'Error Importing Observation Records',
-  importRecordsErrorDialogText: 'An error occurred while importing observation records.'
+  importRecordsErrorDialogText: 'An error occurred while importing observation records.',
+
+  // Fetching TSN Measurements from CritterBase error
+  fetchingTSNMeasurementErrorDialogTitle: 'Error fetching measurement validation',
+  fetchingTSNMeasurementErrorDialogText:
+    'We ran into an error fetching measurement data from Critterbase, taxon selected may not be supported. Please try again. If the error persists, please contact your system administrator.'
```
curtisupshall: Might rephrase this to appear less colloquial:
> An error occurred while fetching measurement data from Critterbase. The selected taxon may not be supported. Please try again. If the error persists, please contact your system administrator.

app/src/features/surveys/observations/observations-table/observation-row-validation/ObservationRowValidationUtils.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    if (minValue === null && maxValue === null) {
+      return null;
+    }
+  }
+
+  // Measurement values are invalid, create an error and return
+  return {
+    field,
+    message: `Value provided is outside of the valid range ${minValue} < ${value} < ${maxValue}`
```
curtisupshall: Technically the `minValue` and `maxValue` are allowed, so the range is represented by `${minValue} ≤ ${value} ≤ ${maxValue}`

curtisupshall: (Opinion) I probably wouldn't use an inequality to represent the range either way, since if this validation error occurs, the value  of `${value}` wouldn't be within the range, so the inequality wouldn't hold. I would just use [square brackets](https://en.wikipedia.org/wiki/Interval_(mathematics))
```javascript
message: `Value provided is outside of the valid range [${minValue}, ${maxValue}]`
```

api/src/utils/xlsx-utils/worksheet-utils.ts
NickPhura: +1

api/src/utils/xlsx-utils/worksheet-utils.ts
al-rosenthal: the catch error cannot be typed like that, the error value is `any` or `unknown` and there is no guarantee of that the thing thrown is of `Error` type. In a catch like that you'd need do a check like 
`if (error instanceof Error)` to enforce a type on error

app/src/constants/i18n.ts
al-rosenthal: updated

app/src/features/surveys/observations/observations-table/observation-row-validation/ObservationRowValidationUtils.ts
al-rosenthal: yeah good point, updated

api/src/utils/xlsx-utils/worksheet-utils.ts
curtisupshall: That's fair. Here's a snippet from `KeycloakService` that you might  find inspiration from:
```javascript
    } catch (error) {
      defaultLog.debug({ label: 'getUserByUsername', message: 'error', error: error });
      throw new ApiGeneralError('Failed to get user info from keycloak', [(error as Error).message]);   
    }
```

curtisupshall: Currently it looks like we're just including the entire `error` object in the `ApiGeneralError` constructor, which is fine by me. Feel free to ignore the above comment

curtisupshall (approved):
Tested locally, LGTM

api/src/utils/xlsx-utils/worksheet-utils.ts
NickPhura: Never fully built out the custom error classes, though I think in an ideal world we would always throw some kind of custom known error, just to ensure consistency, and making it more obvious to upstream code what errors could be expected, and what fields are present on the error (as far as labels, messages, error details, etc). I think eventually it would also be nice to have more of a developer error message vs public error message, and maybe even supporting a custom stack concept, so that we can trace an error back to its source even if it gets captured and re-thrown.

NickPhura (approved):
🐦

PR #1264, opened 2024-03-27
SIMSBIOHUB-536: Refactored UI for downloading observation CSV template

curtisupshall:
## Description of Changes

- Adds the "Download CSV Observation Template" button as an iconbutton on the Manage Observations page.
- When duplicate permit numbers are entered in the edit survey page, the permit number input renders the error on each form field.
- When duplicate funding sources are selected in the edit survey page, the funding source select component renders the error on each form field.
- When a survey participant's job has not been entered, only the participant job select component is styled with an error border.


al-rosenthal (commented):
Ran this locally and noticed 2 things:
- Unrelated to your changes but I cannot create/ edit a survey, the open api spec is complaining about an additional 'kingdom' property 
- It would be nice to have a tooltip for the template button

KjartanE:
> Ran this locally and noticed 2 things:
> 
>     * Unrelated to your changes but I cannot create/ edit a survey, the open api spec is complaining about an additional 'kingdom' property
> 
>     * It would be nice to have a tooltip for the template button

Tested Locally, works great! Would agree with Alfred, the button looks like a download button for the observation list. Rather then a download button for a template.

KjartanE (approved):
Looks good to me!

al-rosenthal (approved):
🦤

PR #1267, opened 2024-04-08
CritterbaseProxy Fix

MacDeluca:


NickPhura (approved):
Was able to search for and add a measurement to an observation and save/load it successfully.

PR #1273, opened 2024-04-17
SIMSBIOHUB-562: Export Survey Data

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-562

### API
New endpoint that fetches and uploads survey data to s3, and returns a pre-signed url to download it.

### APP
New dialog (under survey settings dropdown button) that allows users to execute an export. Currently restricted to project coordinators and collaborators (and system/data admins).

## Testing Notes

Pre-req: S3 credentials in `.env`

From the settings button at the top of the survey page, clicking export pops a dialogue, which lets you choose which data to export. Data is exported to a zip file in S3, and immediately downloaded to the users local machine. 

### API
Currently the export only supports survey metadata, observations (which both just make a very basic select query, and leave the response as json), and telemetry (which returns the same data as the summary telemetry page).

### APP
The dialogue is fairly basic, with no loading bars, etc, for longer exports. It also just auto-downloads the file, which is maybe not what we want if its large. Might be better to just show a link, and let the user trigger it in their browser separately.

## Note
Export button is in the dropdown on the survey page.
image


api/src/services/export-services/export-service.ts
```diff
@@ -0,0 +1,145 @@
...
+      archiveStream.pipe(s3UploadStream);
+
+      // Start the S3 upload
+      const uploadPromise = uploadStreamToS3(s3UploadStream, EXPORT_ARCHIVE_MIME_TYPE, exportConfig.s3Key);
+
+      await Promise.all(
+        exportConfig.exportStrategies.map(async (exportStrategy) => {
+          const exportStrategyConfig = await exportStrategy.getExportStrategyConfig(this.connection);
+
+          defaultLog.silly({ label: 'export', message: 'exportStrategyConfig', exportStrategyConfig });
```
mauberti-bc: Is silly intentional here?

NickPhura: Yeah, though its questionable if its needed. I could probably remove it as it probably doesnt add any real value.

NickPhura: Its easy enough to know what the `exportStrategyConfig` is, so logging it probably doesnt help

NickPhura: removed

api/src/services/export-services/export-service.ts
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * @param {ExportDataConfig} config
+   * @return {*}  {Promise<string[]>}
+   * @memberof ExportService
+   */
+  async export(exportConfig: ExportConfig): Promise<string[]> {
+    defaultLog.debug({ label: 'export', message: 'exportConfig', exportConfig });
+
+    if (exportConfig.exportStrategies.length === 0) {
+      throw new Error('No export strategies have been.');
```
mauberti-bc: No export strategies have been defined?

NickPhura: Oh haha, fixed

NickPhura:
Added some unit tests

mauberti-bc (approved):
👍

PR #1274, opened 2024-04-18
BugFix: Sample Method Edit Bug

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Fixes an issue which prevented sample methods from being edited (due to an extra field being sent in the request, which failed openapi validation).

## Testing Notes

Sample methods (especially with multiple under one sample site) can be created/edited without issue.

### DIFF
--- a/app/src/features/surveys/components/SamplingMethodForm.tsx
+++ b/app/src/features/surveys/components/SamplingMethodForm.tsx
@@ -21,6 +21,7 @@ import MenuItem from '@mui/material/MenuItem';
 import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
 import { CodesContext } from 'contexts/codesContext';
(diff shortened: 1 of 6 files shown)


app/src/features/surveys/components/SamplingMethodForm.tsx
```diff
@@ -44,7 +44,7 @@ const SamplingMethodForm = () => {
 
   const handleMenuClick = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>, index: number) => {
     setAnchorEl(event.currentTarget);
-    setEditData({ ...values.methods[index], index });
```
NickPhura, line 47: This is the root issue basically. The index needs to be tracked, but originally it was being added into the methods data object, which then meant it was sent to the API.

Now, we still track it, but in a separate property of a wrapping object.

MacDeluca (approved):

PR #1277, opened 2024-04-19
BugFix: Fixes From QA In Test

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-541

## Description of Changes

Fix migration (patching missing subcount records for existing observation records). 
- Was causing observation records not to show in the UI (as the subcount record is required).

Fix project/survey user search (does proper backend typeahead search). 
- Previously it was hard-limited to 50 users, which meant not all users were ever shown in the frontend.

Update the clamav library we use to one that is more supported.

## Testing Notes

n/a


mauberti-bc (approved):
Looks good

PR #1284, opened 2024-05-08
BugFix: Strict Zod Validation Fixes

NickPhura:
Catch-all PR for any additional issues noticed in Dev or Test.

## Links to Jira Tickets

n/a

## Description of Changes

- Update the observation measurement types. 
  - Was causing a zod error before when importing an observation CSV that had measurement columns/data defined.

## Testing Notes

Typical QA.


mauberti-bc (approved):
Successfully imported an observation with a measurement 👍

PR #1307, opened 2024-06-18
SIMSBIOHUB-593: Add constraint for a user to have only one role per Project

mauberti-bc:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-593

## Description of Changes

- Added unique constraint on (system_user_id, project_id) in project_participation table, so project members can have only one role in a Project
- Updated the project-participants service to only update records if the user's role has changed; do nothing if the role hasn't changed (deletes and inserts are the same)
- UI styling changes to the Project participant control


NickPhura (dismissed):
✔️

NickPhura (approved):

PR #1308, opened 2024-06-20
Text input for capture and mortality locations

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Added input textfield boxes to manually enter capture and mortality latitude/longitude values
- Updated validation of coordinates
- Bug fix: remove overflow hidden from the telemetry deployments list

## Testing Notes

- There should only ever be one point on the map when adding/editing capture and mortality locations


NickPhura (approved):
🐦

PR #1314, opened 2024-06-25
BugFix: Fix migration to drop missed instance of wldtaxonomic_units_id column

NickPhura:
## Links to Jira Tickets

n/a

#### Add migration to drop some deprecated columns.
- Missed an instance of the wldtaxonomic_units_id column. The original migration deprecated old file_type values AND dropped columns in the same PR. Split those two groups of changes into two separate PRs. Will be safe to move existing migration code from the old file to the new file because they all use `IF EXISTS`.

## Testing Notes

Editing an observation record works (does not result in zod validation errors).

Editing the existing migration should be safe, as the commands moved to the new migration all include `IF EXISTS`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,6 +6,7 @@ on:
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review]
     branches-ignore:
+      - test
       - prod
 
 concurrency:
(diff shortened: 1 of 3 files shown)


mauberti-bc (dismissed):

MacDeluca (approved):
🩴

PR #1318, opened 2024-06-28
Bug fix: Allow for captures without locations on the Manage Animals page

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Fixes a bug affecting PROD where capture events without location data break the frontend
- Displays an alert banner above the animal capture map when capture information is missing
- Resets release location to equal capture location if release location exists but both locations are said to be the same


app/src/features/surveys/animals/profile/captures/components/AnimalCapturesMap.tsx
```diff
@@ -19,16 +19,25 @@ interface IAnimalCapturesMapProps {
...
+              type: 'Feature',
+              geometry: {
+                type: 'Point',
+                coordinates: [capture.capture_location.longitude, capture.capture_location.latitude]
+              },
+              properties: { captureId: capture.capture_id, date: capture.capture_date }
+            } as Feature)
+          : null // Return null instead of undefined
+    )
+    .filter((feature): feature is Feature => isDefined(feature));
```
MacDeluca: I dont think the last filter is needed, same with the ternary in the above map. Casting 'captureMapFeatures' as Feature[] is fine in this case, we know more than the typescript compiler.

Array.filter isn't smart enough to tell typescript (in our version) that the values will be defined in the following map.

MacDeluca (dismissed):
Ran locally and looks good to me!

NickPhura (dismissed):
🦎

NickPhura (approved):
🧙

PR #1320, opened 2024-07-05
Fix bug in observation inserts and updates

mauberti-bc:
## Description of Changes

Fixes a bug in the SQL when inserting and updating species observations.
- species_observation_id could exist in the object but be null, in which case the SQL was incorrect


NickPhura (approved):
🌵

PR #1323, opened 2024-07-15
BugFix: Updated API Error Handling In Response

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Small update to the API error handling. In some cases, it would return an empty object instead of any useful error information. Now correctly parses out the error name and message when returning an error response from the API.

## Testing Notes

For an error (throw new Error('blah') in the API. See that the response in the frontend network tab shows the error message.


MacDeluca (dismissed):
🚢

MacDeluca (approved):
🚢

PR #1326, opened 2024-07-18
TechDebt: Update Multer middleware and OpenAPI spec for file imports

MacDeluca:
## Description of Changes

- Extended express request object to contain files
- Modified multer middleware to re-assign the req.body.media
- Created schema for files + csv files
- Created getFileFromRequest helper function
- Updated upload/import endpoints to use getFileFromRequest
- Updated upload/import endpoints to use new spec for incoming files

## Testing Notes

- All import / upload endpoints should work as expected


api/src/app.ts
```diff
@@ -17,6 +17,16 @@ import { getLogger } from './utils/logger';
 
 const defaultLog = getLogger('app');
 
+/**
+ * Extending express request type.
+ *
+ */
+declare module 'express-serve-static-core' {
+  interface Request {
+    files: Express.Multer.File[];
```
MacDeluca: Im undecided if this should be optional or not.

The 'files' property only exists on multipart/form-data endpoints, but those endpoints now have the correct OpenAPI spec where the 'files' property will be defined.

NickPhura (dismissed):
Looks great!

🚂

NickPhura (approved):
💥

PR #1327, opened 2024-07-19
Techdebt: Extending Express Request Object + Fixing type errors

MacDeluca:
## Description of Changes

- Extended Express request type - src/types/express.d.ts
- Created getters for getting extended properties from request - src/utils/request.ts
- Updated endpoints type safety / property access - req['keycloak_token'] -> req.keycloak_token

## Testing Notes

- App should work as expected


NickPhura (approved):
Looks good.

Tried out a random assortment of requests (surveys, file upload, critter, admin), as both an admin and creator system role, and found no issues.

PR #1328, opened 2024-07-20
BugFix: Fix Migration Order of Operations

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

The audit triggers need to be created before inserting data into the lookup table (otherwise the create_user column fails the non-null constraint).

I just moved the block of audit/journal trigger sql to above the insert into method_technique sql.

These migrations ran successfully in the PR against a fresh database, which shouldn't have mattered since these tables are all brand new. But something must be different for it to fail against Dev which has existing data. Not exactly sure what the cause is at the moment.

## Testing Notes

Migrations should run successfully

I did dry-run this against Dev, and it completes successfully now. Previously was failing due to the create_user non-null constraint on the method_technique table.


MacDeluca (dismissed):
⚡

MacDeluca (approved):
LGTM

PR #1331, opened 2024-07-30
BugFix: Overwriting animal sex when editing an animal

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- When editing an existing animal, the update payload now includes the original animal sex instead of overwriting the sex to "Unknown"

## Testing Notes

- There is no UI for editing the animal sex, but this can be tested by creating an animal, editing the sex in DBeaver, then editing the animal in the UI. Confirm that the update payload includes the original sex instead of UNKNOWN.

### DIFF
--- a/app/src/features/surveys/animals/animal-form/create/CreateAnimalPage.tsx
+++ b/app/src/features/surveys/animals/animal-form/create/CreateAnimalPage.tsx
@@ -25,6 +25,7 @@ import { Link as RouterLink } from 'react-router-dom';
 export const defaultAnimalDataFormValues: ICreateEditAnimalRequest = {
   nickname: '',
   species: null,
+  sex: AnimalSex.UNKNOWN,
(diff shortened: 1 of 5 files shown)


app/src/features/surveys/animals/animal-form/create/CreateAnimalPage.tsx
```diff
@@ -25,6 +25,7 @@ import { Link as RouterLink } from 'react-router-dom';
 export const defaultAnimalDataFormValues: ICreateEditAnimalRequest = {
   nickname: '',
   species: null,
+  sex: AnimalSex.UNKNOWN,
```
NickPhura, line 28: I assume eventually we'd want to fetch the list of allowed sex values from Critterbase?

NickPhura (approved):

PR #1341, opened 2024-08-13
TechDebt: Fix Patch System User SQL

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

- Fix Patch System User SQL
- Fix `/self` endpoint
  - The endpoint was rejecting any system user that had no user_guid, because the authorization check runs before the patch sql, and only uses the user_guid. It is safe to remove the authorization scheme because the endpoint already checks for a valid systemUserId, which can only exist if the connection is opened with the token successfully.
  - Rename endpoint function to `getSelf` for better readability.
  
## Testing Notes

- When a user logs in, everything should work as expected: (the patch is called, the user context is set, and the self endpoint is called).
- If you manually set the `user_guid` column of your `system_user` record to `null`, and then log in, all of the calls should succeed, and your user_guid should be re-added to your system_user record.

### DIFF
--- a/api/src/paths/user/self.test.ts
+++ b/api/src/paths/user/self.test.ts
@@ -6,7 +6,7 @@ import * as db from '../../database/db';
 import { HTTPError } from '../../errors/http-error';
 import { UserService } from '../../services/user-service';
 import { getMockDBConnection, getRequestHandlerMocks } from '../../__mocks__/db';
-import * as self from './self';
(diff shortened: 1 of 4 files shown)


mauberti-bc (approved):
Tested locally by removing my GUID and it successfully patched my record 👍 Also added a missing key prop from an old PR

PR #1342, opened 2024-08-16
SIMSBIOHUB-600: Handle Keycloak Auto-Renew Error, Write Logs to File

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-600

#### Keycloak login infinite loop
- Add signout when keycloak silent renew fails.
- Explicitly declare the `automaticSilentRenew` flag in the authConfig. This is on by default, but declaring to make it more visible.

#### Console log size
- Add new transport to winston logger to write logs to file(s)
  - For non-production builds: Continues to write to the console (as well as the files)
- Add new env vars for managing the log settings of the file transport

- Add PVC config to api deploy yaml (for storing logs)

#### Other
- Remove deprecated comment from useAxios.ts

## Testing Notes

Tested this by logging into SIMS in 2 different tabs, and then logging out of one of the tabs.
When the other tab tries to refresh the token, it fails, because the token has already been revoked, and immediately redirects the user to the home page. Watched it happen live, and it all happens quickly and without any weird UI interactions or anything.


api/src/utils/logger.ts
```diff
@@ -43,46 +44,94 @@ import winston from 'winston';
...
  */
 export const getLogger = function (logLabel: string) {
-  return winston.loggers.get(logLabel || 'default', {
-    transports: [
+  const transports = [];
+
+  // Output logs to file
+  transports.push(
+    new DailyRotateFile({
+      dirname: process.env.LOG_FILE_DIR || 'data',
```
NickPhura:
```suggestion
      dirname: process.env.LOG_FILE_DIR || 'data/logs',
```

mauberti-bc (approved):
Tested locally and looks good 👍

PR #1343, opened 2024-08-19
Add Ecological Units to Survey Focal Species

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- When a focal species is selected, users can now select multiple ecological units for each focal species.

## Testing Notes

- Should be able to add focal species and ecological units for each focal species
- If a species does not have any ecological units, the button is disabled


NickPhura (approved):
🐅

PR #1347, opened 2024-08-23
BugFix: Update Migration To Fix Lingering Duplicate Users Issue

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Update the existing migration that was meant to fix duplicate user issues, but which left some issues unfixed.

## Testing Notes

On a new local build, it should delete all duplicate users (rather than end-date them), and should now correctly prevent duplicate user records from being created.
- Should enforce unique `user_guid` and unique (`user_Identifier` + `user_identity_source`).

## Editing Existing Migration

Normally we dont edit existing migrations, but since this is only in Test, and it only touches the system_user table, it was easier to just modify it, and then I'll manually run some queries to account for the differences in Dev and Test (since this migration wont re-run in either of those environments)


mauberti-bc (dismissed):
Tested locally, looks good

database/tsconfig.json
```diff
@@ -24,7 +24,7 @@
     "strict": true,
     "typeRoots": ["node_modules/@types"]
   },
-  "include": ["src"],
+  "include": ["src", "20240722000002_remove_duplicate_users.ts", "20240722000002_remove_duplicate_users.ts"],
```
MacDeluca: Is this intentional?

NickPhura: nope!

It likes to auto-add files when I move them around

MacDeluca (approved):
LGTM 🪿

PR #1354, opened 2024-08-28
Add Sign Field to Observation Subcounts

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Adds "observation_subcount_sign" to the observations table (the database field was added in an previous migration)
- Includes "sign" as a supported column in the observation CSV validator

## Testing Notes

- Should be able to import observations from a csv file with a "Sign" column, and add observations with a sign value manually.


NickPhura:
@mauberti-bc Minor: Is there a description for the observation sign column?
In `GridColumnDefinitions.tsx -> ObservationSubcountSignColDef`

mauberti-bc:
> @mauberti-bc Minor: Is there a description for the observation sign column? In `GridColumnDefinitions.tsx -> ObservationSubcountSignColDef`

Good catch: "The observed evidence of the species"

NickPhura:
@mauberti-bc 

Added backend validation for the sign field when importing a csv.

Added some missing backend validation for manual observation records (we weren't validating environment columns before).

I also fixed a bug that prevented quantitative environment fields from being returned in the search (when adding columns to the observation table).

I believe everything is working as expected, but if you have time to give it a quick manual test as well, that would be valuable.

mauberti-bc:
> @mauberti-bc
> 
> Added backend validation for the sign field when importing a csv.
> 
> Added some missing backend validation for manual observation records (we weren't validating environment columns before).
> 
> I also fixed a bug that prevented quantitative environment fields from being returned in the search (when adding columns to the observation table).
> 
> I believe everything is working as expected, but if you have time to give it a quick manual test as well, that would be valuable.

Awesome. I'll test it now

NickPhura (approved):
🦌

PR #1358, opened 2024-09-04
Insert new method lookup values for SPI ETL

LouisThedroux:
## Links to Jira Tickets
- N/A 

## Description of Changes
- Adds new reference data used in the SPI ETL code

## Testing Notes
- Should see all the new lookup options (excluding undetermined) when creating a technique


MacDeluca:
> Get this error when running make test 
- Run ``` npm i``` in the API directory. Your local has outdated dependencies from recent PR's which were merged into dev. If you install the newly added ones this should go away.

NickPhura:
@LouisThedroux  Re error: Its a new package that was added just recently, so you probably just need to run `make install` (which just runs `npm install` in each sub-project)

NickPhura (approved):
LGTM

Tested the migration against existing techniques/methods associated to observations. Everything seems to work great.

PR #1363, opened 2024-09-11
BugFix: Migration to add NOT NULL constraint to all uuid columns

MacDeluca:
## Description of Changes

- Created migration to add NOT NULL constraint to all uuid columns
- Generates a uuid if cell currently holds a NULL value

## Testing Notes
- Run this command to check for existing tables with 'uuid' column.
- All uuid columns should now be NOT NULL
```sql
SELECT * FROM information_schema.COLUMNS WHERE column_name = 'uuid';
```


mauberti-bc (approved):
Ran the migration locally and it worked 👍

NickPhura (approved):
👍

PR #1364, opened 2024-09-11
Spi migrate users ups into test spi

AMEIJER1:
## Links to Jira Tickets

## Description of Changes

## Testing Notes


spi/src/index.ts
```diff
@@ -49,6 +49,9 @@ async function main() {
     // STEP 6. Transforms SPI Design Components into SIMS Sampling Sites
     await transformSampleSites(connection);
```
mauberti-bc, line 59: It looks like the same function is being called twice

spi/src/transformations/project.ts
```diff
@@ -29,24 +29,65 @@ export const transformProjects = async (connection: IDBConnection): Promise<void
...
+
+        CASE 
+        WHEN p.coordinator IS NOT NULL AND w_mapping.full_name LIKE '%' || TRIM(BOTH ', ' FROM p.coordinator) || '%' AND spp.email_address LIKE '%@gov.bc.ca%' THEN 
+            (SELECT project_role_id FROM biohub.project_role WHERE name = 'Coordinator')
+        WHEN NOT EXISTS (
+            SELECT 1 
+            FROM biohub.project_participation 
+            WHERE project_id = w_mapping.project_id 
+              AND project_role_id = (SELECT project_role_id FROM biohub.project_role WHERE name = 'Coordinator')
+        ) THEN (SELECT project_role_id FROM biohub.project_role WHERE name = 'Coordinator'), (SELECT system_user_id FROM biohub.system_user WHERE user_identifier = 'spi')
```
mauberti-bc: Is there a syntax error here where the SPI user is being selected?

spi/src/transformations/project.ts
```diff
@@ -29,24 +29,65 @@ export const transformProjects = async (connection: IDBConnection): Promise<void
...
+         FROM public.migrate_spi_user_deduplication 
+         WHERE w_mapping.person_id = ANY (spi_person_ids)
+        ), 
+        (SELECT system_user_id 
+         FROM biohub.system_user 
+         WHERE user_identifier = 'spi')
+        ) AS system_user_id,
+
+        CASE 
+        WHEN p.coordinator IS NOT NULL AND w_mapping.full_name LIKE '%' || TRIM(BOTH ', ' FROM p.coordinator) || '%' AND spp.email_address LIKE '%@gov.bc.ca%' THEN
```
mauberti-bc: The case when can be simplified

```
CASE
  WHEN TRIM(p.coordinator) = w_mapping.full_name AND spp.email_address LIKE '%@gov.bc.ca%' THEN 
    (SELECT project_role_id FROM biohub.project_role WHERE name = 'Coordinator')
  WHEN spp.email_address LIKE '%@gov.bc.ca%' THEN 
    (SELECT project_role_id FROM biohub.project_role WHERE name = 'Collaborator')
  ELSE 
    (SELECT project_role_id FROM biohub.project_role WHERE name = 'Observer')
END
```

spi/src/transformations/study-area.ts
```diff
@@ -0,0 +1,58 @@
...
+    ),
+    buffered_geographies AS (
+        SELECT 
+            sp.survey_id,
+            sp.study_area_name,
+            sp.study_area_description,
+            sp.convex_hull,
+            sp.geo,
+            sp.so_count,
+            (0.2 * sqrt(ST_Area(sp.convex_hull)) + 0.05 * pow((ST_Area(sp.convex_hull) / sp.so_count), 0.75)) AS buffer_value
```
mauberti-bc: We should add a comment in the code indicating that these constants came from the Python script for inferring study areas based on observations.

spi/src/transformations/user.ts
```diff
@@ -29,39 +29,54 @@ export const transformUsers = async (connection: IDBConnection): Promise<void> =
...
+        'Migrated from SPI as user ' || md.id,
+        COALESCE(spp.email_address, 'default') AS email
     FROM 
-      migrate_spi_user_deduplication;
+        migrate_spi_user_deduplication md
+    LEFT JOIN 
+        public.spi_secure_persons spp ON spp.first_name = md.given_name AND spp.last_name = md.family_name
+    LEFT JOIN 
+        w_existing_users eu ON eu.given_name = md.given_name AND eu.family_name = md.family_name
+    WHERE eu.system_user_id IS NULL;
```
mauberti-bc, line 64: eu.system_user_id will never be null because it is the primary key

AMEIJER1: We are left-joining eu onto deduplication table (which is the from table in the insert statement) -- in this case, everything from deduplication is joined to only matching records in eu, wouldnt therefore any deduplication records without a match in eu have a null eu.system_user_id as they haven't been "inserted" yet?

spi/src/index.ts
AMEIJER1: fixed! and added other functions from the remaining completed transformations in

spi/src/transformations/project.ts
AMEIJER1: simplified in code! Would this still ensure that when no coordinator is set, that the spi user is set as coordinator?

spi/src/transformations/study-area.ts
AMEIJER1: comment inserted!

PR #1365, opened 2024-09-11
TechDebt: Centralize Virus Scan

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Move the virus scan step into the app.ts file, so that it doesn't need to be included in every endpoint that handles file uploads.

## Testing Notes

Uploading/Downloading files works as before.

Can use this string (paste into a random txt file) to trigger our APIs file upload virus scans: https://en.wikipedia.org/wiki/EICAR_test_file

I've tested a few places we upload files, and they work as expected. Every file upload goes through the same field and should be handled by the same multer middleware anyways, so if one works they should all work.

### DIFF
--- a/api/src/app.ts
+++ b/api/src/app.ts
@@ -104,7 +104,12 @@ const openAPIFramework = initialize({
           }
         });
 
-        await Promise.all(virusScanPromises);
+        try {
+          await Promise.all(virusScanPromises);
+        } catch (error) {
+          // If a virus is detected, return error and do not continue
+          return next(error);
+        }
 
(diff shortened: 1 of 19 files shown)


NickPhura:
@mauberti-bc I know we talked about patching the error handling as part of this PR, but on second thought there is no real overlap, so in the interest of small PRs we may as well merge this when its ready. I can just open a separate PR for the frontend changes.

mauberti-bc (approved):
👍

MacDeluca (approved):
🫎

PR #1371, opened 2024-09-18
BugFix: Request Measurements as API User for the Standards Page

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Request measurements from Critterbase as the API user, instead of a keycloak user (which does not exist because the endpoint does not have authentication)

## Testing Notes

- Requires the keycloak_uuid column in Critterbase to be changed to varchar instead of nbchar(32), to prevent the SIMS API user's uuid ("biohub_api") from being padded to 32 characters in Critterbase.


NickPhura (approved):
🦁

PR #1372, opened 2024-09-18
BugFix: Show Qualitative Options for Environments on Standards Page

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- The cards did not display the options for qualitative environmental measurements, they only showed the measurement name and description
- Adds units to quantitative species measurements and environments

## Testing Notes
- The species measurements search is currently broken due to a Critterbase bug related to the keycloak_uuid field
- Standards page should display options for all qualitative environmental variables, method attributes and species attributes.


NickPhura (approved):
🧍‍♂️

PR #1377, opened 2024-09-26
Add Comments to Observation Subcount

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Adds an action column to the observations table for adding a comment to a subcount
- Allows a comment to be added to a subcount in a .csv import
- 

## Testing Notes

- Should be able to add and edit a comment to an observation through the UI
- Should be able to import a .csv of observations with a "comment" column


MacDeluca (approved):
🦠 
Made a small change for the ObservationSubcounCommentDialog to prevent having to use the `useEffect`.
But otherwise looks good to me! Was able to add and edit multiple comments and import with the CSV.

PR #1380, opened 2024-09-27
SIMSBIOHUB-449: Upgrade Postgres Version

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-449

## Description of Changes

Upgrade local and remote postgres versions from 12.5 to 14.2.

## Testing Notes

Database builds/deploys as usual.

Existing data produced under version 12.x gets picked up by the new version correctly. Worst case scenario we need to manually migrate the data, but ideally it just works without issues.


NickPhura:
Will re-open in order to test version migration from 12.5 to 15

PR #1387, opened 2024-10-02
BugFix: Remove Duplicate Environment Columns in Observations Table

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-624

## Description of Changes

- Changed SELECT to SELECT DISTINCT when retrieving quantitative environment definitions
- Fixes a bug where if a quantitative environment value is given for 2 or more records while editing, that environment variable's column gets duplicated in the observations table

## Testing Notes

- When editing 2 rows and adding a quantitative environment value to both (eg. setting air temperature for both records), the modified measurement's column should not be duplicated


NickPhura (approved):
Tested locally, looks good, nice catch!

Tried some of the other SQL in that repository as well, but they all worked as expected.

PR #1389, opened 2024-10-03
SIMSBIOHUB-619: Import Observations With Sampling Information

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-619

## Description of Changes

- Allow observations to be imported from a .csv with sampling site, sampling method, and sampling period
- Sampling information is optional
- Replaces the observation latitude/longitude NOT NULL constraints with a check for either latitude/longitude or survey_sample_period_id being NOT NULL
- Replaces the observation date/time NOT NULL constraints with a check for either date or survey_sample_period_id being NOT NULL
- If sign is not specified, sign defaults to "direct sighting"

## Testing Notes

- Sampling information is optional in the csv
- If a site but no method is specified, and only one method exists for that site, SIMS will use the only method
- If a method but no period is specified, and only one period exists for that method, SIMS will use the only period
- If a method but no period is specified and multiple periods exists, SIMS will try to find the period based on the observation date/time. 
- There is no validation that the observation date/time lies within the period's start and end

NOTE: Dates are assumed to be in UTC format when read in with xlsx.read(), causing observation dates to be incorrectly offset by 1 day earlier. This will be fixed in a separate PR.


api/src/openapi/schemas/observation.ts
```diff
@@ -60,19 +60,23 @@ export const observervationsWithSubcountDataSchema: OpenAPIV3.SchemaObject = {
             nullable: true
           },
           latitude: {
-            type: 'number'
+            type: 'number',
+            nullable: true
```
MacDeluca: Do we want min / max values on these?

api/src/services/observation-service.ts
```diff
@@ -556,7 +560,7 @@ export class ObservationService extends DBService {
 
     const measurementsToValidate: IMeasurementDataToValidate[] = worksheetRowObjects.flatMap((row) => {
       return measurementColumnNames.map((columnName) => ({
-        tsn: String(row['ITIS_TSN'] ?? row['TSN'] ?? row['TAXON'] ?? row['SPECIES']),
+        tsn: String(getColumnCellValue(row, 'ITIS_TSN').cell),
```
MacDeluca, line 565: String(undefined) will lead to unexpected results -> 'undefined'. Are we sure this value is defined at this point?

api/src/services/observation-service.ts
```diff
@@ -1019,4 +1194,30 @@ export class ObservationService extends DBService {
...
+    const value = cellValue?.toLowerCase(); // Normalize the cell value
+
+    // No value exists and no default value, so must be null
+    if (!value && !defaultValue) {
+      return null;
+    }
+
+    // The cell has no value so return the default
+    if (!value && defaultValue) {
+      return codeOptions.find((option) => option.name.toLowerCase() === defaultValue.toLowerCase())?.id || null;
```
MacDeluca: suggestion: We could pass the defaultCodeId as a param to prevent having to loop through all the codes for every code cell.
ie: Loop through once externally vs every time internally

api/src/services/observation-service.ts
```diff
@@ -1019,4 +1194,30 @@ export class ObservationService extends DBService {
...
+  /**
+   * Gets the code id value with a matching name from a set of options. If the function returns null, the
+   * request should probably throw an error.
+   *
+   * @param cellValue The name of a code to find the id for
+   * @param codeOptions The reference codes from which to find the id
+   * @param defaultValue A default value for when cellValue is null
+   * @returns
+   */
+  _getCodeIdFromCellValue(cellValue: string | null, codeOptions: ICode[], defaultValue?: string | null): number | null {
```
MacDeluca: nit: We can pass a map (value, id) to this function to prevent having to loop through each item in the codeOptions.

new Map<value, id>()

api/src/services/observation-service.ts
```diff
@@ -799,6 +835,145 @@ export class ObservationService extends DBService {
...
+   * @return { { sampleSiteId: number, sampleMethodId: number, samplePeriodId: number } | null } The sampling data with IDs, or null if no valid data is found.
+   */
+  /**
+   * Extracts sampling data from the worksheet row object and maps site names, method techniques, and periods
+   * to their respective IDs using the provided samplingLocations.
+   *
+   * @param {Record<string, any>} row - The current row of the worksheet being processed.
+   * @param {SampleLocationRecord[]} samplingLocations - The available sampling locations for the survey, used for mapping names to IDs.
+   * @return { { sampleSiteId: number, sampleMethodId: number, samplePeriodId: number } | null } The sampling data with IDs, or null if no valid data is found.
+   */
```
NickPhura: Two jsdoc comments

api/src/services/observation-service.ts
```diff
@@ -615,13 +626,11 @@ export class ObservationService extends DBService {
...
-      // as it is the only code set currently being used in the observation CSVs, and is required. This logic will need
-      // to be updated to be more generic if other code sets are used in the future, or if they can be nullable.
-      const observationSubcountSignId = codeTypeDefinitions.OBSERVATION_SUBCOUNT_SIGN.find(
-        (option) =>
-          option.name.toLowerCase() === getColumnCellValue(row, 'OBSERVATION_SUBCOUNT_SIGN')?.cell?.toLowerCase()
-      )?.id;
+      const observationSubcountSignId = this._getCodeIdFromCellValue(
+        getColumnCellValue(row, 'OBSERVATION_SUBCOUNT_SIGN').cell,
+        codeTypeDefinitions.OBSERVATION_SUBCOUNT_SIGN,
+        'direct sighting'
```
NickPhura: just checking: this is meant to be hard coded?  If it is, I would move this to a const at the top of the file. Best practice anytime you have seemingly random numbers or strings in the middle of some code.

api/src/services/observation-service.ts
mauberti-bc: Yes, if ```getColumnCellValue(row, 'ITIS_TSN').cell``` returns undefined, the ```validateCsvFile``` function on line 515 would throw an error, prior to reaching the above code.

api/src/openapi/schemas/observation.ts
mauberti-bc: Fixed

api/src/services/observation-service.ts
mauberti-bc: Fixed

NickPhura (dismissed):
🐧

NickPhura (approved):

PR #1393, opened 2024-10-10
SIMSBIOHUB-629: Paginate Sampling Sites & Separate Spatial Data in API Responses

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-629

## Description of Changes

- Adds a ```/sample-site/spatial``` endpoint for retrieving the geojson data for all sites in the survey
- Removes geojson property from the response when requesting all sampling sites in the survey, replacing it with a "geometry_type" string
- Paginates sampling sites on the frontend (backend pagination was already in place)
- For the sampling site accordions on the observations page, does not load accordion content until expanded
- When retrieving observations, sampling sites related to the requested observations are included in the supplementaryData object
- Removes sample site data loader from the survey context

## Testing Notes

- Should not be any double requests for sampling site data
- Since sampling periods depend on the same request as sampling methods, we still need to solve sample period pagination (records shown in the sample period table are limited to those retrieved in the latest paginated sample method request).


NickPhura:
Just going to change this to a draft temporarily as we are hitting the limit on our openshift resources.

NickPhura:
@mauberti-bc 
Okay, other than any minor CSS tweaks, etc, you might have, it should be good to go.

I'll do the other TODOs in a separate PR:
- Add pagination to the period table on the survey page.
- See if I can make the dropdown modal wider (somehow its working for the taxon/site control, but not the technique/period, even though they use the same autocomplete under the hood. But havent poked at it too hard yet)

api/src/openapi/schemas/observation.ts
```diff
@@ -412,7 +414,111 @@ export const observervationsWithSubcountDataSchema: OpenAPIV3.SchemaObject = {
               }
             }
           }
-        }
+        },
+        sample_sites: SampleLocationSchema
+        // sample_sites: {
```
MacDeluca: 👁️

NickPhura: Removed

MacDeluca (approved):
Poked around the UI and it all seems happy
🦖

PR #1396, opened 2024-10-16
Update Survey Job Lookup Options for Survey Participants

AMEIJER1:
## Links to Jira Tickets

- NA

## Description of Changes

- Adding Crew lead and Crew member as survey job options
- Updating any survey participation records selecting the biologist role to select the crew member role 
- Removing the biologist from the survey job options list 

## Testing Notes

- Make test succeeded


database/src/migrations/20241012021000_survey_jobs.ts
```diff
@@ -0,0 +1,50 @@
...
+ *  add crew member and crew lead
+ *
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
+    
+    SET SEARCH_PATH=biohub
```
mauberti-bc: Overall looks good, but there's a missing semiclon after biohub

```SET SEARCH_PATH=biohub;```

Comment (AMEIJER1) [reply database/src/migrations/20241012021000_survey_jobs.ts]: oops! thanks!

File: database/src/migrations/20241012021000_survey_jobs.ts [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+        INSERT INTO biohub.survey_job (name, record_effective_date, description)
+        VALUES
+          ('Crew lead', NOW(), 'A participant of a survey in a crew lead role.'),
+          ('Crew member', NOW(), 'A participant of a survey in a crew member role.');
+
+    ----------------------------------------------------------------------------------------
+    -- Reassigning sims biologist roles to crew member
+    ----------------------------------------------------------------------------------------
+      
+        UPDATE biohub.survey_participation sp
```
mauberti-bc: The migration is failing here because you can't give your table an alias in this part of the update. You can drop the table references, and you since every table touched is in the biohub schema that you set for the search path, you can drop the schema reference:

```UPDATE survey_participation
        SET survey_job_id = (
            SELECT survey_job_id 
            FROM survey_job
            WHERE name = 'Crew member'
        )
        WHERE survey_job_id = (
            SELECT survey_job_id 
            FROM survey_job
            WHERE name = 'Biologist'
        );
```

mauberti-bc:
In the seed (mock) data that gets inserted after the migrations, the SQL for inserting survey participant data was grabbing the survey_job_id = 1 (line 168 in 03_basic_project_survey_setup.ts). However, since we're hard deleting the Biologist option, which was survey_job_id = 1, the seed is trying to insert a survey participant with a null job. You'll have to update the seed file to successfully find a survey_job_id.

Here's one option:

```
const insertSurveyParticipationData = (surveyId: number) => `
  INSERT into survey_participation
    ( survey_id, system_user_id, survey_job_id )
  VALUES
    (
      ${surveyId},
      (
        SELECT COALESCE((
          SELECT
            system_user_id
          FROM
            system_user su
          WHERE
            su.user_identifier = '${PROJECT_SEEDER_USER_IDENTIFIER}'
        ), 1)
      ),
      (SELECT survey_job_id FROM survey_job LIMIT 1)
    )
  ;
`;
```

That will grab whatever the first record in survey_job is, based on the sorting order, for all survey participants. To pick a potentially different job for each participant, you could randomize the sort:
```SELECT survey_job_id FROM survey_job ORDER BY random() LIMIT 1```

File: compose.yml [hunk 3]
```diff
@@ -7,7 +7,7 @@ services:
       context: ./database/.docker/db
       dockerfile: Dockerfile
       args:
-        - POSTGRES_VERSION=${POSTGRES_VERSION}
+        - POSTGRES_VERSION=14.2
```
mauberti-bc: This should stay ${POSTGRES_VERSION}, but since it's reading from your .env file, you'll need to change the POSTGRES_VERSION value from 12.5 to 14.2 in your .env.

NickPhura (approved):
👍

PR #1401, opened 2024-10-18
SIMSBIOHUB-620: Manage System Alerts

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-620

## Description of Changes

- Adds the ability for system admins to create Alerts that all users see
- Displays alerts on relevant pages, with each page requesting only relevant alerts (eg. alerts of type "Survey" for Survey page)
- Repurposes Manage Users page to generic Admin page with controls for managing users and alerts
- Adds system alerts to the database seeds

## Testing Notes

- Should be able to create, edit and delete alerts
- Alerts should only display on the appropriate pages (eg. General on the Survey page)


NickPhura:
Converting to draft temporarily to save Openshift resources.

database/src/migrations/20240927000000_alert_table.ts
```diff
@@ -0,0 +1,127 @@
...
+    -- Create new table
+    ----------------------------------------------------------------------------------------
+
+    CREATE TABLE alert (
+      alert_id           integer            GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+      alert_type_id      integer            NOT NULL,
+      name               varchar(100)       NOT NULL,
+      message            varchar(250)       NOT NULL,
+      data               json,
+      severity           varchar(100)       NOT NULL,
```
NickPhura: nit: missing comment for this column, below.

NickPhura: Also, does this need to reference a lookup table of severity codes?

app/src/hooks/api/useAlertApi.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Returns a set of supported api methods for managing alerts
+ *
+ * @param {AxiosInstance} axios
+ * @return {*} object whose properties are supported api methods.
+ */
+export const useAlertApi = (axios: AxiosInstance) => {
+  /**
+   * Get project alert details based on its ID for viewing purposes.
```
NickPhura: nit: Doc seems to be out of date.

NickPhura: All of the doc strings seem slightly off. They all mention 'project' even though the alerts arent project specific.

app/src/features/alert/banner/SystemAlertBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+import Collapse from '@mui/material/Collapse';
+import grey from '@mui/material/colors/grey';
+import AlertBar from 'components/alert/AlertBar';
+import dayjs from 'dayjs';
+import { useBiohubApi } from 'hooks/useBioHubApi';
+import useDataLoader from 'hooks/useDataLoader';
+import { useEffect, useState } from 'react';
+import { TransitionGroup } from 'react-transition-group';
+
+export enum SystemAlertBannerEnum {
```
NickPhura: I would maybe move this into `useAlertApi.interface.ts`.

The filter param `IAlertFilterParams.types` is just defined as a `string[]`, but it looks like its value should be an array of this type.

app/src/features/admin/users/access-requests/list/pending/AccessRequestPendingList.tsx
```diff
@@ -33,7 +32,10 @@ interface IAccessRequestPendingListProps {
  * @returns
  */
 const AccessRequestPendingList = (props: IAccessRequestPendingListProps) => {
-  const { accessRequests, codes, refresh } = props;
+  const { accessRequests, refresh } = props;
+
+  const codesContext = useCodesContext();
```
NickPhura, line 37: Will want to add a useEffect to ensure codes are loaded:

```typescript
  useEffect(() => {
    codesContext.codesDataLoader.load();
  }, [codesContext.codesDataLoader]);
```

api/src/openapi/schemas/alert.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+};
+
+/**
+ * Schema for system alerts
+ */
+export const systemAlertSchema: OpenAPIV3.SchemaObject = {
+  ...baseSystemAlertSchema,
+  required: ['alert_id', 'name', 'message', 'data', 'alert_type_id', 'record_end_date', 'severity'],
```
NickPhura, line 46: Add the 'additionalProperties: false' line

api/src/openapi/schemas/alert.ts
```diff
@@ -0,0 +1,62 @@
...
+      description: 'Primary key of the alert'
+    }
+  }
+};
+
+/**
+ * Schema for creating system alerts
+ */
+export const systemAlertCreateSchema: OpenAPIV3.SchemaObject = {
+  ...baseSystemAlertSchema,
```
NickPhura, line 80: Add the 'additionalProperties: false' line

api/src/openapi/schemas/alert.ts
NickPhura: Could maybe just add it to the base schemas as well

api/src/paths/alert/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  parameters: [
+    {
+      in: 'query',
+      name: 'types',
+      required: false,
+      schema: {
+        type: 'array',
+        items: {
+          type: 'string',
+          nullable: true
```
NickPhura: You probably dont want this to be nullable. Since this is inside the `items` section, this means that the array is allowed to have null values.
```
types=['type1', null, 'type2']
```
I would just remove the nullable entirely, and leave the query param as not required. That way the caller can omit `types`, or else include it with non-null values.

Same with the query params below.

api/src/paths/alert/index.ts
```diff
@@ -0,0 +1,244 @@
...
+    types: req.query.types ?? []
+  };
+}
+
+export const POST: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          discriminator: 'SystemUser',
```
NickPhura: This should be `'SystemRole'`

There are a few other endpoints I notice with the same issue.

Searching `discriminator: 'SystemUser',` seems to find the other 3 I can see.

api/src/paths/alert/index.ts
```diff
@@ -0,0 +1,244 @@
...
+          schema: {
+            type: 'object',
+            additionalProperties: false,
+            required: ['alert_id'],
+            properties: {
+              alert_id: {
+                type: 'number'
+              }
+            }
+          }
```
NickPhura, line 191: (Just a general comment, no change needed) 

We have a mixture of patterns for responses in SIMS atm.  Returning the newly created record or the primary ID is fine, but its also fine to just return 200 and nothing else:

```
responses: {
    200: {
      description: 'Create alert OK'
    },
    ....
}

File: api/src/paths/alert/{alertId}/index.test.ts [hunk 10]
```diff
@@ -0,0 +1,243 @@
...
+        agency: null
+      };
+
+      const requestHandler = getAlertById();
+
+      await requestHandler(mockReq, mockRes, mockNext);
+
+      expect(mockRes.jsonValue).to.eql(mockAlert);
+      expect(mockDBConnection.open).to.have.been.calledOnce;
+      expect(mockDBConnection.commit).to.have.been.calledOnce;
```
NickPhura, line 61: Nice 👍

NickPhura: Could add a call to check `release` was called too.

(Lots of our old tests don't have these checks, but will help moving forward if we add them)

api/src/paths/alert/{alertId}/index.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+  };
+}
+
+export const PUT: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          discriminator: 'SystemUser',
```
NickPhura: Should be `SystemRole`

database/src/seeds/03_basic_project_survey_setup.ts
```diff
@@ -770,3 +775,31 @@ const insertAccessRequest = () => `
...
+      message,
+      data,
+      severity,
+      record_end_date,
+      create_user,
+      update_user
+    )
+  VALUES (
+    (SELECT alert_type_id FROM alert_type ORDER BY random() LIMIT 1),
+    $$${faker.lorem.words(3)}$$,
```
NickPhura, line 797: I think none of these will show in the UI since the names wont match any of the expected ones ('Projects', etc).

Could define an array of supported names, and then pick a random one to use:

```
const SystemAlertNames = [
  'Summary',
  'Manage Telemetry',
  'Manage Observations',
  'Manage Animals',
  'Manage Sampling',
  'Projects',
  'Surveys',
  'Standards',
  'Administrator',
  'Funding'
];

const insertSystemAlert = () => `
  ...
  $$${SystemAlertNames[Math.floor(Math.random() * SystemAlertNames.length)]}$$,
  ...
`;
```

api/src/paths/alert/{alertId}/index.ts
```diff
@@ -0,0 +1,310 @@
...
+import { getLogger } from '../../../utils/logger';
+
+const defaultLog = getLogger('paths/alert/{alertId}/index');
+
+export const GET: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          discriminator: 'SystemUser'
```
NickPhura: Just checking: 

Do we expect a random user to need to fetch an individual alert? Or should this be restricted to admins?

The only consideration is that the current repo function this calls doesn't block expired alerts. Which is maybe fine? But maybe not something we want random users to be able to fetch?

If Alerts never contain sensitive information, then no issue of course.

api/src/paths/alert/{alertId}/index.ts
```diff
@@ -0,0 +1,310 @@
...
+  description: 'Update an alert by its id.',
+  tags: ['alerts'],
+  security: [
+    {
+      Bearer: []
+    }
+  ],
+  parameters: [
+    {
+      in: 'query',
```
NickPhura: `in: 'path'`

and should be `required: true`

Also, should be an `'integer'`

api/src/paths/alert/{alertId}/index.ts
```diff
@@ -0,0 +1,310 @@
...
+      name: 'alertId',
+      schema: {
+        type: 'string'
+      }
+    }
+  ],
+  requestBody: {
+    description: 'Alert post request object.',
```
NickPhura: nit: `put`

api/src/paths/alert/{alertId}/index.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+  };
+}
+
+export const DELETE: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          discriminator: 'SystemUser',
```
NickPhura: `'SystemRole'`

api/src/paths/alert/{alertId}/index.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    try {
+      await connection.open();
+
+      const alertId = Number(req.params.alertId);
+      const alert = req.body;
+
+      const alertService = new AlertService(connection);
+
+      const id = await alertService.updateAlert({ ...alert, alert_id: alertId });
```
NickPhura, line 202: This doesn't really matter: but since the request body already contains the alert_id, you could just use that one, rather than parsing it from the path. Though both are technically fine.

Getting the id from a specific part of the request only matters when we deal with security. But since this is an admin-only endpoint, we can trust the data we receive.

api/src/repositories/alert-repository.ts
```diff
@@ -0,0 +1,228 @@
...
+        alert_id = ${alert.alert_id}
+      RETURNING alert_id
+      ;
+      `;
+
+    const response = await this.connection.sql(sqlStatement);
+
+    const result = response?.rows?.[0] ?? null;
+
+    if (!result) {
```
NickPhura: Its slightly better to check `response.rowCount`, which will always be the number of rows processed. Why that matters is because if we make an update query, and forget a where clause, it might be that it updates more than 1 record.

And so we really want to check if `response.rowCount === 1`. Because less than 1 or more than 1 is equally bad (in this case).

NickPhura: You'll see some checks like that in SIMS, and the message will be:
`''response.rowCount was !== 1, expected rowCount === 1''`

api/src/services/alert-service.ts
```diff
@@ -0,0 +1,81 @@
...
+   * @param {number} alertId
+   * @return {*}  Promise<IAlert>
+   * @memberof AlertService
+   */
+  async getAlertById(alertId: number): Promise<IAlert> {
+    return this.alertRepository.getAlertById(alertId);
+  }
+
+  /**
+   * Create and associate alert for survey.
```
NickPhura: This mentions 'survey' but I dont think its survey specific, same with the below functions.

api/src/services/alert-service.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Deactive an alert (soft delete by entering a record_end_date)
+   *
+   * @param {number} alertId
+   * @param {number} recordEndDate
+   * @return {*}  Promise<number>
+   * @memberof AlertService
+   */
+  async deactivateAlert(alertId: number, recordEndDate: string): Promise<number> {
```
NickPhura: This is never called (outside of tests).
Same with the matching repo function.

NickPhura:
@mauberti-bc Really slick in general, nice work! I like the way the alerts are displayed in the admin page and form (with the preview).

2 things I noticed when adding/editing/deleting alerts in the UI:
- The alert name has a max length of 50 in the UI, but max length of 100 in the database. Not a major issue, but maybe they were both meant to be 50 or 100?
- When I edit an alert, if I clear the expiration date, I can't save (I get an 'Invalid date' error).

Regarding CSS:
Also, was dabbling with some CSS tweaks, if you're interested. Just on my local.

app/src/features/admin/users/access-requests/list/pending/AccessRequestPendingList.tsx
NickPhura: (TBH, I've been wondering about just putting the useEffect in the `useCodesContext` hook, that way no one ever has to think about it. But, something for another PR.)

database/src/seeds/03_basic_project_survey_setup.ts
NickPhura: Disregard, misread the schema!

database/src/migrations/20240927000000_alert_table.ts
mauberti-bc: The severity values are just MUI severity codes. I switched to an enum
```
CREATE TYPE alert_severity AS ENUM ('info', 'warning', 'error', 'success');
```

app/src/hooks/api/useAlertApi.ts
mauberti-bc: Fixed

app/src/features/alert/banner/SystemAlertBanner.tsx
mauberti-bc: Thanks, fixed.

app/src/features/admin/users/access-requests/list/pending/AccessRequestPendingList.tsx
mauberti-bc: Fixed

api/src/paths/alert/index.ts
mauberti-bc: Fixed

api/src/paths/alert/{alertId}/index.ts
mauberti-bc: Not right now - I changed all the alert/{alertId} endpoints to be data/system admins only.

api/src/paths/alert/{alertId}/index.ts
mauberti-bc: Good point. I figured that the path's alertId was more reliable, in the (unlikely) scenario where the path's alertId differs from the request body's alertId.

api/src/repositories/alert-repository.ts
mauberti-bc: Thanks, that makes sense

api/src/services/alert-service.ts
mauberti-bc: Removed - they were from an earlier iteration

mauberti-bc:
> @mauberti-bc Really slick in general, nice work! I like the way the alerts are displayed in the admin page and form (with the preview).
> 
> 2 things I noticed when adding/editing/deleting alerts in the UI:
> 
>     * The alert name has a max length of 50 in the UI, but max length of 100 in the database. Not a major issue, but maybe they were both meant to be 50 or 100?
> 
>     * When I edit an alert, if I clear the expiration date, I can't save (I get an 'Invalid date' error).
> 
> 
> Regarding CSS: Also, was dabbling with some CSS tweaks, if you're interested. Just on my local.

Fixed the bug preventing end dates from being removed by updating the DateField component to correctly catch invalid dates. Also changed the max length of alert name to 50 in the database.

NickPhura (approved):
⚠️ 👍

aubertinyoung:
> ⚠️ 👍

Is that an alert banner or a warning 🥴

PR #1404, opened 2024-10-22
SIMSBIOHUB-627: Device Endpoints

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-627

## Description of Changes

Add new device endpoints.

## Testing Notes

I added some postman endpoints, so I've verified that the endpoints work as intended (happy path at least).


api/src/paths/project/{projectId}/survey/{surveyId}/devices/delete.ts
```diff
@@ -0,0 +1,136 @@
...
+      in: 'path',
+      name: 'surveyId',
+      schema: {
+        type: 'integer',
+        minimum: 1
+      },
+      required: true
+    }
+  ],
+  requestBody: {
```
MacDeluca, line 58: suggestion: We can include `required: true` to prevent empty request bodies.

api/src/services/telemetry-services/telemetry-device-service.ts
```diff
@@ -59,6 +72,61 @@ export class TelemetryDeviceService extends DBService {
...
+
+  /**
+   * Get the total count of all devices for a survey.
+   *
+   * @param {number} surveyId
+   * @return {*}  {Promise<number>}
+   * @memberof TelemetryDeviceService
+   */
+  async getDevicesCount(surveyId: number): Promise<number> {
+    return this.telemetryDeviceRepository.getDevicesCount(surveyId);
```
MacDeluca, line 115: question: Couldn't we just use the length of `this.telemetryDeviceRepository.getDevicesForSurvey()`?

NickPhura: I preemptively included pagination support for the other call, so its length might not always be the total count.

NickPhura: tbh, all of our "get all X by survey id" should really include pagination support. I think most do now? Though I don't think I did that for the get all deployments endpoint, so I might add that (in another branch)

api/src/paths/project/{projectId}/survey/{surveyId}/devices/delete.ts
NickPhura: We definitely should, good catch. I hadn't even considered that the whole body could be optional/marked required. Its probably missing in a ton of endpoints.

NickPhura: I might open a separate PR for that then

NickPhura: Added a JIRA ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-630

MacDeluca: Yea, I just caught it yesterday when I was testing my endpoint. Without it, it allows you to get pretty far into the endpoint before it throws.

MacDeluca (approved):
🩴

PR #1411, opened 2024-10-24
SIMSBIOHUB-627: Deprecate BCTW Device Service

MacDeluca:
## Description of Changes

- Deprecates BCTWDeviceService and related endpoints
- Adds new section to CodesService to return telemetryDeviceMakes

## Testing Notes

- App should work as normal / compile
- paths/codes should return all codes + new telemtry_device_makes


NickPhura (approved):
👍

PR #1414, opened 2024-10-24
Add Duration to Sampling Period UI Table

LouisThedroux:
## Links to Jira Tickets

NA

## Description of Changes

A new column, duration, has been added to the sampling period table.

The duration is calculated by determining the difference between the start_date/start_time and the end_date/end_time fields.


package-lock.json
MacDeluca, line 1: This file can be dropped. It probably appeared from running `npm i` in the root directory instead of one of the sub directories ie: api / app / database

app/src/features/surveys/sampling-information/periods/table/SamplingPeriodTable.tsx
```diff
@@ -82,6 +111,20 @@ export const SamplingPeriodTable = (props: ISamplingPeriodTableProps) => {
       field: 'end_time',
       headerName: 'End time',
       flex: 1
+    },
+    {
+      field: 'duration',
+      headerName: 'Duration',
+      flex: 1,
+      renderCell: (params) => {
+        const duration = calculateDuration(
```
mauberti-bc: Nice! This is just preference but I think you could avoid the separate function and just put the function's code in the renderCell function. We could also render the text in a more intuitive way using the humanize function

```
renderCell: (params) => {
        const startDateTime = params.row.start_time
          ? dayjs(`${params.row.start_date} ${params.row.start_time}`)
          : dayjs(params.row.start_date);
        const endDateTime = params.row.end_time
          ? dayjs(`${params.row.end_date} ${params.row.end_time}`)
          : dayjs(params.row.end_date);

        // Calculate the difference in milliseconds
        const diff = endDateTime.diff(startDateTime);

        return dayjs.duration(diff, 'millisecond').humanize();
      }
}
```

Humanize requires the relativeTIme plugin, too: 
```
import duration from 'dayjs/plugin/duration';
import relativeTime from 'dayjs/plugin/relativeTime';

dayjs.extend(duration);
dayjs.extend(relativeTime);
```

mauberti-bc: Ah, humanize actually displays "a few seconds" for > 30 seconds and "a few minutes" for < 3 minutes, which is confusing. I pushed up something more precise.

app/src/utils/datetime.ts
```diff
@@ -11,3 +17,46 @@ export const combineDateTime = (date: string, time?: string | null) => {
...
+  endTime: string | null
+) => {
+  const startDateTime = startTime ? dayjs(`${startDate} ${startTime}`) : dayjs(startDate);
+  const endDateTime = endTime ? dayjs(`${endDate} ${endTime}`) : dayjs(endDate);
+
+  // Calculate the difference in milliseconds
+  const diff = endDateTime.diff(startDateTime);
+
+  // Calculate years, days, hours, minutes, and seconds
+  const years = Math.floor(diff / (1000 * 60 * 60 * 24 * 365));
```
MacDeluca: suggestion: You can use an additional param on diff to get specific date differences.
[DayJS Diff](https://day.js.org/docs/en/display/difference)

```js
const days = endDateTime.diff(startDateTime, 'd'); // 123
```

app/src/utils/datetime.ts
```diff
@@ -1,3 +1,9 @@
+import dayjs from 'dayjs';
+import duration from 'dayjs/plugin/duration';
+import { pluralize } from './Utils';
+
+dayjs.extend(duration);
```
NickPhura, line 7: Is this being leveraged at all?

app/src/utils/datetime.ts
mauberti-bc: oh gosh, that simplifies this. Thanks!

app/src/utils/datetime.ts
NickPhura: No idea if this is sufficient or not, but noticed this plugin: https://day.js.org/docs/en/plugin/relative-time

mauberti-bc: Oops, leftover from an earlier version

NickPhura:
We should add some unit tests as well, for `datetime.ts`. 
They should be fairly straightforward to write, since there is no rendering or api calls to worry about.

mauberti-bc:
> We should add some unit tests as well, for `datetime.ts`. They should be fairly straightforward to write, since there is no rendering or api calls to worry about.

Roger, I pushed up a test file for datetime.ts. Some tests are failing due to a timezone issue, where the test framework seems to treat local time as PDT (UTC-8) instead of PST (UTC-7) despite the main app correctly using PST. Haven't been able to figure it out 😖 (We will switch to PDT on Nov. 3 when we roll back the clocks)

MacDeluca (approved):
🌡️

PR #1425, opened 2024-11-12
TechDebt: API: Environment Variables Validation

MacDeluca:
## Description of Changes

Note: In my previous version `env-config.ts` exported a getter and setter for ENV values. I think the added complexity made it not worthwhile. 
Now you get and set `process.env` values the regular way with the addition of `process.env` being fully typed. The app will crash on start up if the environment variables are not set correctly, still undecided if the `loadEnvironmentVariables` function should always exit the process, or only for `development` environments instead.

- Validates environment variables against a zod schema
- Extends the `process.env` type to include the inferred zod type
- Updates some existing casting and default values

## Testing Notes

- App should work normally (when all environment variables exist in .env)
- Need to validate that all the values in the schema are actually required. Should any be set to optional?


api/src/app.ts
```diff
@@ -13,20 +13,24 @@ import {
...
 import { rootAPIDoc } from './openapi/root-api-doc';
 import { authenticateRequest, authenticateRequestOptional } from './request-handlers/security/authentication';
+import { loadEvironmentVariables } from './utils/env-config';
 import { scanFileForVirus } from './utils/file-utils';
 import { getLogger } from './utils/logger';
 
 const defaultLog = getLogger('app');
 
+// Load and validate the environment variables
+loadEvironmentVariables();
```
NickPhura, line 21: Does this need to be called before calling `getLogger()` above?

api/src/utils/env-config.ts
```diff
@@ -0,0 +1,122 @@
...
+  GCNOTIFY_EMAIL_URL: ZodEnvString,
+  GCNOTIFY_SMS_URL: ZodEnvString,
+
+  // ClamAV
+  CLAMAV_PORT: z.coerce.number(),
+  CLAMAV_HOST: ZodEnvString,
+  ENABLE_FILE_VIRUS_SCAN: z.enum(['true', 'false']),
+
+  // Extra
+  FEATURE_FLAGS: ZodEnvString.optional() // flagA,flagB,flagC
```
NickPhura: This seems to fail when my `.env` includes `FEATURE_FLAGS=` but it has no value. I think because it still resolves to an empty string, which fails the `min` length check.

Might have to just be `z.string().optional()`

api/src/utils/env-config.ts
```diff
@@ -0,0 +1,122 @@
...
+// Schema for environment configuration
+export const EnvSchema = z.object({
+  // Environment
+  NODE_ENV: z.enum(['development', 'test', 'production']),
+  NODE_OPTIONS: ZodEnvString,
+  TZ: z.literal('America/Vancouver'),
+
+  // API server
+  API_HOST: ZodEnvString,
+  API_PORT: z.coerce.number(),
```
NickPhura: Should some of these number fields include a `required` constraint (similar to the string ones)?

NickPhura:
Looks great!

api/src/app.ts
MacDeluca: I don't think it "needs" to, but calling it higher up is probably preferred. It will crash regardless if the .env is not to spec.

Updated: This is now called first

api/src/utils/env-config.ts
MacDeluca: Oh nice catch, updated.

api/src/utils/env-config.ts
MacDeluca: I added a schema for numbers that is more appropriate. 
ie: min of 0.

api/src/app.ts
NickPhura: Right, ya I guess the process.env values still exist, they just havent been verified yet

NickPhura (approved):
🕊️

PR #1426, opened 2024-11-13
SIMSBIOHUB-449: Upgrade Postgres from 12.5 to 17 (WIP)

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-449

## Description of Changes

Upgrade local and remote postgres versions from 12.5 to 15.

## Testing Notes

Database builds/deploys as usual.

Existing data produced under version 12.x gets picked up by the new version correctly. Worst case scenario we need to manually migrate the data, but ideally it just works without issues.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -23,6 +23,7 @@ services:
       - POSTGRES_PASSWORD=${DB_ADMIN_PASS}
       - POSTGRES_DB=${DB_DATABASE}
       - PORT=5432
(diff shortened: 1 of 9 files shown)


NickPhura:
Closing and reopening because openshift wont re-create the image stream with tag '1426' for some reason (other numbers work...)

PR #1427, opened 2024-11-14
Add Tooltips to Form Fields

AMEIJER1:
## Links to Jira Tickets

- N/A

## Description of Changes

- Adds a tooltip to the input fields (AutocompleteField, CustomTextField, etc.) that can be shown when a "helpText" prop is given


mauberti-bc:
@NickPhura There's a failing frontend test that I couldn't figure out how to fix

NickPhura (dismissed):
Very nice! Seems to work really well.

It could be interesting to introduce a bit of a delay in the onMouseLeave so that if the user's mouse slips off the button briefly, and then moves back onto it, the tooltip doesn't immediately start closing and then re-opening. But, certainly not an essential feature, assuming its also not annoying to implement.

NickPhura (approved):

PR #1429, opened 2024-11-15
BugFix: Update ENV Values

MacDeluca:
## Links to Jira Tickets

## Description of Changes

## Testing Notes

### DIFF
--- a/api/.pipeline/lib/api.deploy.js
+++ b/api/.pipeline/lib/api.deploy.js
@@ -37,6 +37,7 @@ const apiDeploy = async (settings) => {
         // Persistent Volume
         VOLUME_CAPACITY: phases[phase].volumeCapacity,
         // BioHub Platform (aka: Backbone)
+        BACKBONE_PUBLIC_API_HOST: phases[phase].backbonePublicApiHost,
         BACKBONE_INTERNAL_API_HOST: phases[phase].backboneInternalApiHost,
         BACKBONE_INTAKE_PATH: phases[phase].backboneIntakePath,
         BACKBONE_ARTIFACT_INTAKE_PATH: phases[phase].backboneArtifactIntakePath,
--- a/api/.pipeline/lib/api.deploy.js
+++ b/api/.pipeline/lib/api.deploy.js
@@ -37,7 +37,6 @@ const apiDeploy = async (settings) => {
         // Persistent Volume
         VOLUME_CAPACITY: phases[phase].volumeCapacity,
         // BioHub Platform (aka: Backbone)
-        BACKBONE_PUBLIC_API_HOST: phases[phase].backbonePublicApiHost,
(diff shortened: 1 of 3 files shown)


api/.pipeline/templates/api.dc.yaml
```diff
@@ -336,6 +346,13 @@ objects:
                   value: ${KEYCLOAK_API_HOST}
                 - name: KEYCLOAK_API_ENVIRONMENT
                   value: ${KEYCLOAK_API_ENVIRONMENT}
+                # Request limits
+                # - name: MAX_REQ_BODY_SIZE
+                #   value: ${MAX_REQ_BODY_SIZE}
```
NickPhura: Commented out

api/.pipeline/templates/api.dc.yaml
```diff
@@ -336,6 +346,13 @@ objects:
...
+                - name: MAX_UPLOAD_NUM_FILES
+                  value: MAX_UPLOAD_NUM_FILES
```
NickPhura: Probably needs to be wrapped in ${}

api/.pipeline/templates/api.dc.yaml
MacDeluca: Intentional, just testing it fails when it's missing an env value

api/.pipeline/lib/api.deploy.js
```diff
@@ -37,6 +37,7 @@ const apiDeploy = async (settings) => {
         // Persistent Volume
         VOLUME_CAPACITY: phases[phase].volumeCapacity,
         // BioHub Platform (aka: Backbone)
+        BACKBONE_PUBLIC_API_HOST: phases[phase].backbonePublicApiHost,
```
NickPhura: Hmm, it seems like this is never actually used in the API anywhere. Maybe we should just remove it here, and remove it from the env-config schema

NickPhura: Its passed into the compose.yml#api too, but I guess is never referenced after that

api/.pipeline/templates/api.dc.yaml
NickPhura: Oh, unless you're testing it fails :)

api/.pipeline/templates/api.dc.yaml
MacDeluca: Ooops my bad, nice catch

api/.pipeline/lib/api.deploy.js
MacDeluca: Yea doesn't look like it hey. Dropping.

api/.pipeline/templates/api.dc.yaml
NickPhura: Oh weird, I thought I wrote my second message before you commented, but I guess I didnt refresh or something

NickPhura (approved):

PR #1438, opened 2024-11-28
SIMSBIOHUB-633: Revamp CSV Imports - Critter

MacDeluca:
## Links to Jira Tickets

- [SIMSBIOHUB-633: Revamp CSV Imports](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-633)

## Description of Changes

- Revamps the existing CSV import services to an improved structure
- New CSVConfig concept - Static / Dynamic header configs
- Supports ALL data types
- New validation utils / functions for validating CSVConfigs
- New ImportCrittersService for importing Critter CSVs
- Util files for cellValidator and cellSetter functions
- CSV validation now returns CSVErrors

## Testing Notes

- Should be able to import a Critter CSV with Postman
- Should see validation errors when issues exist in the CSV
- Correcting the issues and re-importing will succeed


api/src/services/import-services/critter/import-critters-service.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    // Add critters to Critterbase
+    const bulkResponse = await this.critterbaseService.bulkCreate(payloads.critterbasePayload);
+
+    // Check critterbase inserted the full list of critters
+    // In reality this error should not be triggered, safeguard to prevent floating critter ids in SIMS
+    if (bulkResponse.created.critters !== payloads.simsPayload.length) {
+      throw new ApiGeneralError('Unable to fully import critters from CSV', [
+        'importCrittersStrategy -> insertCsvCrittersIntoSimsAndCritterbase',
+        'critterbase bulk create response count !== critterIds.length'
```
NickPhura: nit: old error message

api/src/utils/csv-utils/csv-header-configs.ts
```diff
@@ -0,0 +1,113 @@
...
+ * Rules:
+ *  1. The cell must be a number greater than or equal to 0
+ *  2. The cell must be a real ITIS TSN (from the provided set)
+ *
+ * @param {Set<number>} tsns Set of allowed ITIS TSNs
+ * @returns {*} {(params: CSVParams) => CSVError[]} The validate cell callback
+ */
+export const getTsnCellValidator = (tsns: Set<number>): ((params: CSVParams) => CSVError[]) => {
+  return (params: CSVParams) => {
+    const cellErrors = validateZodCell(params, z.number().min(0));
```
NickPhura: Is this check required? 

We already have a list of allowed tsns, which have come from the biohub endpoint. So if the `Number(params.cell)` value below is in that list, then it would have to be valid

NickPhura: I guess maybe it depends on the error messaging.

if the user enters "moose", do we tell them "must be a number" or just always say "not a valid tsn"

NickPhura: Might be a question for Macgregor, what the error message should be and matching solution text.

api/src/services/import-services/critter/critter-header-configs.ts
```diff
@@ -0,0 +1,183 @@
...
+      });
+    }
+
+    return cellErrors;
+  };
+};
+
+/**
+ * Get the critter collection unit cell validator.
+ *
```
NickPhura, line 60: nit: could add the validator rules to the jsdoc (like in some of the other functions)

api/src/services/import-services/critter/critter-header-configs.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (cellErrors.length || !params.cell) {
+      return cellErrors;
+    }
+
+    const rowTsn = Number(configUtils.getCellValue('ITIS_TSN', params.row));
+    const cellValue = String(params.cell).toLowerCase();
+
+    const rowDictionaryTsn = get(rowDictionary, rowTsn);
+    const rowDictionaryHeader = get(rowDictionary, [rowTsn, params.header]);
+    const rowDictionaryUnit = get(rowDictionary, [rowTsn, params.header, cellValue]);
```
NickPhura: minor nit: If `rowDictionaryTsn` is invalid, you won't need to have calculated `rowDictionaryHeader` or `rowDictionaryUnit`, etc.

So you could move those down, after each check, and then just make them all `if` statements with their own returns

NickPhura: Similar for the `getCritterSexCellValidator` function

api/src/services/import-services/critter/critter-header-configs.ts
```diff
@@ -0,0 +1,183 @@
...
+ */
+export const getCritterSexCellSetter = (
+  rowDictionary: { [tsn: number]: { [sex: string]: string } },
+  configUtils: CSVConfigUtils<CritterCSVConfig>
+): ((params: CSVParams) => string) => {
+  return (params: CSVParams) => {
+    const tsn = Number(configUtils.getCellValue('ITIS_TSN', params.row));
+    const cellValue = String(params.cell).toLowerCase();
+
+    return get(rowDictionary, [tsn, cellValue]);
```
NickPhura: Some extra comments or jsdoc might be helpful here. Its a bit hard to understand whats happening here, especially because the `tsn` const is calling `getCellValue` but the other const is called `cellValue`. And the lodash function doesnt include any jsdoc on hover.

api/src/utils/csv-utils/csv-config-validation.interface.ts
```diff
@@ -0,0 +1,108 @@
...
+   * Record containing the static headers, their aliases, and the `validateCell` and `setCellValue` callbacks
+   * to be called for each static cell.
+   *
+   * @type {Record<THeader, { aliases: Uppercase<string> } & CSVHeaderConfig>}
+   */
+  staticHeadersConfig: Record<THeader, { aliases: Uppercase<string> } & CSVHeaderConfig>;
+  /**
+   * Contains the `validateCell` and `setCellValue` callbacks to be called for each dynamic cell.
+   *
+   * @type {CSVHeaderConfig | undefined}
```
NickPhura: Might be worth explaining what we consider a static vs dynamic cell/header

api/src/utils/csv-utils/csv-config-validation.ts
```diff
@@ -0,0 +1,179 @@
...
+};
+
+/**
+ * Validate the CSV cells against the CSV config.
+ *
+ * @param {WorkSheet} worksheet - The worksheet
+ * @param {CSVConfig} config - The CSV configuration
+ * @returns {CSVError[]} - The CSV errors
+ */
+export const validateCSVCells = (worksheet: WorkSheet, config: CSVConfig): CSVError[] => {
```
NickPhura: nit: should this be prefixed with an underscore?

Same with the `validateCSVCells` function below.

api/src/utils/csv-utils/csv-config-validation.ts
```diff
@@ -0,0 +1,179 @@
...
+ * @param {CSVConfig} config - The CSV configuration
+ * @returns {CSVRow[]} - The CSV rows
+ */
+export const setCSVCellValues = (worksheet: WorkSheet, config: CSVConfig): CSVRow[] => {
+  const rows: CSVRow[] = [];
+
+  _forEachCSVCell(worksheet, config, (params, staticHeaderConfig) => {
+    if (staticHeaderConfig && staticHeaderConfig.setCellValue) {
+      delete params.row[params.header];
+      params.row[staticHeaderConfig.staticHeader.toUpperCase()] = staticHeaderConfig.setCellValue(params);
```
NickPhura: Just curious: any reason for deleting the value first, rather than just overriding it? Unless the key also needs replacing, not just the value?

api/src/utils/csv-utils/csv-config-validation.interface.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * The cell header name.
+   * @type {string}
+   */
+  header: string;
+  /**
+   * The cell row index.
+   * @type {number}
+   */
+  rowIndex: number;
```
NickPhura: Might be worth specifying where this index starts so we know if it needs to be offset when being used in log messages, etc. (just because we generally start at 0, but in an excel file the first real row is 2).

NickPhura:
Looks really good overall. 🦕 

Was just reading through the code so far. 

My only real note (which I included some comments about) is just adding more comments/jsdoc, since some of the functions or steps within the functions can be a bit complicated, so adding some helper comments would be great.

api/src/services/import-services/critter/import-critters-service.ts
MacDeluca: Fixed

api/src/services/import-services/critter/critter-header-configs.ts
MacDeluca: Ok added :)

api/src/utils/csv-utils/csv-config-validation.interface.ts
MacDeluca: Oops I totally forgot to come back to this. Updated.

api/src/utils/csv-utils/csv-config-validation.ts
MacDeluca: Yea potentially both the key and value get replaced.

It reforms the rows into the staticHeader structure, so we can easily grab out the values when inserting the validated data.

api/src/services/import-services/critter/critter-header-configs.ts
MacDeluca: Updated

api/src/utils/csv-utils/csv-config-validation.interface.ts
MacDeluca: Updated

api/src/services/import-services/critter/critter-header-configs.ts
MacDeluca: I updated the jsdoc and renamed some variables for clarity.

Note: The rowDictionary was my best attempt at handling these types of validation relationships, but I'm open to additional solutions if theres a better data structure be it native or custom that I didn't think of

api/src/services/import-services/critter/import-critters-service.ts
```diff
@@ -0,0 +1,306 @@
...
+   */
+  constructor(connection: IDBConnection, worksheet: WorkSheet, surveyId: number) {
+    super(connection);
+
+    this._config = {
+      staticHeadersConfig: {
+        ITIS_TSN: { aliases: ['TAXON', 'SPECIES', 'TSN'] },
+        ALIAS: { aliases: ['NICKNAME', 'NAME', 'ANIMAL_ID'] },
+        SEX: { aliases: ['TEST'] },
+        WLH_ID: { aliases: ['WILDLIFE_HEALTH_ID'], validateCell: getWlhIDCellValidator() },
```
mauberti-bc: We should include "Wildlife Health ID" with spaces and "WLHID" as aliases

api/src/services/import-services/critter/import-critters-service.ts
```diff
@@ -0,0 +1,306 @@
...
+    return {
+      validateCell: getTsnCellValidator(allowedTsns)
+    };
+  }
+
+  /**
+   * Get the CSV Alias header config.
+   *
+   * Validation rules:
+   *  1. Alias must be a string
```
mauberti-bc, line 218: Alias can be a number if the animal is named "2", for example

api/src/services/import-services/critter/import-critters-service.ts
MacDeluca: Updated

api/src/services/import-services/critter/import-critters-service.ts
MacDeluca: Updated

mauberti-bc:
Looks great! It's just missing the ability to make headers optional - wildlife health ID and description should be optional when adding animals

* Add ability to make certain headers optional
* Sex should be nullable and optional (able to be omitted as a header)

mauberti-bc (approved):
👍

PR #1441, opened 2024-12-03
BugFix: Change Date Format from UTC to Local Time in combineDatetime

mauberti-bc:
## Links to Jira Tickets

## Description of Changes

- The combineDatetime function is only used to combine local date and times, but the format string was treating the local date and time as UTC, yielding incorrect outputs
- Removed the ```[Z]``` from the end of the format string

## Testing Notes

- When creating an animal capture/mortality event, the displayed date and time should match the event's true date and time (the API responds with local time so the timestamp in the API response should be what is shown on the capture/mortality card)


MacDeluca (approved):
🚢

PR #1442, opened 2024-12-03
BugFix: Allow Geojson of Survey Blocks to be Null

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Updates OpenAPI schema to let the geojson property of a survey block to be null
- Adds blocks to the seed data
- This fixes the bug causing the {surveyId}/view endpoint to fail in dev

### DIFF
--- a/api/src/repositories/survey-block-repository.ts
+++ b/api/src/repositories/survey-block-repository.ts
@@ -19,7 +19,7 @@ export const SurveyBlockRecord = z.object({
   survey_id: z.number(),
   name: z.string(),
   description: z.string(),
-  geojson: z.any(),
+  geojson: z.any().nullable(),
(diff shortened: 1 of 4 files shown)


api/src/repositories/survey-block-repository.ts
```diff
@@ -19,7 +19,7 @@ export const SurveyBlockRecord = z.object({
   survey_id: z.number(),
   name: z.string(),
   description: z.string(),
-  geojson: z.any(),
+  geojson: z.any().nullable(),
```
MacDeluca, line 22: praise: Technically this nullable is not needed (any will override) but I think this is good for readability. I would also write it like this.

MacDeluca (approved):
🚀

PR #1444, opened 2024-12-03
SIMSBIOHUB-647: Create/Edit Observations Page

mauberti-bc:
## Links to Jira Tickets

## Description of Changes

- Allows observations to be added from a dedicated page
- Allows observations to be edited from a dedicated page

## Testing Notes


app/src/hooks/api/useObservationApi.ts
```diff
@@ -68,6 +42,22 @@ const useObservationApi = (axios: AxiosInstance) => {
...
 
+  /**
+   * Creates a new observation for the survey
+   *
+   * @param {number} projectId
+   * @param {number} surveyId
+   * @param {IObservationTableRowToSave[]} surveyObservations
+   * @return {*}  {Promise<void>}
+   */
+  const createObservation = async (
```
NickPhura: It looks like this function (and the matching endpoint) is identical to the `insertUpdateObservationRecords` function + matching `PUT` endpoint.

app/src/constants/i18n.ts
```diff
@@ -533,3 +533,9 @@ export const AlertI18N = {
   deleteAlertDialogTitle: 'Delete Alert?',
   deleteAlertDialogText: 'Are you sure you want to permanently delete this alert? This action cannot be undone.'
 };
+
+export const CreateObservationI18N = {
+  createErrorTitle: 'Error Exporting Survey Data',
+  createErrorText:
+    'An error has occurred while attempting to export survey data. Please try again. If the error persists, please contact your system administrator.'
+};
```
NickPhura: Wrong text, presumably just leftover from copy/paste

app/src/hooks/api/useObservationApi.ts
mauberti-bc: I was partway through separating the insertUpdate logic into separate insert and update functions - might be unnecessary but I was thinking that the original insertUpdateObservationRecords would just handle updates (haven't implemented that yet).

NickPhura: Gotcha. Yeah if all new observations are created through the new page, then the table would just be doing updates. So that makes sense.

app/src/features/surveys/observations/create/CreateObservationPage.tsx
```diff
@@ -0,0 +1,282 @@
...
+                  ? {
+                      environment_qualitative_id: environment.environment_qualitative_id,
+                      environment_qualitative_option_id: environment?.environment_qualitative_option_id
+                    }
+                  : undefined
+              )
+              .filter((environment) => environment?.environment_qualitative_id !== undefined)
+          };
+        })
+      } as ICreateObservationRequest;
```
NickPhura: Rather than cast this to the type, we should just assign this to the const.

```
const data: ICreateObservationRequest = { ... }
```
Otherwise, there is no actual type safety. And actually if you make that change, the type isn't happy.

Partially related: It looks like the comment field is in the wrong spot. The form assigns it to the standard columns section, but its being added here under the subcounts section.

NickPhura:
I pushed up a few misc changes. Mostly just some component cleanup and addressing a few react dependency complaints, etc.

app/src/constants/i18n.ts
NickPhura: fixed

app/src/features/surveys/observations/create/CreateObservationPage.tsx
NickPhura: fixed

api/src/database-units/environment_unit.ts
MacDeluca, line 1: question: Do you think we should name this directory `database-units` or something like `database-types / database-enums`. I see in the database there is an additional type / enum for `alert_severity` would this also be a `database-unit`?

api/src/openapi/schemas/observation.ts
```diff
@@ -1,5 +1,102 @@
...
+  type: 'object',
+  additionalProperties: false,
+  required: [
+    'observation_environment_qualitative_id',
+    'environment_qualitative_id',
+    'environment_qualitative_option_id'
+  ],
+  properties: {
+    observation_environment_qualitative_id: {
+      type: 'integer'
```
MacDeluca: nit: minimum 1

api/src/openapi/schemas/observation.ts
```diff
@@ -1,5 +1,102 @@
...
+  }
+};
+
+const ObservationQuantitativeEnvironment: OpenAPIV3.SchemaObject = {
+  type: 'object',
+  additionalProperties: false,
+  required: ['observation_environment_quantitative_id', 'environment_quantitative_id', 'value'],
+  properties: {
+    observation_environment_quantitative_id: {
+      type: 'integer'
```
MacDeluca: nit: minimum 1

api/src/openapi/schemas/observation.ts
```diff
@@ -1,5 +1,102 @@
...
+
+const ObservationSubcountSchema: OpenAPIV3.SchemaObject = {
+  type: 'object',
+  description:
+    'An observation subcount record. Each subcount defines additional attributes/details about a subset of the observed taxa in the observation.',
+  additionalProperties: false,
+  required: ['observation_subcount_id', 'subcount', 'comment', 'qualitative_measurements', 'quantitative_measurements'],
+  properties: {
+    observation_subcount_id: {
+      type: 'integer'
```
MacDeluca: nit: minimum 1

api/src/paths/observation/flattened/index.ts
```diff
@@ -0,0 +1,252 @@
...
+    },
+    {
+      in: 'query',
+      name: 'itis_tsns',
+      description: 'ITIS TSN numbers',
+      required: false,
+      schema: {
+        type: 'array',
+        items: {
+          type: 'integer'
```
MacDeluca: nit: minimum 1 / 0?

api/src/paths/observation/flattened/index.ts
```diff
@@ -0,0 +1,252 @@
...
+        nullable: true
+      }
+    },
+    {
+      in: 'query',
+      name: 'itis_tsn',
+      description: 'ITIS TSN number',
+      required: false,
+      schema: {
+        type: 'integer',
```
MacDeluca, line 71: nit: minimum 1 / 0?

api/src/database-units/environment_unit.ts
NickPhura: Open to any name for the folder. Can't remember how I landed on `unit`. Probably just trying to pick something generic that could cover database enums (but also any other enum-like concept, if there are others?), and which aren't specifically table models.

api/src/services/observation-services/observation-service.ts
```diff
@@ -57,21 +49,47 @@ export interface InsertSubCount {
...
-    value: number;
-  }[];
 }
 
 export type InsertUpdateObservations = {
   standardColumns: InsertObservation | UpdateObservation;
   subcounts: InsertSubCount[];
 };
 
+export type InsertObservations = {
```
MacDeluca: question: Where did we land with this, is this still being used or do we only need `InsertUpdateObservations`

app/src/components/data-grid/GenericGridColumnDefinitions.tsx
```diff
@@ -297,7 +302,11 @@ export const GenericCommentColDef = <T extends GridValidRowModel>(props: {
       return (
         <Tooltip title={params.value} arrow>
           <span>
-            <IconButton aria-label="observation-subcount-comment" disabled>
+            <IconButton
+              aria-label="observation-subcount-comment"
```
MacDeluca: nit: label is specific when it should be generic

api/src/services/observation-services/observation-service.ts
NickPhura: I believe we will eventually remove the `InsertUpdate` function, in favour of separate insert and update functions.

But, we haven't implemented the new edit observation flow yet. We will probably have to add the edit flow to this PR, or a feature branch, before merging any of this to Dev. So this function will probably get removed as part of that effort.

app/src/components/data-grid/GenericGridColumnDefinitions.tsx
NickPhura: Fixed

mauberti-bc:
Looks great! Some additions:
* Include observation ID column in the manage observations table to show which subcounts are related
* Sign should default to direct sighting (just as a convenience since this is the most common value)

MacDeluca (approved):
Create and edit seem to be working correctly!

PR #1445, opened 2024-12-04
Create Support Page

LouisThedroux:
## Description of Changes
- N/A

## Description of Changes

* Adds Support Page with information about how to use the system


mauberti-bc:
Check out the "106 issues" in the Quality Gate comment on this PR page. Sonar cloud identifies patterns in the code that are considered bad, like using the index as the key prop within map(), or missing the key prop (in which case React cannot reliably distinguish between UI elements when trying to smartly re-render the component upon some change).

MacDeluca (approved):
💯 Code looks great!

PR #1446, opened 2024-12-05
SIMSBIOHUB-648: Create Sampling Period Page

mauberti-bc:
## Links to Jira Tickets

## Description of Changes

The main goal of this PR is to separate the sampling site and sampling period CRUD operations. Previously sampling periods were created/updated as part of creating/updating a sampling site. Now, sampling periods have their own create/edit pages/components.

All of the changes are around that concept, handling all of the code that naturally broke as a result of the related database schema changes, and workflow changes, etc.

### New DB Schema
- 1 new migration
  - See JSDoc notes in the migration.
- 1 new procedure
  -  Check constraint for the sample_period table.

Picture of the relevant tables, and how they look after running the migration:

### New Functionality
- Separate table/page for creating, editing, deleting sampling periods (via the Manage Sampling Information page).
  - New endpoints/services/repositories/interfaces

### Updated Functionality
- Updated the 3 sampling columns in the Observations table. Behaviour is similar to before, but underlying data is different.
- Updated the sampling side bar (from the Manage Observations page). Behaviour is similar to before, but underlying data is different.
- Updated Analytics tab (below the map on the Survey page). Behaviour is similar to before, but underlying data is different.
- Updated the sampling information portion of the existing Observation CSV import backend code. The rest of the Observation CSV import code should work as before.


NickPhura:
A few things I noticed:

---

When creating a period, if you select a site, and then remove the period from it, the validation is probably failing because the form won't submit, but there is no error message shown.

image

---

When editing an existing sample period, I can't select or change the sample sites.

image

---

Nit: I can save a period with a start and end date that are exactly the same (with no time). 

image

If I pick the same times, then the save stops working, though there is no error message about why.

---

I have 1 sampling site, 1 technique, and 3 periods. The techniques aren't being combined in the observations table when adding a record:

image

mauberti-bc:
- I disallowed changing the site because at that point you may as well create a new period for the correct site and delete the wrong one. Seems along the lines of turning a skateboard into a car instead of just buying a car. Maybe worth allowing though? 
- We do want to allow a period that starts and stops on the same day with no time, in case the time is not known. I agree this seems off but is realistic data. (The duration column is null for those periods because there isn't an obvious way to represent the unknown duration)

mauberti-bc:
image
There's a failing request when expanding a sampling site in the left side panel of the observations page. Otherwise haven't found any issues

mauberti-bc:
* No observations are being returned on the observations table page, but the request succeeds
* When you select a site or technique, then remove that site or technique from the form, that option does not return to the autocomplete options (you have to refresh the autocomplete for it to show up). 
* We should also load the sampling site and technique autocompletes when the component mounts, before any input.

MacDeluca:
Heres some screenshots of some of the issues I found:

1. When deleting the sampling site from the "Manage Observations" page it throws a failed to execute SQL error: column "survey_sample_site_id" does not exist
image

2. The sampling techniques table renders the wrong page when the rows overflow into the next page. (renders page 2 vs rendering page 1) when 11 sampling techniques http://localhost:7100/admin/projects/2/surveys/4/sampling
image


3. The survey page Sampling Information (Technique tab) says 11 records but renders nothing on page 2 http://localhost:7100/admin/projects/2/surveys/4/details
image

NickPhura:
> Heres some screenshots of some of the issues I found:

Pushed up some changes that should address those 3 issues.

database/src/procedures/tr_before_survey_sample_period_check_minimum_data_requirements.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function seed(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
+    SET search_path = 'biohub';
+
+    CREATE OR REPLACE FUNCTION biohub.check_survey_sample_period_minimum_data_requirements()
```
MacDeluca: suggestion: We might be able to replace this trigger with a CHECK contraint using `num_nonnulls`
https://stackoverflow.com/questions/15178859/postgres-constraint-ensuring-one-column-of-many-is-present

NickPhura: Will try it out

api/src/repositories/sample-site-repository/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+          .groupBy('ssst.survey_sample_site_id');
+      })
+      // Fetch sample sites and include the corresponding sample methods, blocks, and stratums
+      .select(
+        'sss.survey_sample_site_id',
+        'sss.survey_id',
+        'sss.name',
+        'sss.description',
+        'sss.geojson',
+        // 'technique',
```
MacDeluca: 💀

api/src/services/geo-service.ts
```diff
@@ -228,6 +230,13 @@ export class GeoService {
 
   constructor(options?: { baseUrl?: string }) {
     this.baseUrl = options?.baseUrl ?? process.env.BcgwBaseUrl ?? 'https://openmaps.gov.bc.ca/geo/pub/ows';
+
+    axios.interceptors.response.use(
+      (response) => response,
+      (error) => {
+        throw new ApiGeneralError('Failed to geospatial data from external service.', [formatAxiosError(error)]);
```
MacDeluca: grammar mistake: `Failed to retrieve geospatial data from external service`

database/src/procedures/tr_before_survey_sample_period_check_minimum_data_requirements.ts
NickPhura: pushed up new migration (20250117000001_survey_sample_period_check_constraints) and dropped the trigger

app/src/features/surveys/sampling-information/techniques/edit/EditTechniquePage.tsx
```diff
@@ -71,6 +72,7 @@ export const EditTechniquePage = () => {
     description: technique?.description ?? null,
     distance_threshold: technique?.distance_threshold ?? null,
     method_lookup_id: technique?.method_lookup_id ?? null,
+    method_response_metric_id: 1, // NICK TODO: This is a placeholder value.
```
MacDeluca: nit: Still needed? Can set to `-1` to prevent issues later.

app/src/features/surveys/sampling-information/techniques/create/CreateTechniquePage.tsx
```diff
@@ -63,6 +65,7 @@ export const CreateTechniquePage = () => {
       const createTechniqueRequestData: ICreateTechniqueRequest = {
         ...values,
         distance_threshold: values.distance_threshold || null,
+        method_response_metric_id: 1, // NICK TODO: This is a placeholder value.
```
MacDeluca: nit: Same as above comment

database/src/migrations/20250117000001_survey_sample_period_check_constraints.ts
```diff
@@ -0,0 +1,40 @@
...
+     * table. Specifically, checks that at least one of the following items is not null:
+     * - survey_sample_site_id
+     * - method_technique_id
+     * - start_date, end_date (either both null or both not null)
+     */
+    ALTER TABLE survey_sample_period ADD CONSTRAINT check_minimum_data_requirement CHECK (
+      NUM_NONNULLS(survey_sample_site_id, method_technique_id, start_date) >= 1
+    );
+
+    COMMENT ON CONSTRAINT check_minimum_data_requirement ON survey_sample_period IS 'Checks that at least one of the following items is not null: survey_sample_site_id, method_technique_id, start_date + end_date. Note: This constraint only checks start_date because the check_start_and_end_date_requirement constraint already ensures that start_date and end_date are either both null or both not null.';
```
MacDeluca, line 34: praise: Hell yeah 🎸

app/src/features/surveys/sampling-information/techniques/edit/EditTechniquePage.tsx
NickPhura: Will check with @mauberti-bc. The technique table has a non-null foreign key to the method_response_metric table, but it appears the UI no longer has the field for entering that value. Setting to `1` sets it to `count` (albeit unsafely since the ids aren't stable). Probably need to either re-add that form control so it can be set by the user, or else make that field nullable and pass null here instead. Or get the value from the codes, and still set it to the id of the `count` code, which will be the same result but safer.

Note to self:
- Check both edit and create technique pages
- Check `ObservationCountDataGridEditCell` as it is also setting this value to `1`

api/src/services/geo-service.ts
NickPhura: Fixed!

api/src/repositories/sample-site-repository/utils.ts
NickPhura: Removed

app/src/features/surveys/sampling-information/techniques/edit/EditTechniquePage.tsx
NickPhura: Fixed. Added proper form control to populate this field.

app/src/features/surveys/sampling-information/techniques/create/CreateTechniquePage.tsx
NickPhura: Fixed. Added proper form control to populate this field.

api/src/services/import-services/utils/datetime.ts
```diff
@@ -21,15 +22,15 @@ interface ICaptureStub {
  * @returns {string | undefined}
  */
 export const formatTimeString = (time?: string | null): string | undefined => {
-  const fullTime = dayjs(time, 'HH:mm:ss');
+  const fullTime = dayjs(time, DefaultTimeFormat);
   const shortTime = dayjs(time, 'HH:mm');
```
MacDeluca: nit: This can be updated to the new `DefaultTimeFormatNoSeconds`

NickPhura: Updated

MacDeluca (approved):
👍🏼

PR #1448, opened 2024-12-06
BugFix: Make Environments Case Insensitive in Observations Import

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Removes case sensitivity when importing environments as part of observations
- Fixes a SQL query to allow multiple environments of the same type to be imported together

## Testing Notes

- Should be able to import observations with multiple environment variables irrespective of case


MacDeluca (approved):
🩴

PR #1451, opened 2024-12-11
SIMSBIOHUB-656: Import CSV Markings

MacDeluca:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-656

## Description of Changes

- CSV import for Critter markings
- New CSVErrors table for rendering validation errors

## Testing Notes
http://localhost:7100/admin/projects/2/surveys/4/animals/captures
- Should be able to import a markings CSV into SIMS
- Critters and captures will need to be created before markings can be imported

image


app/src/utils/file-utils.ts
MacDeluca, line 1: Note: Is there a better file location (frontend) for these types / values. 
ie: CSV_VALIDATION_ERROR / CSVError / CSVValidationError

NickPhura: Could make a separate file for csv related stuff:
- csv-utils.ts
- csv-validation.ts
- csv-import.ts
- etc

api/src/services/import-services/marking/import-markings-service.ts
```diff
@@ -0,0 +1,201 @@
...
+ */
+export class ImportMarkingsService extends DBService {
+  worksheet: WorkSheet;
+  surveyId: number;
+
+  surveyCritterService: SurveyCritterService;
+  utils: CSVConfigUtils<MarkingCSVStaticHeader>;
+
+  /**
+   * Construct an instance of ImportMarkingsStrategy.
```
NickPhura: nit: 'strategy'

api/src/services/import-services/marking/marking-header-configs.ts
```diff
@@ -0,0 +1,237 @@
...
+ * @param {Set<string>} markingTypes The marking types
+ * @returns {*} {CSVCellValidator} The validate cell callback
+ */
+export const getMarkingTypeCellValidator = (markingTypes: Set<string>): CSVCellValidator => {
+  return (params: CSVParams) => {
+    if (params.cell === undefined) {
+      return [];
+    }
+
+    if (!markingTypes.has(String(params.cell).toLowerCase())) {
```
NickPhura: In the parent, the markings are transformed to lowercase. Wondering if that should be done at this level instead, so the caller doesn't have to know what case the values need to be in.

NickPhura: Though, I guess that is less performant since it would be done every time then. In that case, maybe just add a jsdoc comment that the values will be compared in lowercase and must be transformed

NickPhura: Same for the other string based validators

api/src/utils/csv-utils/csv-config-utils.ts
```diff
@@ -118,16 +118,42 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string>> {
...
   /**
    * Get the cell value from a CSV row.
    *
    * @param {StaticHeaderType} header - The header name
    * @param {CSVRow} row - The CSV row
-   * @returns {unknown} - The cell value
+   * @returns {any} - The cell value
    */
-  getCellValue(header: StaticHeaderType, row: CSVRow) {
+  getCellValue(header: StaticHeaderType, row: CSVRow): any {
```
NickPhura: Is this really `any` or can it be like `string | undefined`?

Possibly we might need to check if `row[header] / row[alias]` exists, and then cast it to string or else return undefined?

NickPhura: But, ignore if that doesn't help anything or makes things downstream worse

api/src/services/import-services/marking/marking-header-configs.ts
MacDeluca: Yea the pattern I have been using is just casting everything to lowercase. But to your point, it's kinda confusing for the caller to have to know to case everything to lowercase. I can make a change for this.

api/src/utils/csv-utils/csv-config-utils.ts
MacDeluca: Updated

api/src/services/import-services/marking/marking-header-configs.ts
MacDeluca: Updated

api/src/services/import-services/marking/import-markings-service.ts
MacDeluca: Updated

app/src/utils/file-utils.ts
MacDeluca: Created a csv-utils file

api/src/services/import-services/marking/marking-header-configs.ts
```diff
@@ -140,10 +140,6 @@ export const getMarkingBodyLocationCellValidator = (
   utils: CSVConfigUtils<MarkingCSVStaticHeader>
 ): CSVCellValidator => {
   return (params: CSVParams) => {
-    if (params.cell === undefined) {
-      return [];
-    }
```
NickPhura: Should this be based on the `optional` config flag?

NickPhura: (Read the doc, nvm!)

NickPhura: Yeah I guess the optional vs required cell value is a trickier check. 

For single columns (like taxon or alias) it would be easy enough to have a flag to make them optional. 

But for multi-column concepts, you would have to have some fancier check. Thinking of what it would look like to make "markings" as a concept optional, when it spans multiple columns, some of which are essential for a valid marking, and some of which are optional.

Anyways, something for later maybe

NickPhura (approved):
🦌 🦌

api/src/services/import-services/marking/marking-header-configs.ts
MacDeluca: Yea that's a good point. It reminds me of my brainstorms around row level validations vs cell validations (CSVCellValidator). Currently we only have the ability to validate the individual cells, but I could see some use in being able to create CSVRowValidators for extra validation control around groups or sets of values.

MacDeluca: You could have it run the row validators before the cell validators to validate structural cell data.
 
examples:
CSVRowValidator: alias + date + time -> maps to a critter capture
CSVCellValidator: alias === 'string' and is in the survey

PR #1452, opened 2024-12-11
SIMSBIOHUB-627: Fix Manage Telemetry Page

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-627

## Description of Changes

Fix the add/edit/delete manual telemetry functionality.
Add server-side pagination/sorting to manage telemetry table.
Fix device credentials upload endpoint (remove call to bctw)

#### Misc
Minor UI cleanup/tweaks.
Resolves some warnings about incorrect html.
Fix two incorrect useEffects for the device/deployment edit forms.

## Testing Notes

- I did not touch the import telemetry feature, so it is likely still broken. Will be updated to use the latest csv import code in a separate PR. Disabled the button in the meantime.


api/src/paths/telemetry/vendor/deployments.ts
```diff
@@ -5,7 +5,7 @@ import { getBctwUser } from '../../../services/bctw-service/bctw-service';
 import { BctwTelemetryService } from '../../../services/bctw-service/bctw-telemetry-service';
 import { getLogger } from '../../../utils/logger';
 
-const defaultLog = getLogger('paths/telemetry/manual');
+const defaultLog = getLogger('paths/telemetry/vendor/deployments');
```
mauberti-bc: Should this endpoint end with ```telemetry```? like vendor/deployments/telemetry? I would assume that fetching deployments would omit telemetry points and for the telemetry points to be a different request.

EDIT: It looks like this endpoint is making a request to BCTW and is never called in SIMS - can we remove it?

mauberti-bc:
Not sure if related to this PR but deployment do not display in the left side panel of the manage telemetry page, yet the request succeeds
image

api/src/paths/telemetry/vendor/deployments.ts
NickPhura: Yeah this endpoint is going to be dropped. This change isn't really needed, I was just updating some other instances of 'telemetry/manual' and this got caught in the crossfire.

NickPhura: Most of the related code will need to be updated or removed, etc, as part of the larger work needed to update the Manage Telemetry page.

NickPhura:
> Not sure if related to this PR but deployment do not display in the left side panel of the manage telemetry page, yet the request succeeds image:
[In the side bar](https://github.com/bcgov/biohubbc/blob/SIMSBIOHUB-627-cleanup/app/src/features/surveys/telemetry/list/SurveyDeploymentList.tsx#L395-L401)
[In the deployment colum in the table](https://github.com/bcgov/biohubbc/blob/SIMSBIOHUB-627-cleanup/app/src/features/surveys/telemetry/table/TelemetryTable.tsx#L58-L63)

If you add a new critter, then you can add them to a deployment, and they show up.

image

(Creating the manual telemetry record fails, because its calling that old endpoint)

mauberti-bc:
Looks great! Made me think of something made small enough to add to this PR: we need to let users somehow remove vendor telemetry from a survey, in case they imported the wrong device or just want it gone. Right now you cannot remove vendor telemetry records and therefore can't get around the foreign key constraint preventing deployments from being deleted.

MacDeluca (approved):
🥅

PR #1455, opened 2025-01-08
TechDebt: CSV Row Validation With Row State

MacDeluca:
## Description of Changes

- Adds row-level validation the the CSVConfig
- Adds row level state for additional validation metadata
- Slightly modifies existing Critter CSV import to use the CSVRowState

## Testing Notes

- Critter import should work as normal
- Check the Critter sex values are persisted (it now uses CSVRowState to store the uuid value for the sex_qualitative_option_id)

### DIFF
--- a/api/src/services/import-services/critter/import-critters-service.ts
+++ b/api/src/services/import-services/critter/import-critters-service.ts
@@ -172,10 +172,6 @@ export class ImportCrittersService extends DBService {
       // SIMS payload
       simsPayload.push(critterId);
 
-      if (!row[CSVRowState]?.sexId) {
-        throw new Error('no state');
-      }
-
       // Critterbase static headers payload
       critterbasePayload.critters?.push({
         critter_id: critterId,
--- a/api/src/services/import-services/critter/import-critters-service.ts
+++ b/api/src/services/import-services/critter/import-critters-service.ts
@@ -169,19 +169,13 @@ export class ImportCrittersService extends DBService {
     for (const row of rows) {
       const critterId = v4();
 
-      console.log({ rowState: row[CSVRowState] });
-
-      if (!row[CSVRowState]?.id) {
-        throw new Error('Row does not have row state');
-      }
-
       // SIMS payload
       simsPayload.push(critterId);
 
       // Critterbase static headers payload
       critterbasePayload.critters?.push({
         critter_id: critterId,
-        sex_qualitative_option_id: row[CSVRowState]?.id,
+        sex_qualitative_option_id: row[CSVRowState]?.sexId,
         itis_tsn: row.ITIS_TSN,
         animal_id: row.ALIAS,
         wlh_id: row.WLH_ID,
@@ -275,7 +269,6 @@ export class ImportCrittersService extends DBService {
 
     return {
       validateCell: getCritterSexCellValidator(rowDictionary, this.configUtils)
-      //setCellValue: getCritterSexCellSetter(rowDictionary, this.configUtils)
     };
   }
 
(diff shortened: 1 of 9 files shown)


api/src/utils/csv-utils/csv-config-validation.ts
```diff
@@ -30,18 +33,36 @@ export const validateCSVWorksheet = <StaticHeaderType extends Uppercase<string>>
...
-  forEachCSVCell(worksheet, config, (params, headerConfig) => {
-    // Validate the cell value and modify the errors
-    executeValidateCell(params, headerConfig, errors); // Mutates `errors`
+  // Iterate over each row in the worksheet and execute the row validators
+  forEachCSVRow(worksheet, config, (rowParams, rowValidators) => {
+    // Execute the row validators and modify the errors
+    rowValidators.forEach((rowValidator) => {
+      executeRowValidator(rowParams, rowValidator, errors);
+      // Update the row state for each row validator
+      updateRowState(rowParams, rows);
```
NickPhura: Since this is a `forEach` loop, is there any concern around the `updateRowState` calls conflicting with one another? Though, maybe its safe to require that the validation functions don't read/write to the same properties in the state.

MacDeluca: Latest version handles this better

api/src/services/import-services/critter/import-critters-service.ts
```diff
@@ -172,6 +172,10 @@ export class ImportCrittersService extends DBService {
       // SIMS payload
       simsPayload.push(critterId);
 
+      if (!row[CSVRowState]?.sexId) {
+        throw new Error('no state');
```
NickPhura: Can we use the `ApiGeneralError` instead (or similar), also would be good to include the explicit issue, at least in the `errors` param.

Maybe something along the lines of: 
```
throw new ApiGeneralError('Critter import failed', ['CSVRowState.sexId is undefined'])
```

(whatever error content is the most helpful)

api/src/utils/csv-utils/csv-config-validation.ts
```diff
@@ -126,105 +154,119 @@ export const validateCSVHeaders = (worksheet: WorkSheet, config: CSVConfig): CSV
...
-  }
+export const executeValidateCell = (params: CSVParams, validateCell: CSVCellValidator): CSVError[] => {
+  const cellErrors = validateCell(params);
+
+  return cellErrors.map((error) => ({
+    error: error.error,
+    solution: error.solution,
+    values: error.values ?? null,
+    cell: (error.cell === undefined ? params.cell : error.cell) ?? null, // Use cell value if intentionally null
+    header: (error.header === undefined ? params.header : error.header) ?? null, // Use header value if intentionally null
```
NickPhura, line 267: nit: id be interested in knowing more about this code, why the values might be intentionally null, etc.

NickPhura: Maybe add more detail to these comments, for future maintainers. Unless the reason would be obvious to someone that read through the code in more detail (rather than bouncing around like I am).

NickPhura:
Not specifically related to this PR: 

In the `CSVError` interface, it might be good to more explicitly define some of the fields. Mostly I was thinking of the `header` field, and was wondering if that was the user facing header, or the internal header. I would have assumed user facing header, since the error needs to make sense to the user. But wasn't sure if it could also be the internal header, and then is re-mapped before being sent to the frontend.

NickPhura (approved):

PR #1456, opened 2025-01-08
Update CSV Error Import Dialog Styling

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Replaces the data grid of errors with a stack of cards
- Updates the solution text for missing columns or unknown columns to include the column name


app/src/components/csv/CSVSingleImportDialog.tsx
```diff
@@ -90,24 +90,20 @@ export const CSVSingleImportDialog = (props: CSVSingleImportDialogProps) => {
       // Wait for the complete status to be rendered + 500ms before closing the dialog
       await waitForRenderCycle(500);
 
+      // Show a success snackbar message
+      dialogContext.setSnackbar({
```
MacDeluca, line 94: note: This was initially in the finally block to display a snackbar message when the CSV import failed but not because of CSVValidationErrors. Just as a little bit more feedback if something explodes in the backend.

app/src/components/csv/CSVErrors.tsx
```diff
@@ -0,0 +1,67 @@
...
+  errors: CSVError[];
+}
+
+/**
+ * Returns a stack of CSV errors with information about solutions
+ *
+ * @param {CSVErrorsProps} props
+ * @returns {*}
+ */
+export const CSVErrors = (props: CSVErrorsProps) => {
```
MacDeluca: nit: Maybe a more descriptive name like `CSVErrorsCardStack`, just something different than some of the language used previously for CSVErrors etc.

app/src/components/csv/CSVErrors.tsx
```diff
@@ -0,0 +1,67 @@
...
+      return {
+        id: v4(),
+        ...error
+      };
+    });
+  }, [props.errors]);
+
+  return (
+    <Stack gap={1}>
+      {rows.map((error) => {
```
MacDeluca: question: What happens when there are lots of validation errors ie: 100+

app/src/components/csv/CSVDropzoneSection.tsx
```diff
@@ -21,20 +20,22 @@ interface CSVDropzoneSectionProps {
  */
 export const CSVDropzoneSection = (props: PropsWithChildren<CSVDropzoneSectionProps>) => {
```
MacDeluca, line 21: note: This component was also being used in the Captures/Measurements/Markings page, maybe just verify that it looks as you expect still.
https://biohubbc-app-1456-af2668-dev.apps.silver.devops.gov.bc.ca/admin/projects/2/surveys/3/animals/captures

app/src/components/csv/CSVSingleImportDialog.tsx
mauberti-bc: The dialog doesn't close unless the import succeeds so the error message is displayed in the dialog instead of the snackbar. I just removed the failure snackbar because it seemed redundant

mauberti-bc: Ok, I'll undo this change

app/src/components/csv/CSVErrors.tsx
mauberti-bc: Good point, I think we can just display the first 20 or so (eg. "and 183 others" at the bottom). If there are more than 100 errors, it's probably the same error for every row?

mauberti-bc: I added just a simple UI pagination control to show 10 at a time. Tested with 35000 errors and there's no apparent issue

app/src/components/csv/CSVDropzoneSection.tsx
mauberti-bc: Still looks good

MacDeluca: Ok sweet 😺

app/src/components/csv/CSVErrorsCardStack.tsx
```diff
@@ -0,0 +1,101 @@
...
+                    <Typography variant="body2">{error.row ?? 'N/A'}</Typography>
+                  </Stack>
+                  <Stack>
+                    <Typography variant="body2" fontWeight={700}>
+                      Column
+                    </Typography>
+                    <Typography variant="body2">{error.header ?? 'N/A'}</Typography>
+                  </Stack>
+                  <Stack>
+                    <Typography variant="body2" fontWeight={700}>
```
MacDeluca, line 75: note: The previous version also handled "values", which are the available options the user can enter for the specific cell (if applicable). ie: ['red', 'blue', ...]
`CSVErrorsTableOptionsMenu`

mauberti-bc: image

Fixed

MacDeluca: suggestion: We should check what the "Allowed Values" looks like with lots of options. I'm thinking mostly about measurements or collection units where a taxon / critter potentially has lots of options to pick from (25+)

mauberti-bc: image

NickPhura (approved):

PR #1459, opened 2025-01-13
SIMSBIOHUB-659: Add Pagination to System Alerts on Admin Page

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-659

## Description of Changes

- Adds pagination for retrieving system alerts on the admin page


NickPhura:
nit: the alerts table allows you to sort on the preview column and the status column, but I don't think either of those are setup to sort. Might be worth adding `sortable: false` to the datagrid column definitions for those two columns.

mauberti-bc:
> nit: the alerts table allows you to sort on the preview column and the status column, but I don't think either of those are setup to sort. Might be worth adding `sortable: false` to the datagrid column definitions for those two columns.

Fixed

api/src/models/alert-view.ts
```diff
@@ -9,13 +9,16 @@ export const IAlert = z.object({
   severity: z.enum(['info', 'success', 'error', 'warning']),
   data: z.object({}).nullable(),
   record_end_date: z.string().nullable(),
-  status: z.enum(['active', 'expired'])
+  status: z.enum(['active', 'expired']),
+  create_date: z.string()
```
NickPhura: (not this line specifically)

Do you mind replacing the `IAlert` type here with an `AlertRecord`. See the existing files in `api/src/database-models`. 

The goal of the database-models file is to have a dedicated type for each table in the database (as needed). And then all of the other API types will be subsets or combinations of the base types.

The inspiration was that if every type utilizes a base type where possible, then when we update a database table, all we have to do is update the base type, and everywhere that gets touched or broken in the API should cascade from that.

api/src/models/alert-view.ts
```diff
@@ -9,13 +9,16 @@ export const IAlert = z.object({
...
 });
 
 // Infer types from the schema
 export type IAlert = z.infer<typeof IAlert>;
-export type IAlertCreateObject = Omit<IAlert, 'alert_id' | 'status'>;
-export type IAlertUpdateObject = Omit<IAlert, 'status'>;
+
+export type IAlertUpdateObject = Omit<IAlert, 'status' | 'create_date'>;
+
+export type IAlertCreateObject = Omit<IAlertUpdateObject, 'alert_id'>;
```
NickPhura: (Not right or wrong, but something to consider)

Using `Omit` does mean that if the base type changes, it will automatically change these types. Which might be desirable. But, might also be preferable to use `Pick` so that its more explicit which values this type is defining.

app/src/features/admin/alert/AlertContainer.tsx
```diff
@@ -34,41 +47,55 @@ const AlertListContainer = () => {
...
+  const paginationSort: ApiPaginationRequestOptions = useMemo(() => {
+    const sort = firstOrNull(sortModel);
+    return {
+      limit: paginationModel.pageSize,
+      sort: sort?.field || undefined,
+      order: sort?.sort || undefined,
+      page: paginationModel.page + 1 // API pagination pages begin at 1, but MUI DataGrid pagination begins at 0.
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [paginationModel]);
```
NickPhura: Just checking: Intentionally disabled sorting? 

It looks like the backend has support for it to some degree.

If intentionally disabling it, I would maybe mark the remaining columns as `sortable: false` and maybe just add a comment above the `// eslint-disable` line explaining that this should not run on sort model changes.

mauberti-bc: I was trying to fix a quirk where the useMemo gets triggered twice when the component mounts, causing the initial request to be made without any sorting. I've returned sortModel to the dependency array now and just added sorting to the repo function if the request doesn't have sorting params.

api/src/models/alert-view.ts
mauberti-bc: I removed IAlert and replaced it with AlertRecord 👍

api/src/database-models/alert.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @description Data record for `Alert`.
+ */
+export const AlertRecord = AlertModel.omit({
+  create_user: true,
+  update_date: true,
+  update_user: true,
+  revision_count: true
+}).extend({
+  status: z.enum(['active', 'expired'])
+});
```
NickPhura: Do you mind moving this 'extend status' piece into a separate type? Defined somewhere else (at the repo level, or in the alert-view.ts file maybe).

We want these base Model and Record types in the database-models folder to be 1-to-1 mappings to the real database table (with the Record excluding audit columns for convenience) so they can be universally used in all downstream situations.

NickPhura:
Looks great, just had 1 last comment above the database-model type.

api/src/database-models/alert.ts
mauberti-bc: I created a separate "AlertRecordWithStatus" type in alert-view.ts. There are a few other types like the filter objects for getting alerts that live in that same file.

NickPhura (approved):
🐐 🐑

PR #1461, opened 2025-01-14
TECHDEBT: validateZodCell params tweak

MacDeluca:
## Description of Changes

- Passing the cell value to `validateZodCell` vs the entire `params` object


NickPhura (approved):
⚡

PR #1467, opened 2025-01-21
SIMSBIOHUB-655: CSV Measurements

MacDeluca:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-655

## Description of Changes

- Refactors existing CSV measurement import to use new CSVConfig structure
- Adds reusable row validator for alias/capture date/capture time

## Testing Notes

- Should be able to import a measurement CSV with dynamic columns and see the measurements associated to the Capture


api/src/services/import-services/utils/measurement.ts
MacDeluca, line 1: question: Should I move these higher up?

NickPhura (dismissed):
👋

NickPhura (approved):
👋 👋

PR #1468, opened 2025-01-21
TechDebt: Replace Deprecated DeploymentConfig With Deployment.

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

`DeploymentConfig` is now deprecated in OpenShift. They should still function for now, but are no longer receiving updates, etc.

### Deployment changes
- Update all templates that referenced `DeploymentConfig` to use `Deployment`
  - Remove the `triggers` section, as this does not exist in `Deployment`
  - Update the strategy section to `RollingUpdate` which is the only one `Deployment` supports, with fewer configurable settings.
  - Fill in the `image` property, which was previously auto-populated by the `DeploymentConfig` based on the `triggers` section (I believe).
  - For `Deployment` specifically, under `selector` you have to include `matchLabels`

### Other changes
- I updated the yaml files for the backup and clamav containers. These are probably irrelevant as they are out of date anyways, and should probably be pulled from an external source, rather than be stored in our repo. But, that is techdebt for another PR.
- Updated the uuid package (I noticed a deprecation warning in the build logs)
- Re-enable running the migrations in the static deploy (was disabled as part of upgrading Postgres)

## Testing Notes

OpenShift pipeline works the same as it did before.

The deployments worked in this PR. I also tried making an app change and adding a new migration, just to ensure that they re-deployed correctly, and they did (I reverted those changes afterwards).


.github/workflows/cleanClosedPR.yml
```diff
@@ -81,5 +81,5 @@ jobs:
         env:
           POD_SELECTOR: biohubbc
         run: |
-          oc --namespace af2668-dev   get all,pvc,secret,pods,ReplicationController,DeploymentConfig,HorizontalPodAutoscaler,imagestreamtag -o name | grep $POD_SELECTOR | grep $PR_NUMBER | awk '{print "oc delete --ignore-not-found " $1}' | bash
-          oc --namespace af2668-tools get all,pvc,secret,pods,ReplicationController,DeploymentConfig,HorizontalPodAutoscaler,imagestreamtag -o name | grep $POD_SELECTOR | grep $PR_NUMBER | awk '{print "oc delete --ignore-not-found " $1}' | bash
+          oc --namespace af2668-dev   get all,pvc,secret,pods,ReplicationController,DeploymentConfig,Deployment,HorizontalPodAutoscaler,imagestreamtag -o name | grep $POD_SELECTOR | grep $PR_NUMBER | awk '{print "oc delete --ignore-not-found " $1}' | bash
```
MacDeluca, line 84: question: Are we keeping `DeploymentConfig` temporarily until all previous PR's are closed?

MacDeluca (approved):
🦖

.github/workflows/cleanClosedPR.yml
NickPhura: Yeah that was my thought, just in case. It should be harmless even if we have no deployment configs, at least until they fully drop it, which is probably not for a while.

PR #1471, opened 2025-01-29
SIMSBIOHUB-669: Show Column Options in Configure Observations Dialog General Tab

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-669

## Description of Changes

- Show code values for observation subcount sign in the configure columns dialog on the manage observations page
- Removes delete controls for measurements/environments outside from the general tab (you can delete in the other tabs of the configure dialog)


app/src/features/surveys/observations/observations-table/configure-columns/ConfigureColumnsButton.tsx
```diff
@@ -45,10 +55,75 @@ export const ConfigureColumnsButton = (props: IConfigureColumnsButtonProps) => {
...
-  const hideableColumns = useMemo(() => {
-    return columns.filter((column) => column?.hideable);
-  }, [columns]);
+  // Columns that can be hidden from the table (visibility toggled on/off)
+  const hideableColumns: IHideableColumn[] = useMemo(() => {
+    // Map columns to their IHideableColumn definitions to ensure no duplicates
+    const columnMap = new Map<string, IHideableColumn>();
+
+    // Add all standard, hideable columns
+    columns
```
MacDeluca: suggestion: Instead of looping through the columns 3 times, we should loop through a single time and set the `columnMap`

app/src/features/surveys/observations/observations-table/configure-columns/ConfigureColumnsButton.tsx
```diff
@@ -45,10 +55,75 @@ export const ConfigureColumnsButton = (props: IConfigureColumnsButtonProps) => {
...
+      .filter((column) =>
+        [
+          ...observationsTableContext.environmentColumns.quantitative_environments,
+          ...observationsTableContext.environmentColumns.qualitative_environments
+        ].some((environment) => environment.name === column.headerName)
+      )
+      .forEach((column) => {
+        const foundEnvironment = observationsTableContext.environmentColumns.qualitative_environments.find(
+          (environment) => column.headerName === environment.name
```
MacDeluca: suggestion: Should this be `column.headerName.toLowercase() === environment.name.toLowerCase()`

app/src/features/surveys/observations/observations-table/configure-columns/ConfigureColumnsButton.tsx
```diff
@@ -45,10 +55,75 @@ export const ConfigureColumnsButton = (props: IConfigureColumnsButtonProps) => {
...
+        if (foundEnvironment) {
+          columnMap.set(column.field, {
+            ...column,
+            options: foundEnvironment.options ?? []
+          } as IHideableColumn);
+        }
+      });
+
+    // Return the updated columns, using the map to ensure no duplicates
+    return Array.from(columnMap.values());
```
MacDeluca, line 101: suggestion: I think we could use a `Set` instead of a `Map` (columnMap) here, if we only care about the values.

app/src/features/surveys/observations/observations-table/configure-columns/ConfigureColumnsButton.tsx
mauberti-bc: Fixed

app/src/features/surveys/observations/observations-table/configure-columns/ConfigureColumnsButton.tsx
mauberti-bc: Yes, thanks

app/src/features/surveys/observations/observations-table/configure-columns/ConfigureColumnsButton.tsx
mauberti-bc: I used map because it overwrites existing values, which this requires (Nick's suggestion). 

The first operation adds all table columns to the map, including measurements. However, the qualitative measurement options aren't added in this first operation.

The next operations add the measurement columns but this time with the qualitative options. So, life stage, for example, will be in the map once and with its options listed.

app/src/features/surveys/observations/observations-table/configure-columns/ConfigureColumnsButton.tsx
```diff
@@ -45,10 +55,49 @@ export const ConfigureColumnsButton = (props: IConfigureColumnsButtonProps) => {
...
-  // The array of columns that may be toggled as hidden or visible
-  const hideableColumns = useMemo(() => {
-    return columns.filter((column) => column?.hideable);
-  }, [columns]);
+  // Columns that can be hidden from the table (visibility toggled on/off)
+  const hideableColumns: IHideableColumn[] = useMemo(() => {
+    const columnMap = new Map<string, IHideableColumn>();
+
+    columns.forEach((column) => {
+      if (!column?.hideable) return;
```
NickPhura: We should include the brackets in if statements like this. Helps prevent mistakes, and also prevents a few weird edge cases that can crop up when writing them without brackets.

```
if (!column?.hideable) {
  return;
}
```

NickPhura (approved):
🦖

PR #1474, opened 2025-01-30
SIMSBIOHUB-653: Observation CSV Import

MacDeluca:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-653

## Description of Changes

- Updates existing Observation CSV import to use the new structure (CSVConfig)
- Added taxon row validator - that accepts scientific name OR TSN
- Deprecated `process` endpoint and related S3 pre-upload/process code
- Refactored existing logic to handle CSVErrors
- Created new validation functions for environments / measurements
- Added row-state getters which validate the row state to ensure correct end state
- Updated measurement service to use new row-state getters
- Added `CaseInsensitiveMap` data strucuture class
- Updated Observation `upload` endpoint -> `import`
- Modified frontend useObservationApi hook slightly and included `surveySampleSiteId`

## Testing Notes

- Should be able to import this file with a matching `Technique` and period
[Bulk Observations.csv](https://github.com/user-attachments/files/18607610/Bulk.Observations.csv)

## Notes:
- I am chipping away slowly on test coverage


NickPhura:
When importing an observation file, some things:

If I set the `comment` cell to a number:
image

When I have no measurements or environment columns:
image

NickPhura:
Weird edge case where I've set the date/time to be the exact start date/time.

image

image

MacDeluca:
> When importing an observation file, some things:
> 
> If I set the `comment` cell to a number:

This should be resolved now, updated the description validator to coerce numbers into strings
> When I have no measurements or environment columns:

Updated the worksheet utils to handle this.

api/src/services/import-services/measurement/utils/measurement-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Get the measurement row TSN getter - ie: Get the ITIS TSN from the row
+ *
+ * @param {Map<string, ICritterDetailed>} surveyAliasMap The survey alias map
+ * @param {CSVConfigUtils} utils The CSV config utils
+ * @returns {*} {(params: CSVParams) => number} The get measurement row TSN callback
+ */
+export const getTsnFromMeasurementRow = (
+  surveyAliasMap: Map<string, ICritterDetailed>,
```
NickPhura, line 14: super nit: The name of this param might be more clear as something like "critterAliasMap" or "surveyCritterAliasMap"

Same for the calling function, which also names the variable "surveyAliasMap".

api/src/services/import-services/observation/utils/observation-sampling-row-validator.ts
```diff
@@ -0,0 +1,446 @@
...
+    if (!worksheetHasSamplingInformation && !worksheetObservationDate) {
+      return [
+        {
+          error: 'Row does not contain sampling information or observation date / time',
+          solution: 'Please provide sampling information or a valid observation date and time',
+          header: null,
+          cell: null
+        }
+      ];
+    }
```
NickPhura: Does this need to check for lat/long as well?

api/src/services/import-services/measurement/utils/measurement-utils.ts
MacDeluca: Yea that's totally fair. I just checked and there's about 90 references, I'll probably just follow this PR with one that find and replaces those references.

api/src/services/import-services/observation/utils/observation-sampling-row-validator.ts
MacDeluca: I actually handle this case in `observation-location-row-validator.ts`

api/src/services/import-services/measurement/utils/measurement-utils.ts
NickPhura: Yeah fair. Not actually a big deal, just caught my eye and I was thinking "surveys have aliases?"

api/src/services/import-services/utils/qualitative.ts
```diff
@@ -0,0 +1,67 @@
...
+      {
+        error: 'Qualitative value must be a string',
+        solution: 'Update the value to match the expected type'
+      }
+    ];
+  }
+
+  const matchingQualitativeOption = qualitativeTypeDefinition.options.find(
+    // Not sure why I need to cast params.value to string here after checking above...
+    (option) => option.option_name.toLowerCase() === String(value).toLowerCase()
```
NickPhura: My vscode seems happy if I remove the String cast. Seems to know it is a string.

NickPhura: The similar comment in `measurement-header-configs.ts` is because its a value from the params object. If you assigned it to a new variable, and then checked the type, it would be happy.

Since the object is passed by reference, the function cant be certain that the property hasn't changed between when it was type checked and when it was used in the find.

MacDeluca: Oh I just realized I forgot to drop that file `measurement-header-configs.ts` (it's deprected now and replaced with `measurement-dynamic-header-config.ts`)

api/src/services/import-services/utils/quantitative.ts
```diff
@@ -0,0 +1,66 @@
...
+    return [
+      {
+        error: 'Quantitative value must be a number',
+        solution: 'Update the value to match the expected type'
+      }
+    ];
+  }
+
+  // Validate value is withing the environment min max bounds
+  if (quantitativeTypeDefinition.max != null && value > quantitativeTypeDefinition.max) {
```
NickPhura, line 46: Should this be `>=` rather than `>`?

Same for the min?

api/src/services/import-services/utils/qualitative.ts
MacDeluca: Interesting, my editor also knows it's a string now, I must be going crazy

api/src/services/import-services/utils/row-state.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export const createRowStateGetter = <SchemaType extends z.ZodSchema>(schema: SchemaType) => {
+  return (rowOrState: CSVRow | CSVRow[typeof CSVRowState]): z.infer<SchemaType> => {
+    // Check if the state or row is being passed
+    // Note: The row state is stored as a symbol on the row object
+    const isRow = Object.getOwnPropertySymbols(rowOrState).includes(CSVRowState);
+
+    // Get the state from the row
+    const state = isRow ? rowOrState?.[CSVRowState] : rowOrState;
+
```
NickPhura: (optional) I think this type guard works (just to resolve the complaint about `rowOrState?.[CSVRowState]`).

```
const isRow = isCSVRow(rowOrState);
const state = isRow ? rowOrState[CSVRowState] : rowOrState;

export const isCSVRow = (rowOrState: CSVRow | CSVRow[typeof CSVRowState]): rowOrState is CSVRow => {
  return Object.getOwnPropertySymbols(rowOrState).includes(CSVRowState);
};
```

app/src/features/surveys/observations/sampling-sites/components/ImportObservationsButton.tsx
NickPhura, line 1: Should this file live higher up? Since it is used by the sampling component but also by the parent table component.

api/src/utils/xlsx-utils/worksheet-utils.ts
```diff
@@ -149,6 +149,11 @@ export const getWorksheetRowObjects = (worksheet: xlsx.WorkSheet): Record<symbol
     let rowHasValues = false;
 
     for (let j = 0; j <= originalRange.e.c; j++) {
+      // Skip empty headers
+      if (!headers[j]) {
+        continue;
+      }
+
       // Always add the header (key) to the row object
       rowObject[headers[j]] = undefined;
```
NickPhura: Just checking: 

Does this skip empty headers check contradict the comment below about always adding the header key?

(As to whether its best to omit empty headers or not, I can't say)

api/src/utils/xlsx-utils/cell-utils.ts
```diff
@@ -56,19 +49,22 @@ export function replaceCellDates(cell: CellObject): CellObject {
...
 
     return { ...cell, z: DefaultDateFormat, v: date ?? 'Invalid Date Format' };
   }
   // If time cell - convert the epoch value (ie: 0.5) to a time string (ie: '12:00:00')
   else if (isTimeCell(cell)) {
-    const time = dayjs.duration(Number(cell.v), 'days');
-
+    // Round the time fraction to the nearest millisecond to avoid floating point errors
+    // Excel stores time as a fraction of a day, where each day has 86400000 milliseconds.
+    // Ex: , 0.25 is 6:00:00 AM, 0.5 is 12:00:00 PM, 0.75 is 6:00:00 PM
```
NickPhura: 👀 Oh gawd that random comma 😄

api/src/services/import-services/utils/qualitative.ts
NickPhura: haha

api/src/services/import-services/observation/utils/observation-sampling-row-validator.ts
NickPhura: I wonder if the error messages from this validator and the lat/lng validator should be more aligned, just because they are doing very similar checks regarding defined vs undefined sampling.

`Latitude is required when sampling information is not provided`
`Please provide sampling information or a valid latitude`

`Longitude is required when sampling information is not provided`
`Please provide sampling information or a valid longitude`

`Date / time is required when sampling information is not provided`
`Please provide sampling information or a valid observation date / time`

api/src/services/import-services/observation/utils/observation-location-row-validator.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    if (!worksheetLongitude) {
+      errors.push({
+        error: 'Longitude is required when sampling information is not provided',
+        solution: 'Please include a longitude for the observation',
+        header: utils.getWorksheetHeader('LONGITUDE', params.row),
+        cell: null
+      });
+    }
```
NickPhura: We could technically have a third error message for when BOTH latitude and longitude are undefined.

Could be as combined message similar to the `date / time` message here: https://github.com/bcgov/biohubbc/pull/1474#discussion_r1941782257

`Latitude / longitude is required when sampling information is not provided` etc

api/src/utils/xlsx-utils/cell-utils.ts
MacDeluca: rm -rf *

api/src/services/import-services/utils/row-state.ts
MacDeluca: This is another one of those cases where my editor isn't picking up on these warnings, so my apologies. (I still need to investigate why this is happening)

Fixed.

api/src/utils/xlsx-utils/worksheet-utils.ts
MacDeluca: Yea they do contradict a bit, nice catch

Updated

api/src/services/import-services/observation/utils/observation-location-row-validator.ts
MacDeluca: I have them seperate so the `header` value is more clear about what it is referencing. It will return two errors when both are missing.

MacDeluca: I might be missunderstanding though

app/src/features/surveys/observations/sampling-sites/components/ImportObservationsButton.tsx
MacDeluca: Yea that would probably make sense, I think there was two version of this file and I probably deleted the wrong one (that was higher up) incorrectly.

MacDeluca: updated

api/src/services/import-services/utils/row-state.ts
MacDeluca: Just let me know on your end if it is still present

api/src/services/import-services/observation/utils/observation-sampling-row-validator.ts
MacDeluca: Yea I was uncertain if it was better to have this all combined or not. Mostly have them seperate as it was an afterthought tbh. I can combine them and drop the other validator.

api/src/services/import-services/observation/utils/observation-location-row-validator.ts
MacDeluca: Outdated. Dropped this and included this logic in the sampling site validator.

api/src/services/import-services/observation/utils/observation-sampling-row-validator.ts
NickPhura: I dont think separate validators is necessarily an issue. I mostly just mean having the error messages written in a similar way, whether its across multiple validators or not.

api/src/services/import-services/utils/row-state.ts
NickPhura: Seems happy on my end.

app/src/features/surveys/observations/sampling-sites/components/ImportObservationsButton.tsx
NickPhura: Nit: can we put it inside a `components` folder at its current level. Just so its not sitting alongside the `SurveyObservationPage.tsx` component, which is really the entrypoint for that whole observations feature.

Actually, the `SurveyObservationHeader.tsx` file could also be moved to that components folder, since its just a child of the page component.

api/src/services/import-services/observation/utils/observation-location-row-validator.ts
NickPhura: > I have them seperate so the header value is more clear about what it is referencing. It will return two errors when both are missing.

Right, was forgetting about the header field. Ignore me!

app/src/features/surveys/observations/sampling-sites/components/ImportObservationsButton.tsx
MacDeluca: Yea I can do that

MacDeluca: Fixed

NickPhura:
A few things:

If I add a measurement column to the csv that isnt supported by the taxon, the error is just the generic dynamic header error (rather than something more specific to the measurement not being valid or found or something).

If I add an empty column to my file, all of the columns after it get ignored.

api/src/services/import-services/measurement/utils/measurement-utils.ts
MacDeluca: Wrote a note and will resolve in the next PR

NickPhura (approved):
🔍

PR #1475, opened 2025-01-30
Show Animal Nickname in Capture/Mortality Map Popup

AMEIJER1:
## Links to Jira Tickets

* N/A

## Description of Changes

* When a capture/mortality is clicked on the map of animals on the survey page, the popup now shows the critter's nickname
* The table animal on the survey page now includes Sex as a column


app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
```diff
@@ -66,7 +68,7 @@ export const SurveySpatialAnimal = (props: ISurveySpatialAnimalProps) => {
           properties: {}
         }
       })) ?? [],
-    popup: (feature) => <SurveySpatialAnimalCapturePopup feature={feature} />,
+    popup: (feature) => <SurveySpatialAnimalCapturePopup captureId={`${feature.id}`} />,
```
MacDeluca: nit: `captureId={String(feature.id)}`

question: What is the type for `id`?

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalCapturePopup.tsx
```diff
@@ -1,37 +1,41 @@
...
+
+    const { capture_date, capture_time, capture_location } = captureDataLoader.data;
+    const { animal_id } = animalDataLoader.data;
 
-  const getCaptureMetadata = (capture: ICaptureResponse) => {
     return [
-      { label: 'Capture ID', value: String(capture.capture_id) },
-      { label: 'Date', value: dayjs(capture.capture_date).format(DATE_FORMAT.LongDateTimeFormat) },
-      { label: 'Time', value: String(capture.capture_time ?? '') },
+      { label: 'Nickname', value: animal_id },
```
MacDeluca, line 37: question: Is `Nickname` the new term for `Alias`?

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalCapturePopup.tsx
```diff
@@ -1,37 +1,41 @@
...
+      { label: 'Date', value: dayjs(capture_date).format(DATE_FORMAT.LongDateTimeFormat) },
+      { label: 'Time', value: String(capture_time ?? '') },
       {
         label: 'Coordinates',
-        value: [capture.release_location?.latitude ?? null, capture.release_location?.longitude ?? null]
+        value: [capture_location?.latitude, capture_location?.longitude]
```
NickPhura, line 42: With this change, If either `capture_location`, `latitude`, or `longitude` is undefined then this will blow up (`<undefined>.toFixed(6)` is not valid), because the `.filter` step is only checking that they are not null.

We either need to cast these to null like it did before OR add an undefined check to the filter.

```
.filter((coord): coord is number => coord !== null && coord !== undefined)
```

and actually we have a helper function for this scenario

```
.filter((coord): coord is number => isDefined(coord))
```

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalCapturePopup.tsx
```diff
@@ -45,15 +49,17 @@ export const SurveySpatialAnimalCapturePopup = (props: ISurveySpatialAnimalCaptu
...
         }
       }}>
       <SurveyMapPopup
-        isLoading={captureDataLoader.isLoading || !captureDataLoader.isReady}
-        title="Capture"
-        metadata={captureDataLoader.data ? getCaptureMetadata(captureDataLoader.data) : []}
-        key={`capture-feature-popup-${feature.id}`}
+        isLoading={captureDataLoader.isLoading || animalDataLoader.isLoading}
+        title="Capture Details"
+        metadata={formatPopupMetadata()}
```
NickPhura: (This was there before, and we probably have other places in the app that do this too, as its not always something we look for or even notice)

It is technically more performant to run `formatPopupMetadata()` in advance, and just pass the result to this prop. Additionally, wrapping it in a `useMemo` will ensure that `formatPopupMetadata` only gets called if something has changed.

The way it is now, the function gets called on every render, even if the response is destined to be the same.

```
const popupMetadata = useMemo(() => {
  ...
}, [captureDataLoader.data, animalDataLoader.data]);

<SurveyMapPopup
  metadata={popupMetadata}
  ... 
/>
```

NickPhura: Its quite possible that this component already doesn't re-render much and the performance is negligible, but its a good practice in general.

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalCapturePopup.tsx
```diff
@@ -45,15 +49,17 @@ export const SurveySpatialAnimalCapturePopup = (props: ISurveySpatialAnimalCaptu
       closeButton={true}
       autoPan={true}
       eventHandlers={{
-        add: () => {
-          captureDataLoader.load(String(feature.id));
+        add: async () => {
+          const capture = await captureDataLoader.load();
+          // Fetch critter information to get the animal's nickname to display, which isn't included in the capture response
+          if (capture) animalDataLoader.load(capture.critter_id);
```
NickPhura: We should include the brackets in if statements like this. Helps prevent mistakes, and also prevents a few weird edge cases that can crop up when writing them without brackets.

```
if (capture) {
  animalDataLoader.load(capture.critter_id);
}
```

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalTable.tsx
```diff
@@ -9,80 +12,81 @@ import { ScientificNameTypography } from 'features/surveys/animals/components/Sc
...
+      animals.map(({ critterbase_critter_id }) => critterbase_critter_id)
+    )
   );
 
-  // Load data if animals data is available
-  if (animals.length) {
-    animalsDataLoader.load();
-  }
+  useEffect(() => {
+    if (animals.length) animalsDataLoader.load();
```
NickPhura: Add brackets (https://github.com/bcgov/biohubbc/pull/1475/files#r1943575554)

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalTable.tsx
```diff
@@ -9,80 +12,81 @@ import { ScientificNameTypography } from 'features/surveys/animals/components/Sc
...
+  }, [animals]);
```
NickPhura: React is warning about a missing dependency.

There are a few ways to see the warnings. The easiest is to look at the `app` logs, which you can see if you run `make app-log`.

Additionally, if you have the `eslint` and `prettier` extensions installed in vscode, it will actually highlight them in the editor.

There are some cases where we intentionally omit dependencies from the array because it leads to some infinite loop. But generally we want to try and add all dependencies when possible as it helps prevent unexpected behaviours around when things should be rendering but arent, etc.

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalCapturePopup.tsx
mauberti-bc: Will update - thanks!

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalCapturePopup.tsx
mauberti-bc: Yeah, just a more intuitive frontend label. It's technically animal_id in the backend.

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
mauberti-bc: string | number. I changed it to String()

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimalTable.tsx
mauberti-bc: Forgot to run make fix to catch these. Thanks

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
```diff
@@ -90,7 +92,7 @@ export const SurveySpatialAnimal = (props: ISurveySpatialAnimalProps) => {
           properties: {}
         }
       })) ?? [],
-    popup: (feature) => <SurveySpatialAnimalMortalityPopup feature={feature} />,
+    popup: (feature) => <SurveySpatialAnimalMortalityPopup mortalityId={`${feature.id}`} />,
```
NickPhura:
```suggestion
    popup: (feature) => <SurveySpatialAnimalMortalityPopup mortalityId={String(feature.id)} />,
```

NickPhura: Just to match the one above

NickPhura (approved):
🦌

PR #1476, opened 2025-01-31
Remove Funding Source Amounts & Miscellanous UI tweaks

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Removes all references to "amount" in funding sources
- Drops "amount" from survey_funding_source database table
- Various CSS tweaks

## Testing Notes

- Should be able to create/edit a survey with and without funding sources
- NOTE: Cannot delete a Survey because of unrelated, outdated SQL


NickPhura (dismissed):
👍

NickPhura (approved):
💲

PR #1477, opened 2025-02-04
BugFix: Update API Logger

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Update logger to only create a singleton winston instance.
  - Previously it was creating multiple loggers, which is fine when writing to the console, but doesn't work very well when writing to a file, since we want all log messages to be written to the same log files.
  - Now, it creates a singleton instance, and to maintain compatibility with how the api files were calling it, it returns a custom wrapper that exposes the methods we currently use, and automatically adds the logger name to the logged message. 
 
Update file daily rotate logger transport settings slightly
- Fundamentally the same as before.

Reduce the max log file size slightly to ensure that we don't overload the PVC. 
- It may be possible for a log file to get bigger than the max size set. Currently, max log size is set to `10 x 45mb` while the PVC is set to `500mb`, which should ensure enough extra space that we dont overload the PVC.

Updates to `env.docker` logging vars. 
- Should be backwards compatible with what was there before.

## Testing Notes

Logging should work more-or-less the same as before.

Local development console logs should be the same.
Local development file logs (container -> /data/logs) should work as expected


api/src/utils/logger.ts
```diff
@@ -42,41 +33,69 @@ const getLoggerTransportTypes = (): string[] => {
...
- * LOG_LEVEL - Defines the level of logging that the logger will output to the console. (default: debug)
- *
- * LOG_LEVEL_FILE - Defines the level of logging that the logger will output to persistent log files. (default: debug)
+ * @param {string} logLabel common label for the instance of the logger.
+ * @returns
+ */
+export const getLogger = (logLabel: string) => {
+  const logger = _getLogger(DEFAULT_LOGGER);
+
+  return {
```
MacDeluca, line 63: super nit: We could return the logger with the properties ordered by severity.

ie:
error
warn
info
debug
silly

api/src/utils/logger.ts
```diff
@@ -42,41 +33,69 @@ const getLoggerTransportTypes = (): string[] => {
...
- *
- * ...etc
- *
- * Environment Variables:
- *
- * LOG_LEVEL - Defines the level of logging that the logger will output to the console. (default: debug)
- *
- * LOG_LEVEL_FILE - Defines the level of logging that the logger will output to persistent log files. (default: debug)
+ * @param {string} logLabel common label for the instance of the logger.
+ * @returns
```
MacDeluca: nit: Just missing the return type

api/src/utils/logger.ts
```diff
@@ -42,41 +33,69 @@ const getLoggerTransportTypes = (): string[] => {
...
+
+/**
+ * Get or create a singleton logger instance.
  *
- * @param {string} logLabel common label for the instance of the logger.
+ * @param {string} loggerName The name of the logger instance.
  * @returns
  */
-export const getLogger = function (logLabel: string) {
+export const _getLogger = function (loggerName: string) {
```
MacDeluca: nit: Maybe a different name would help differentiate the two getLogger functions more
ie: `_getOrCreateLoggerSingleton`

api/src/utils/logger.ts
```diff
@@ -130,7 +149,8 @@ export const getLogger = function (logLabel: string) {
     );
   }
 
-  return winston.loggers.get(logLabel || 'default', { transports: transports });
+  // Create new logger instance
+  return winston.loggers.add(DEFAULT_LOGGER, { transports: transports });
```
MacDeluca: question: Is this supposed to be `winston.loggers.add(loggername, ...)`?

NickPhura: Fixed

api/src/utils/logger.ts
NickPhura: Good suggestion, fixed

api/src/utils/logger.ts
NickPhura: Fixed

api/src/utils/logger.ts
NickPhura: Yeah true, fixed

mauberti-bc:
image

Small bug to fix in this PR: Manually adding observations lets you add invalid latitude/longitude values, which then causes the observations endpoint to fail for your Survey

NickPhura:
> Small bug to fix in this PR: Manually adding observations lets you add invalid latitude/longitude values, which then causes the observations endpoint to fail for your Survey

Fixed

MacDeluca (approved):
💯

PR #1479, opened 2025-02-06
SIMSBIOHUB 664:  Sample Periods CSV Import

MacDeluca:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-664

## Description of Changes

- New Sample Period CSV import
- New DualImportButton - For workflows that need to import both individually and bulk (Critter / Sample Periods)

## Testing Notes

- Should be able to import a sample period CSV (you can download the template via the dialog)
- Note: This will not check for duplicates OR overlapping period start / end dates (as Macgregor suggested in the effort of getting this into Prod quickly)


mauberti-bc:
A few changes to add:
* Validate that the end date is after the start date
* Add constraint preventing overlapping periods for a (site, technique) combo in the database

api/src/services/import-services/sampling-periods/import-sample-periods-service.ts
```diff
@@ -0,0 +1,144 @@
...
+   * @param {IDBConnection} connection - DB connection
+   * @param {string} surveyId
+   */
+  constructor(connection: IDBConnection, worksheet: WorkSheet, surveyId: number) {
+    super(connection);
+
+    const initialConfig: CSVConfig<SamplePeriodCSVStaticHeader> = {
+      staticHeadersConfig: {
+        SAMPLE_SITE: { aliases: ['SAMPLE SITE', 'SAMPLING_SITE', 'SAMPLING SITE', 'SITE', 'LOCATION', 'STATION'] },
+        TECHNIQUE_NAME: { aliases: ['TECHNIQUE NAME', 'METHOD_TECHNIQUE', 'METHOD TECHNIQUE', 'TECHNIQUE', 'METHOD'] },
```
NickPhura, line 55: nit: The name of this field is different from the observations field, and the array of aliases is also different.

I'm sure it doesn't change the functionality, but might be good to unify them if we can.

NickPhura: Same with 'SAMPLING_SITE' vs 'SAMPLING_SITE', etc.

(Possibly you mentioned doing that in another PR later, if so, nvm!)

api/src/services/import-services/sampling-periods/import-sample-periods-service.ts
```diff
@@ -0,0 +1,144 @@
...
+        start_date: row.START_DATE,
+        start_time: row.START_TIME,
+        end_date: row.END_DATE,
+        end_time: row.END_TIME
+      };
+    });
+
+    await samplePeriodService.insertSamplePeriods(this.surveyId, samplePeriods);
+
+    defaultLog.debug({ label: 'import sample periods', samplePeriods });
```
NickPhura: Re logging in general: We don't have a strict pattern established per se, but I think it might be worth including the name of the function in the log:

```
defaultLog.debug({ label: 'importCSVWorksheet', message: 'import sample periods', samplePeriods });
```

That combined with the label that gets set at the file level when you call getLogger helps narrow down where the log message came from more precisely.

NickPhura: I was going to make the label and message fields required as part of that previous logging ticket, but there are some logs in the app that dont have them, and I didnt want to make a giant PR

NickPhura:
Do we need to add a validation function for overlapping dates?

I tried a file with bad dates, and the SQL fails as expected, but it just results in a generic request error on the frontend, as opposed to a more human friendly validation error.

NickPhura:
I tried file with valid dates, and they get imported successfully. However, it looks like the dates are getting parsed incorrectly.

I had start date `01/08/2007` and end date `31/12/2007`
The end date worked as expected, but the start date was inserted as `January 8th`, when I was expecting `August 1st`.

When I changed excel to `yyyy-mm-dd` then it did work as expected.

The file I was testing with:
[Bulk.Periods.csv](https://github.com/user-attachments/files/18711527/Bulk.Periods.csv)

api/src/services/import-services/sampling-periods/import-sample-periods-service.ts
MacDeluca: Updated.

api/src/services/import-services/sampling-periods/import-sample-periods-service.ts
MacDeluca: Updated.

MacDeluca:
> I tried file with valid dates, and they get imported successfully. However, it looks like the dates are getting parsed incorrectly.
> 
> I had start date `01/08/2007` and end date `31/12/2007` The end date worked as expected, but the start date was inserted as `January 8th`, when I was expecting `August 1st`.
> 
> When I changed excel to `yyyy-mm-dd` then it did work as expected.
> 
> The file I was testing with: [Bulk.Periods.csv](https://github.com/user-attachments/files/18711527/Bulk.Periods.csv)

I think this was from not handling single digits (1/8/2024) for days / months. I updated the `formatDateString` function to handle those cases. I also updated the tests to be slightly more clear about which dates it needs to validate correctly.

MacDeluca:
> Do we need to add a validation function for overlapping dates?
> 
> I tried a file with bad dates, and the SQL fails as expected, but it just results in a generic request error on the frontend, as opposed to a more human friendly validation error.

Not yet, Macgregor said to prioritize getting this in first.

api/src/utils/csv-utils/row-validators/start-end-date-order-row-validator.ts
```diff
@@ -0,0 +1,64 @@
...
+    startTime: StaticHeaderType | undefined;
+    endDate: StaticHeaderType;
+    endTime: StaticHeaderType | undefined;
+  }
+): CSVRowValidator => {
+  return (params) => {
+    let startTimeStamp = utils.getCellValue(staticHeaders.startDate, params.row) as string;
+    let endTimeStamp = utils.getCellValue(staticHeaders.endDate, params.row) as string;
+
+    if (staticHeaders.startTime) {
```
NickPhura: It looks like `staticHeaders.startTime` is always truthy, and so the `startTimeStamp` is being created as `'2007-08-11 undefined'`. Same for the `endTimeStamp`

Should this `if` be checking the cell value instead?

NickPhura: In this case, I was testing with a file that had no cell values in either time column.

MacDeluca: Ooops brain dont work gud anymore. 

Fixed :)

NickPhura (dismissed):
🧪 ⌛

NickPhura (approved):
🦙

PR #1482, opened 2025-02-06
Show All Animals on Manage Animals Page

AMEIJER1:
## Description of Changes

- When an animal is not selected on the Manage Animals page, show the map/table of all Animals in the Survey


app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
```diff
@@ -25,28 +22,23 @@ interface ISurveySpatialAnimalProps {
...
-  // Data loader for fetching animal capture data for the map ONLY. Table data is fetched separately in `SurveySpatialAnimalTable.tsx`
   const geometryDataLoader = useDataLoader((critter_ids: string[]) =>
     crittersApi.critters.getMultipleCrittersGeometryByIds(critter_ids)
   );
 
   useEffect(() => {
-    if (!critterIds.length) {
-      return;
+    if (critterIds.length) {
+      geometryDataLoader.load(critterIds);
```
NickPhura: (No change needed, just some context around the way this was written before, with the negative length check and early return)

In general, It is considered best practice to keep code as flat as possible (with as little nesting as possible, when its reasonable to do so).

---

In this case, we have to do the `if` check for the length no matter what, but it is slightly nicer to not nest 'real' code when possible. Hence writing it like this:

```
if (what_we_dont_want) {
  return; // end early
}

if (also_what_we_dont_want) {
  return; // end early
}

return what_we_do_want();
```
---
In a similar vein:
```
if (some_condition) {
  return 1;
} else {
  return 2;
}
```
is better written as:
```
if (some_condition) {
  return 1;
}

return 2;
```

Since the `else` is implied, we can remove that extra level of nesting and reduce how many brackets you have to parse to understand the code.

---

In this case, the `useEffect` has so little code that no one is seriously going to worry about it. Just more of a heads up for future, especially if you have more than one `if/else` statement.

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
```diff
@@ -3,20 +3,17 @@ import { IStaticLayer } from 'components/map/components/StaticLayers';
...
 import { coloredCustomMortalityMarker } from 'utils/mapUtils';
+import { SurveySpatialAnimalTable } from './SurveySpatialAnimalTable';
 
-interface ISurveySpatialAnimalProps {
-  /**
-   * Array of additional static layers to be added to the map.
-   */
-  staticLayers: IStaticLayer[];
+interface ISurveySpatialAnimalMapProps {
+  staticLayers?: IStaticLayer[];
```
NickPhura, line 19: Can we add the JSDoc back?

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
```diff
@@ -25,28 +22,23 @@ interface ISurveySpatialAnimalProps {
...
   useEffect(() => {
-    if (!critterIds.length) {
-      return;
+    if (critterIds.length) {
+      geometryDataLoader.load(critterIds);
     }
-
-    geometryDataLoader.load(critterIds);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [critterIds]);
```
NickPhura: Missing dependency: `geometryDataLoader`

NickPhura: It is quite possible that including that dependency will result in an infinite loop of requests. In which case we need to include that comment `// eslint-disable-next-line react-hooks/exhaustive-deps` which tells the linter (eslint) to ignore the fact that there is a missing dependency.

NickPhura: In this case it may be fine to include it, as the `load` function is designed to only run once. So even if the useEffect is triggered more than once, the `load` function internally will only actually make the request one time ever.

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
NickPhura: More documentation is always a win

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
AMEIJER1: Hi! Thank you so much for taking the time to provide insightful, detailed feedback! I will review and revise :)

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
MacDeluca: resolved

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
MacDeluca: resolved

app/src/features/surveys/view/survey-spatial/components/animal/SurveySpatialAnimal.tsx
MacDeluca: resolved

MacDeluca (dismissed):
Looks good!

MacDeluca (approved):
🌡️

PR #1484, opened 2025-02-07
Include Species' Ecological Units on Standards Page

AMEIJER1:
## Description of Changes

- made changes to backend critterbase api service to allow for passing of ecological unit categories 
- updated standards page to have an ecological units tab under species, and to return the ecological categories (but not yet the category options)

## Testing Notes

- have not yet hooked up ecological unit options, Macgregor mentioned critterbase may need to be updated prior to continuing this ticket (critterbase updated to return all values at once?)


api/src/paths/standards/taxon/{tsn}/index.ts
```diff
@@ -141,6 +141,30 @@ GET.apiDoc = {
...
+                    },
+                    category_name: {
+                      type: 'string'
+                    },
+                    description: {
+                      type: 'string',
+                      nullable: true
+                    },
+                    itis_tsn: {
+                      type: 'integer'
```
MacDeluca: nit: minimum 1

app/src/features/standards/view/species/SpeciesStandardsResults.tsx
```diff
@@ -92,6 +94,31 @@ const SpeciesStandardsResults = (props: ISpeciesStandardsResultsProps) => {
...
         )}
+        {activeView === SpeciesStandardsViewEnum.ECOLOGICAL_UNIT && (
+          <>
+            {props.data.ecologicalUnits.map((category) => (
+              <AccordionStandardCard
+                key={category.collection_category_id}
+                label={category.category_name}
+                subtitle={category.description}
+                colour={grey[100]}>
+                {/* <Stack gap={2} my={2}>
```
MacDeluca: nit: We can drop this deprecated code if it is not needed anymore

api/src/paths/standards/taxon/{tsn}/index.ts
```diff
@@ -141,6 +141,30 @@ GET.apiDoc = {
                   }
                 }
               },
+              ecologicalUnits: {
```
MacDeluca, line 145: question: Is this intended to be the `Collection Categories` ('Population Unit'...) or is this supposed to be `Collection units` ('Telkwa'...)?

AMEIJER1: We're wanting it to pull collection categories (population unit for example), and then the accordion would expand to show the collection units (telkwa) - but I think currently there is no API request that returns both for a species so Macgregor mentioned there may need to be some updates to critterbase before we can finish this ticket. Currently the UI is only displaying a card for categories, with no collection unit options.

mauberti-bc (approved):
Looks good. I updated some tests and tweaked the frontend.

PR #1485, opened 2025-02-07
SIMSBIOHUB-668 Parse Telemetry Device Keys

oscar-ip:
## Links to Jira Tickets

- SIMSBIOHUB-668 Parse Telemetry Device Keys

## Description of Changes

- Vectronic Keyx/zip and Lotek Cfg/zip files are validated and inserted into the backend tables

## Testing Notes

- Tested with provided files and with fake files created with proper format.


api/src/utils/media/media-utils.ts
```diff
@@ -126,37 +184,153 @@ export const isZipMimetype = (mimetype: string): boolean => {
...
+};
+
+/**
+ * As A check test, fetch vectronic device telemetry separation count from the Vectronic API.
+ *
+ * @param { string } collarId - Vectronic collar device ID
+ * @param { string } collarKey - Vectronic collar key
+ * @returns {*} {Promise<number>}
+ */
+const fetchCollarSepCount = async (collarId: string, collarKey: string): Promise<number> => {
```
MacDeluca: suggestion: This function could be in the `TelemetryVectronicService` but It looks like we already have a method called `fetchTelemetryCountFromVectronic`, which might allow you to do the same type of validation check.

api/src/utils/media/media-utils.ts
```diff
@@ -126,37 +184,153 @@ export const isZipMimetype = (mimetype: string): boolean => {
...
+  });
+  try {
+    const response = await vectronicApi.get(`/collar/${collarId}/sep/count`, {
+      params: {
+        collarkey: collarKey
+      }
+    });
+    return response.data;
+  } catch (error: any) {
+    console.error(error.message);
```
MacDeluca: note: We don't use `console.logs/errors...` in the app, you can use the logger. Search for `getLogger` to see examples

api/src/utils/media/media-utils.ts
```diff
@@ -126,37 +184,153 @@ export const isZipMimetype = (mimetype: string): boolean => {
...
+    throw error;
+  }
+};
+
+const findVectronicExpectedTags = (str: string, substrings: ReadonlyArray<string>) => {
+  return substrings.filter((substring) => str.includes(substring));
+};
+
+// Map keyx data to common JSON
+const mapKeyxData = (input: any): IKeyxData => {
```
MacDeluca: nit: We like to add JsDoc to all functions, you can find other examples in the app of what the proper structure should look like.
nit: `input` should be fully typed to avoid using any

api/src/utils/media/media-utils.ts
```diff
@@ -169,84 +343,276 @@ export const isValidTelementryCredentialFile = (
...
+      });
+      break;
+    }
+
+    // Parse data out of the XML file
+    const parser = new XMLParser(xmlParserOptions);
+    const jsonKeyxData = mapKeyxData(parser.parse(keyxFileString));
+
+    // Validate key using Vectronic API separation count
+    const keyStatus = await fetchCollarSepCount(jsonKeyxData.id, jsonKeyxData.key)
```
MacDeluca: suggestion: I prefer using the `try/catch await` syntax vs then.catch when possible. In my oppinion makes the code more readable.

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
```diff
@@ -169,7 +167,12 @@ export function postSurveyTelemetryCredentialAttachment(): RequestHandler {
       });
     } catch (error) {
       defaultLog.error({ label: 'postSurveyTelemetryCredentialAttachment', message: 'error', error });
-      await connection.rollback();
+
+      // The rollback connection method triggers an exception if the connection is not open
+      // This exception error message is the one that bubbles up to the front end and not the intended error set when throwing HTTP400 exception
+      // Added to connection isConnectionOpen to rollback only when the connection has been opened
+      if (connection.isConnectionOpen()) await connection.rollback();
```
MacDeluca: nit: Prioritize using the full `if` syntax with brackets
ie: 
```
if (connection.isConnectionOpen()) {
await connection.rollback()
}
```

api/src/utils/media/media-utils.ts
```diff
@@ -126,37 +184,153 @@ export const isZipMimetype = (mimetype: string): boolean => {
...
+  // return null when all is good
+  return null;
+};
+
+const cfgToJSON = (input: string): ICfgData[] => {
+  const regex = /\[(\d+)\]\s*((?:Key=[^\n]+(?:\s+Iridium IMEI=\d+)?|\s+Iridium IMEI=\d+\s+Key=[^\n]+)+)/g;
+  return [...input.matchAll(regex)].map(([, id, block]) => {
+    // Extracting Key and Iridium IMEI directives in any order
+    const keyMatch = block.match(/Key=([^\s]+)/);
+    const imeiMatch = block.match(/Iridium IMEI=(\d+)/);
```
MacDeluca: question: Is `Iridium IMEI` in all lotek .cfg's? Or is that a specific model of Lotek devices?

mauberti-bc: I believe all devices have an IMEI in the .cfg file, and it represents the device's internationally unique ID (International Mobile Equipment Identity). It's specific to each device. I think it's reasonable to also confirm that the IMEI matches along with the key since it's always in the .cfg file.

api/src/repositories/attachment-repository.ts
```diff
@@ -1665,16 +1676,23 @@ export class AttachmentRepository extends BaseRepository {
    * @param {string} fileType
    * @param {number} surveyId
    * @param {string} key
-   * @return {*}  {Promise<{ survey_telemetry_credential_attachment_id: number }>}
+   * @return {*}  {Promise<IResponseTelemetryCredentialAttachment>}
    * @memberof AttachmentRepository
    */
   async insertSurveyTelemetryCredentialAttachment(
```
MacDeluca: suggestion: It looks like this method is probably handling a bit too much at once. I think we could probably create a some new classes called something like `TelemetryVendorCredentialService / TelemetryVendorCredentialRepository`. It could have some methods like `createLotekCredentials` and `createVectronicCredentials` as bulk actions.

api/src/utils/media/media-utils.ts
MacDeluca, line 1: suggestion: I think most of these utilities can be moved into a utils files under telemetry-services
ie: `telemetry-services/telemetry-credential-utils.ts`

api/src/utils/media/media-utils.ts
oscar-ip: hey Mac, I think I looked at that function and I think I decided to not use it because of the parameters it takes.
I agree, a better place for it to be is in the TelemetryVectronicService 👍 
I will move it there

api/src/utils/media/media-utils.ts
oscar-ip: good catch, I used it for my own debugging I will remove

api/src/utils/media/media-utils.ts
```diff
@@ -169,84 +343,276 @@ export const isValidTelementryCredentialFile = (
...
+};
+
+/**
+ * Returns IValidationData.
+ *
+ * @export
+ * @param {Express.Multer.File} file
+ * @return {IValidationData}
+ */
+export const checkFileForCfg = (file: Express.Multer.File): IValidationData => {
```
MacDeluca: nit / suggestion: This function might be easier to understand as two functions. That way you can have two distinct return types.
ie: `isFileALotekCredentialCfg(file)` and `convertLotekCredentialFileToJson(file)`

api/src/utils/media/media-utils.ts
MacDeluca: Hey! I think its actually accepting the same params with just some additional optional params that arent needed?

api/src/utils/media/media-utils.ts
MacDeluca: Ok sweet, that was the missing information I needed 🦖

api/src/utils/media/media-utils.ts
oscar-ip: I see, as you mention it is a preference I just like the then.catch as it is the modern way and I saw it is used in few other places in the api.

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
MacDeluca: praise: This is a great comment

api/src/utils/media/media-utils.ts
MacDeluca: suggestion: You can also create an interface file for your types which is a pattern we have been doing.
ie `telemetry-services/telemetry-credential-utils.interface.ts`

api/src/utils/media/media-utils.ts
```diff
@@ -71,21 +118,32 @@ export const parseUnknownS3File = async (rawMedia: GetObjectCommandOutput): Prom
...
+        .map((item) => {
+          const fileName = item?.name;
+          const mimetype = mime.getType(fileName) || '';
+          const buffer = item?.getData();
+
+          return new MediaFile(fileName, mimetype, buffer);
+        })
+    };
+  } catch (err: any) {
+    console.error(err);
```
MacDeluca: nit: console.error -> logger

api/src/utils/media/media-utils.ts
MacDeluca: I think the then.catch sytax is often used when nesting under an existing try / catch await

MacDeluca: Not a hard rule by any means, more just a style preference that we try to follow. Alternatively, which I think would be preferred is an additional method on the `TelemetryVectronicService` that uses the existing method `fetchTelemetryCountFromVectronic` to validate the key.

```js
async function validateVectronicDeviceCredentials (...): boolean {
try {
   await fetchTelemetryCountFromVectronic()
   return true
} catch (err) {
   return false
   }
}
```

api/src/repositories/attachment-repository.ts
NickPhura: Related: Ideally all of the regular logic is in the service and/or related util files, and the repository only handles building and running the SQL string.

So for example all of the for-loop logic would be in the service level.

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
oscar-ip: good catch, one would hope the linter would catch that :-)

api/src/utils/media/media-utils.ts
oscar-ip: I think yes,  this is what shows the config file I got
Iridium IMEI=300434068368170

api/src/utils/media/media-utils.ts
oscar-ip: already removed :-)

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
NickPhura: We had the same thought, as this came up recently in another PR, Its on the todo list to add that rule

MacDeluca: Agreed, I think we are gonna add a rule for it

api/src/utils/media/media-utils.ts
oscar-ip: I remember now that I looked again into the code I didn't put this call in the service because I didn't want to instantiate the service for a lightweight check that doesn't need to persist anything.

NickPhura: One consideration in this case, is that the service requires a database connection, which this util function probably doesn't have.

That aside, as a general rule of thumb, I wouldn't be too worried about the cost of instantiating a service when needed. They don't actually persist anything. They are really just collections of functions. 

May have been mentioned already, but I do think we should move these telemetry/collar/key related functions into a separate utils file. Just to keep each util file smaller and more specialized.

api/src/utils/media/media-utils.ts
NickPhura: Optional: I use a vscode extension called `Document This` which auto-generates the base JSDoc for a function, and then you just fill in any additional details/descriptions.

oscar-ip: I used the JSDoc Generate extension for vscode, I noticed it is being recently maintained.

api/src/utils/media/media-utils.ts
oscar-ip: Sounds good Mac that will certainly help to reduce the number of lines in the module 👍

api/src/repositories/attachment-repository.ts
oscar-ip: Makes sense Nick, I will get that changed 👍

api/src/utils/media/media-utils.ts
NickPhura: Oh nice, Ill take a look

api/src/utils/media/media-utils.ts
oscar-ip: Moved to TelemetryVectronic service.

api/src/utils/media/media-utils.ts
oscar-ip: This is done

api/src/utils/media/media-utils.ts
oscar-ip: Done

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
oscar-ip: Done

api/src/repositories/attachment-repository.ts
oscar-ip: I refactored the code now, it will need another review

api/src/utils/media/media-utils.ts
oscar-ip: I think this is already addressed in the latest (3rd) commit

api/src/utils/media/media-utils.ts
oscar-ip: done

database/src/migrations/20250207000001_update_vectronic_lotek_credential_tables.ts
```diff
@@ -0,0 +1,25 @@
+import { Knex } from 'knex';
+
+/**
+ * Drops the `amount` column from survey_funding_source.
+ *
+ * Rationale: SIMS does not intend to track dollar amounts because they will be inaccurate. Dollar amounts are known by funding providers.
```
NickPhura: nit: update doc

api/src/services/attachment-service.ts
```diff
@@ -910,43 +952,75 @@ export class AttachmentService extends DBService {
...
+      survey_telemetry_vendor_credential_id: [],
+      telemetry_credential_lotek_id: [],
+      telemetry_credential_vectronic_id: []
+    };
+
+    responseJSON.survey_telemetry_credential_attachment_id =
+      await this.attachmentRepository.insertSurveyTelemetryCredentialAttachment(deviceKeyData);
+
+    const vendor = TELEMETRY_CREDENTIAL_ATTACHMENT_TYPE.CFG === deviceKeyData.fileData.type ? 'Lotek' : 'Vectronic';
+    if (deviceKeyData.fileData.keyData) {
```
NickPhura: Suggestion: I would reverse this `if` and return early if it fails.
```
if (!deviceKeyData.fileData.keyData) {
  return responseJSON;
}

for (...) {}
```

That way the `for` loop doesn't need to be nested.

api/src/services/attachment-service.ts
```diff
@@ -910,43 +952,75 @@ export class AttachmentService extends DBService {
...
+      for (const keyFile of deviceKeyData.fileData.keyData) {
+        // Iterate through the keys array
+        if (keyFile.keysData) {
```
NickPhura: Same as the above:

```
if (!keyFile.keysData?.length) {
  continue;
}

for (...) {}
```

database/src/migrations/20250207000001_update_vectronic_lotek_credential_tables.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @returns {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    
+    SET SEARCH_PATH=biohub;
+
+    ALTER TABLE telemetry_credential_vectronic ALTER COLUMN idcom SET DATA TYPE NUMERIC;
+    ALTER TABLE telemetry_credential_lotek ADD COLUMN devicekey VARCHAR(1000) NOT NULL;
+    COMMENT ON COLUMN telemetry_credential_lotek.devicekey IS 'The Lotek device key.';
```
MacDeluca: suggestion: Maybe make this a bit more clear about the difference between this value (`devicekey`) and `device_key`

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
```diff
@@ -169,7 +172,14 @@ export function postSurveyTelemetryCredentialAttachment(): RequestHandler {
...
     } catch (error) {
       defaultLog.error({ label: 'postSurveyTelemetryCredentialAttachment', message: 'error', error });
-      await connection.rollback();
+
+      // The rollback connection method triggers an exception if the connection is not open
+      // This exception error message is the one that bubbles up to the front end and not the intended error set when throwing HTTP400 exception
+      // Added to connection isConnectionOpen to rollback only when the connection has been opened
+      if (connection.isConnectionOpen()) {
+        await connection.rollback();
+      }
```
NickPhura, line 181: 👍 We could probably just start incorporating this `if is open` check around all of our `rollback` calls, as we touch other endpoint files in other PRs.

(For context, the way things evolved, the try/catch was originally added just to catch connection related errors. So in most cases the very first line of the `try` block was the `open` call. And so if the catch ever triggered, it was more-or-less guaranteed that the connection would be open already, and then it would be safe to call `rollback`. But, what you have here is definitely safer, and allows the `try` to wrap more of the code above the `open` call.)

api/src/repositories/attachment-repository.ts
```diff
@@ -1657,24 +1659,7 @@ export class AttachmentRepository extends BaseRepository {
...
-   * @memberof AttachmentRepository
-   */
-  async insertSurveyTelemetryCredentialAttachment(
-    fileName: string,
-    fileSize: number,
-    fileType: string,
-    surveyId: number,
-    key: string
-  ): Promise<{ survey_telemetry_credential_attachment_id: number }> {
+  async insertSurveyTelemetryCredentialAttachment(deviceKeyData: IDeviceKeyData): Promise<number> {
```
NickPhura, line 1669: nit: add jsdoc

api/src/repositories/telemetry-repositories/telemetry-lotek-repository.interface.ts
```diff
@@ -27,3 +27,16 @@ export const LotekPayload = TelemetryLotekRecord.omit({
...
+ * Lotek device key data structure
+ *
+ * @export
+ * @interface ICfgData
+ * @typedef {ICfgData}
+ */
+export interface ICfgData {
+  id: number;
+  key: string;
+  'Iridium IMEI': number;
```
NickPhura, line 41: Did we confirm if `Iridium` is always used? or if it could be another name?

api/src/services/telemetry-services/telemetry-utils.ts
```diff
@@ -41,3 +46,186 @@ export const keysToLowerCase = <T>(obj: Record<string, any>): T => {
...
+};
+
+/**
+ * As A check test, fetch vectronic device telemetry separation count from the Vectronic API.
+ *
+ * @param { string } collarId - Vectronic collar device ID
+ * @param { string } collarKey - Vectronic collar key
+ * @returns {*} {Promise<number>}
+ */
+export const fetchCollarSepCount = async (collarId: string, collarKey: string): Promise<number> => {
```
MacDeluca: nit: Duplicate of this function exists in `telemetry-vectronic-service`, this instance can be dropped.

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
MacDeluca: And alternatively we could incorporate the `isConnectionOpen()` check within the `connection.rollback()` so the connection can be "closed" safely regardless if it was opened.

api/src/repositories/telemetry-repositories/telemetry-vectronic-repository.ts
```diff
@@ -69,4 +70,50 @@ export class TelemetryVectronicRepository extends BaseRepository {
...
   }
+
+  /**
+   * Insert Vectronic device key data
+   *
+   * @async
+   * @param {IKeyxData} key
+   * @returns {Promise<number>}
+   */
+  async insertTelemetryCredentialAttachmentVectronic(key: IKeyxData): Promise<number> {
```
NickPhura: nit: We could remove `attachment` from the function name, since this function doesn't actually know about any file data.

Same for the similar `Lotek` function

api/src/services/telemetry-services/telemetry-utils.ts
NickPhura: It doesn't look like this function is ever called. Can it be deleted?

NickPhura:
Gave the code a read over, looks great overall.
Tried running it locally, and throwing some valid/invalid files through it, works great. The error messages in the UI are nice.


It looks like there are telemetry file-parsing related functions in both `src/media-utils.ts` and `src/services/telemetry-services/telemetry-utils.ts`. Can they be combined? It might be ideal to move all of the telemetry related code from `media-utils`, into `telemetry-utils`, just so its all in one place.

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
NickPhura: +1

api/src/services/telemetry-services/telemetry-utils.ts
oscar-ip: Yes, I noticed I missed that :-)

oscar-ip: Yes, I just missed removing it 
gone now

api/src/repositories/telemetry-repositories/telemetry-vectronic-repository.ts
oscar-ip: Done, in the next commit

api/src/repositories/telemetry-repositories/telemetry-lotek-repository.interface.ts
oscar-ip: I think Mac, asked the same question 
this is @mauberti-bc  response https://github.com/bcgov/biohubbc/pull/1485#discussion_r1947048439

api/src/repositories/attachment-repository.ts
oscar-ip: done in next commit

api/src/paths/project/{projectId}/survey/{surveyId}/attachments/telemetry/index.ts
oscar-ip: In my opinion, calling the function only when is needed  should work fine and as Nick mentions in most cases that logic is not required as the connection is already open.

database/src/migrations/20250207000001_update_vectronic_lotek_credential_tables.ts
oscar-ip: changed the comment to
'The Lotek device key that corresponds to the Key label in the CFG file.'

api/src/services/attachment-service.ts
oscar-ip: In this case, I had to use positive logic to keep typescript happy
switching the if leads to this typescript complain in the for loop 'deviceKeyData.fileData.keyData' is possibly 'undefined'. 

Hopefully leaving it as is, is ok in this case

database/src/migrations/20250207000001_update_vectronic_lotek_credential_tables.ts
oscar-ip: done, next commit

api/src/repositories/telemetry-repositories/telemetry-lotek-repository.interface.ts
NickPhura: Sorry to clarify: the "IMEI" portion is probably in all files, based on macgregors comment. But is the word "Iridium" always the same?

Would you ever have "OtherCompany IMEI"

I think Iridium is the name of the satellite company, so would be ever have collars that communicate with other satellite companies?

If it could change, then we'd need to parse that company name using a regex or something, rather than hard-coding it. If it doesn't change, then no worries.

database/src/migrations/20250207000001_update_vectronic_lotek_credential_tables.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @param {Knex} knex
+ * @returns {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    
+    SET SEARCH_PATH=biohub;
+
+    ALTER TABLE telemetry_credential_vectronic ALTER COLUMN idcom SET DATA TYPE NUMERIC;
```
MacDeluca: After looking into the BCTW data more I realized that this type is actually VARCHAR(50).

Some of example values are: `0-779834` and `300434067057670`

api/src/utils/media/media-utils.ts
oscar-ip: Since the same logic was used in 3 places, I changed it to put all the key file extensions check in a single function isCredentialFileExtension

oscar-ip: it will be in next merge, closing this conversation but if you have any further comments just let me know

api/src/repositories/telemetry-repositories/telemetry-lotek-repository.interface.ts
oscar-ip: I see, I guess that is a question for @mauberti-bc

oscar-ip:
> Gave the code a read over, looks great overall. Tried running it locally, and throwing some valid/invalid files through it, works great. The error messages in the UI are nice.
> 
> It looks like there are telemetry file-parsing related functions in both `src/media-utils.ts` and `src/services/telemetry-services/telemetry-utils.ts`. Can they be combined? It might be ideal to move all of the telemetry related code from `media-utils`, into `telemetry-utils`, just so its all in one place.

Great 👍 
on the next merge I put all the telemetry stuff in the telemetry-utils and created its corresponding unit test file.

database/src/migrations/20250207000001_update_vectronic_lotek_credential_tables.ts
oscar-ip: np, I will change that and make another commit

oscar-ip: I made the change now in the last commit

api/src/services/attachment-service.ts
oscar-ip: changes made

api/src/services/telemetry-services/telemetry-utils.ts
```diff
@@ -41,3 +81,515 @@ export const keysToLowerCase = <T>(obj: Record<string, any>): T => {
...
+    // Validate keys and produce keys JSON
+    const cfgDataJSON = processCfgFilesArray(zipEntries.filesArray);
+    if (cfgDataJSON[0].error) {
+      return {
+        type: TELEMETRY_CREDENTIAL_ATTACHMENT_TYPE.CFG,
+        error:
+          TELEMETRY_CREDENTIAL_ATTACHMENT_ERROR_STRING.INVALID_KEY_FILE +
+          cfgDataJSON[0].fileName +
+          ', ' +
+          cfgDataJSON[0].error
```
NickPhura: Recommendation: use template strings, rather than concatenating strings manually:

Also, some of the enum error strings have a space at the end, presumably because it makes concatenating strings easier. If you move to template strings, then you don't need the extra space in the enum, because its super easy to add a space between variables in the template string. You could also drop the colons in the enum too, and just add them directly in the template string when needed. That way the enum could be more like a collection of general error messages, and all of the extra grammar, etc, is added as needed when building the error response.

```typescript
error: `${TELEMETRY_CREDENTIAL_ATTACHMENT_ERROR_STRING.INVALID_KEY_FILE}: ${cfgDataJSON[0].fileName}, ${cfgDataJSON[0].error}`
```

NickPhura:
Looks good! I tried it locally, and ran some keys through it. Everything seems to be working great. 

I just had the one last comment about the template strings.

api/src/services/telemetry-services/telemetry-utils.ts
oscar-ip: Makes sense, I will adjust that in all the pertinent places

NickPhura (approved):
🧭

PR #1495, opened 2025-02-11
BugFix: Fix Sampling Information Cache

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Fixes the sampling information cache.

The sampling site, technique, and period columns in the observations data grid should now load records as expected, based on the correct site -> technique -> period record relationships.

Typing in the sites drop-down should filter the results as expected (fixed in previous PR).
- Note: The technique search always returns all results regardless of what you enter. The findTechnique backend service/repo receives the keyword, it just doesn't do anything with it currently.
- Note: The period search always returns all results regardless of what you enter. The keyword isn't even sent to the backend currently.

## Dev Notes

The cache works by tracking 3 objects (for site, technique, period). Where the key is the primary ID, and the value is the respective autocomplete option.

There are 2 new index objects, which tracks a key against an array of ids.
- site id -> technique id
- site id + technique id -> period id

This way, the technique edit component can filter its drop-down options based on the site value provided (from the site datagrid value).
And the period edit component can filter its drop-down options based on both the site and technique values provided (based on their respective values in the datagrid state).

## Testing Notes

Site column should search and filter as expected from an autocomplete field.

Technique column should show all valid techniques for the chosen site, or nothing if no site value has been selected.
If the site value changes, the technique value will be cleared.

Period column should show all valid periods for the chosen site + technique, or nothing if no site or technique values have been selected.
If the site or technique value changes, the period value will be cleared.

### DIFF
--- a/app/src/features/surveys/observations/observations-table/grid-column-definitions/count/ObservationCountDataGridEditCell.tsx
+++ b/app/src/features/surveys/observations/observations-table/grid-column-definitions/count/ObservationCountDataGridEditCell.tsx
@@ -28,7 +28,7 @@ export const ObservationCountDataGridEditCell = <DataGridType extends GridValidR
   const codesContext = useCodesContext();
 
   const getResponseMetric = () => {
-    const currentTechnique = samplingInformationCache.getCurrentTechnique(dataGridProps);
+    const currentTechnique = samplingInformationCache.getCurrentTechnique(dataGridProps.value);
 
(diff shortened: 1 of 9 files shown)


mauberti-bc (dismissed):
👍

MacDeluca (dismissed):
Skimmed the code and tested the UI LGTM

MacDeluca (approved):
👍🏼

PR #1496, opened 2025-02-12
BUGFIX: Observation Import Period Matching

MacDeluca:
## Description of Changes

- Allows periods to be matched without time signatures ie: 2025-01-01 - 2025-01-01 === 2025-01-01 10:10:00 - 2025-01-01 10:10:00
- Refactored existing `observation-sampling-row-validator` into smaller functions
- Fixed bug that prevented sample periods from being updated (overlapping start / end dates on self)

## Testing Notes

- Should be able to import an Observation CSV to an existing sample period (with a start / end time) when you do not include time signatures in the period


NickPhura:
Ran a bunch of files through this locally, everything seems to work as I believe its supposed to.

I did notice one scenario, though I am not sure if its actually doing what its supposed to or not:

I have 2 sample periods (with the same site/technique). They have the exact same date, but different time ranges.
- 2025-02-20 12:00:00 - 2025-02-20 12:00:00
- 2025-02-20 12:00:01 - 2025-02-20 14:00:00

In my CSV I have 2 rows.
- If I specify the full period with dates and times, then the data is inserted without issue.
- If I only specify the dates (and exclude the times) then the data throws an error for one of the rows (row 3)

image

MacDeluca:
This is a great example you provided and hopefully this explains (to my best understanding) what is happening internally.

> I have 2 sample periods (with the same site/technique). They have the exact same date, but different time ranges.
> 
> * 2025-02-20 12:00:00 - 2025-02-20 12:00:00
> * 2025-02-20 12:00:01 - 2025-02-20 14:00:00
> 

When we include a time signature on either the start or end date of a CSV period we will attempt to exact match an existing period (using the time).

When it is not included we do a "broader" match, finding any matching sample periods with the same date ignoring the time.

The logic for your example would look something like: 

1. Matches the two sample periods for "TechTwo" using the sample period start / end date ranges from the CSV.
2. Passes those two matches to the Observation date / time filter (to find an exact match)
3. The CSV observation date + time both fall WITHIN the matching sample periods start/end date (including time), therefore we return an error indicating we were unable to find an exact match (ambiguous).

Note: The final check is not aware of any exact matches, it only cares if the date and time fall within the subset of fitered sample periods. If it falls in the range of more than one (like in this case) it will return errors.

NickPhura:
>     3. The CSV observation date + time both fall WITHIN the matching sample periods start/end date (including time), therefore we return an error indicating we were unable to find an exact match (ambiguous).

I think this is where I was getting confused, because I don't think the times do fall in the same ranges.

The one period is from 12:00:00 - 12:00:00, and the observation time is 12:00:00. So those presumably line up as we expect.
But the other period is from 12:00:01 - 14:00:00 (1 second after 12:00:00), and the observation time is 13:00:00. So I assumed that should only match the second period, and not the first.

NickPhura (dismissed):
🕥

NickPhura (approved):
🕦

PR #1497, opened 2025-02-12
Show Descriptions in Autocomplete Options

LouisThedroux:
## Links to Jira Tickets
NA. 

ADDING DESCRIPTIONS TO VARIOUS DROPDOWNS IN THE APP.

New Migration created for adding description values, where missing.
Added descriptions for Project roles, Survey roles, Purpose and Methodology form (Collected data & Ecological concepts of interest) and Site Selection Strategy. 

Note, I had to change some tests as they were failing.


api/src/paths/codes.ts
```diff
@@ -122,6 +122,9 @@ GET.apiDoc = {
                     },
                     name: {
                       type: 'string'
+                    },
+                    description: {
+                      type: 'string'
```
mauberti-bc, line 127: If the description can be null in the database, you should include `nullable: true` in the description object here. Otherwise the request would fail for what the database considers valid data. 

(Disregard if the description field of the type database table has a not null constraint.)

app/src/features/surveys/edit/EditSurveyForm.tsx
```diff
@@ -80,6 +81,14 @@ const EditSurveyForm = <
     .concat(SurveyPartnershipsFormYupSchema)
     .concat(SpeciesYupSchema);
 
+  // Add this transformer function
+  const transformSurveyJobs = (jobs: typeof codes.survey_jobs): IGetAllCodeSetsResponse['project_roles'] => {
```
mauberti-bc: I don't think this is doing anything?

app/src/hooks/api/useUserApi.test.ts
```diff
@@ -45,31 +45,45 @@ describe('useUserApi', () => {
...
-      {
-        system_user_id: 1,
-        user_identifier: 'myidirboss',
-        role_names: ['role 1', 'role 2']
-      },
-      {
-        system_user_id: 2,
-        user_identifier: 'myidirbossagain',
-        role_names: ['role 1', 'role 4']
+    // ...existing test setup...
```
mauberti-bc: 👀

app/tsconfig.json
```diff
@@ -12,7 +12,6 @@
     "forceConsistentCasingInFileNames": true,
     "noImplicitThis": true,
     "noImplicitAny": true,
-    "suppressImplicitAnyIndexErrors": true,
```
mauberti-bc: This controls some typescript settings that we want to keep, so we shouldn't be modifying this file.

database/src/migrations/20250128000001_update_type_descriptions.ts
```diff
@@ -0,0 +1,46 @@
+import { Knex } from 'knex';
+
+/**
+ * Add new columns to survey_sample_period table
```
mauberti-bc: Update the jsdoc

app/src/features/projects/components/ProjectUserForm.tsx
```diff
@@ -30,14 +30,15 @@ export const ProjectUserRoleYupSchema = yup.object().shape({
...
+  roles: ICodeWithDescription[];
+  description?: string;
 }
 
 export const ProjectUserRoleFormInitialValues = {
   participants: []
 };
 
-const ProjectUserForm = (props: IProjectUserFormProps) => {
+const ProjectUserForm: React.FC<IProjectUserFormProps> = (props) => {
```
NickPhura: (More-or-less functionally the same, just an older pattern thats being deprecated)

Instead of:
```
// 1. old pattern
const ProjectUserForm: React.FC<IProjectUserFormProps> = (props) => {
```
Its better to say:
```
// 2. new pattern, without children
const ProjectUserForm = (props: IProjectUserFormProps) => {

// 3. new pattern, with children
const ProjectUserForm = (props: PropsWithChildren<IProjectUserFormProps>) => {
```

`React.FC<myType>` is an old way of defining a react component (which includes `children`). Either example 2 or 3 is the preferred way. Also, in the vast majority of components we don't leverage `children`, so often example 2 is what we want.

MacDeluca: I resolved this, and a few others I found while fixing something else.

MacDeluca (approved):
Looks good. Normalized some of the props names ie: subText -> description

PR #1502, opened 2025-02-20
SIMSBIOHUB-668a cross-referencing device keys with telemetry data

oscar-ip:
## Links to Jira Tickets

- SIMSBIOHUB-668 

## Description of Changes

- cross-referencing device keys with telemetry data
- added filter for get and find methods to show telemetry data for loaded keys

## Testing Notes

- Limited local  testing with an admin user more testing with non admin user needed


api/src/repositories/telemetry-repositories/telemetry-vendor-repository.ts
```diff
@@ -109,10 +124,44 @@ export class TelemetryVendorRepository extends BaseRepository {
...
 
+  /**
+   * Filter results by vendor credentials
+   *
+   * @private
+   * @static
+   * @param {Knex.QueryBuilder} queryBuilder
+   * @returns {Knex.QueryBuilder}
+   */
+  private static getLotekOrVectronicTelemetryByCredentialsClause(queryBuilder: Knex.QueryBuilder): Knex.QueryBuilder {
```
NickPhura: nit: do you mind if we remove `private`. 

None of the code in this app is going to be used by anyone else so we don't really benefit from hiding the functions. It can also make unit testing more annoying as you cant directly reference or mock private functions (even though in this case we don't need to test these functions, since they don't actually have any proper logic or decision making).

MacDeluca: It would also be good to include some information about why this method is `static` and the issue with using `this` etc.

NickPhura: > It would also be good to include some information about why this method is `static` and the issue with using `this` etc.

Yeah good point, help someone in future avoid the same issue.

oscar-ip: Sounds good, I think I left the private in there so this methods cannot be called from from anywhere else in the code but since there is no benefit and make unit testing more difficult I will remove.

oscar-ip: And I will add a comment as per why this methods need to be static 👍

NickPhura (approved):
👍

PR #1503, opened 2025-02-24
TECHDEBT: Enhanced Logger

MacDeluca:
## Description of Changes

- Enhances logger to include `requestId` and `user` (username) related metadata
- Also normalizes the logs to all be the same shape
- Note: Used a human readable value for `user` (ie: SBRULE) to make skimming / reading the logs slightly easier than using an inernal id or keycloak uuid
- Note: I am open to any alternative names for the new logger properties (requestId / user)

## Testing Notes

- For each request in the logs you should be able to see a requestId and user property included
- Request ID's should be generated fresh for each request, and trackable through the logs
- Log calls before the AsyncRequestStorage is initialized should show a fallback value for both properties (SYSTEM)

- In app.ts (line 196 after `initRequestStorage` middleware) you can add this snippet to test the requestIds / user in the logs.

```js
// 1. Run first request with `async-request-storage-test` header
// 2. Before the first request finishes, run a second request without the header
// 3. The second request should finish before the first request
// Validate that both requests have different request ids (Correct!)
additionalMiddleware.push(async function (
  req: Request,
  _res: Response,
  _next: NextFunction,
) {
  if (req.headers["async-request-storage-test"]) {
    console.log("Running SLOW request");

    await new Promise((resolve) => setTimeout(resolve, 10000));

    console.log("SLOW request done");
  } else {
    console.log("Running REGULAR request");
  }


  _next();
});

```


api/src/utils/async-request-storage.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @return {*} {void}
+ */
+export function initRequestStorage(req: Request, _res: Response, next: NextFunction) {
+  const requestStore: RequestStore = new Map();
+
+  // Generate the request id for the current request - unique for each request
+  requestStore.set('requestId', uuid());
+
+  // Set the username of the user who made the current request
+  requestStore.set('username', req.keycloak_token?.idir_username ?? req.keycloak_token?.bceid_username);
```
NickPhura: in `keycloak-utils.ts` there is an existing `getUserIdentifier` function, which I believe does the same thing here.

MacDeluca: Fixed!

NickPhura (approved):
🌳 🌲

PR #1505, opened 2025-02-24
BugFix: Fix Device Edits Not Updating Deployment 'device_key'

NickPhura:
## Links to Jira Tickets

n/a

### Database changes
Update device/deployment tables. Add foreign key between device and deployment on device_key, with cascade update.

### Manage telemetry UI changes
Update telemetry table to ensure the deployment and device columns always have some values populated, even if the critter request fails, or there is no critter in sims.

### Database changes
Previously: If you edited a device's serial number, the deployment would not be updated, and the `device_key` column would be out of sync between the 2 records.

Now: You should be able to edit a device's serial number, and the matching deployment record(s) should have their `device_key` column updated automatically to match.

### Manage telemetry UI changes
Previously: on the manage telemetry page, if a critterbase record is found for a deployment, the deployment column would show the deployment id and critter alias (ex: `'14: Big Tom'`). If no critterbase record was found, it would be empty.

Now: It will show the deployment_id and critter_id as a fallback (ex: `'14: 23'`)

## Regrading the bulk create/update manual telemetry changes
@MacQSL I think there may be a way to combine the security checks into the sql itself, so it doesn't have to be done in the service. But it was getting a bit ugly and needed more testing and I didn't want to delay the rest of these changes any longer.

### DIFF
--- a/app/src/features/surveys/telemetry/list/SurveyDeploymentListItem.tsx
+++ b/app/src/features/surveys/telemetry/list/SurveyDeploymentListItem.tsx
@@ -108,7 +108,7 @@ export const SurveyDeploymentListItem = (props: ISurveyDeploymentListItemProps) 
                 </Typography>
               </Stack>
               <Typography variant="body2" color="textSecondary" title="Animal">
-                {`${deployment.deployment_id}: ${animal?.animal_id || 'Unknown'}`}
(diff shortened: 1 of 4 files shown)


MacDeluca (approved):
👍🏼 Tested locally and tried updating the deployment and device

MacDeluca (approved):
💭

PR #1508, opened 2025-02-26
SIMSBIOHUB-580 initial telemetry csv export

oscar-ip:
## Links to Jira Tickets

- SIMSBIOHUB-580

## Description of Changes

- Implemented initial telemetry csv export

## Testing Notes

- performed basic local env testing


api/src/services/export-services/telemetry/export-telemetry-strategy.ts
```diff
@@ -92,4 +105,17 @@ export class ExportTelemetryStrategy extends DBService implements ExportStrategy
 
     return stream;
   };
+
+  static readonly generateTelemetryCSV = (item: Telemetry): string => {
+    const { dateStr, timeStr } = parseDateAndTimeString(item.acquisition_date);
+    return [
+      item.telemetry_id,
+      `${item.vendor}:${item.serial}`,
```
NickPhura: Can use the `getTelemetryDeviceKey` util function from `telemetry-utils.ts`, just to ensure the device_key is always constructed the same way, etc.

oscar-ip: good catch, done in the latest commit

api/src/services/export-services/export-utils.ts
```diff
@@ -62,12 +74,88 @@ export function getQueryParams(query: SQLStatement | Knex.QueryBuilder): { text:
...
+  header: string,
+  measurementsMap?: Map<string, string>
+): Transform {
+  // Function to get either option_label or measurement_name by id
+  const getLabelById = (id: string): string | undefined => {
+    return measurementsMap ? measurementsMap.get(id) : undefined;
+  };
+
+  const processMeasData = (measData: MeasDataItem[] | undefined): void => {
+    if (measData) {
```
NickPhura, line 119: Nit: rather than checking if measData is defined here, we could check it on line 155, and only call this function if it exists. And then we could remove the ` | undefined` from the parameter type here.

```typescript
if(chunk.meas_data) {
   processMeasData(chunk.meas_data);
}
```

api/src/services/export-services/export-utils.ts
```diff
@@ -99,3 +187,65 @@ export function registerStreamErrorHandler(stream: Readable): Readable {
...
+};
+
+/**
+ * Check if a value is a uuid or not.
+ * Optimized for performance
+ *
+ * @param {string} uuid
+ * @returns {boolean} true if it is a uuid
+ */
+export const isUUID = (uuid: string | null | undefined): boolean => {
```
NickPhura, line 231: The uuid library we use to generate UUIDs also has a validation function that might be relevant here

[uuid.validate(str)](https://www.npmjs.com/package/uuid#uuidvalidatestr)

oscar-ip: In this case I didn't used the library because of performance, the string comparison was faster.

api/src/services/export-services/export-utils.ts
oscar-ip: good point, I adjusted in the latest commit

api/src/services/export-services/export-utils.ts
NickPhura: Gotcha, fair

api/src/services/export-services/export-utils.ts
```diff
@@ -62,12 +74,88 @@ export function getQueryParams(query: SQLStatement | Knex.QueryBuilder): { text:
...
+
+/**
+ * Get a query data record transform stream, that expects objects and outputs csv.
+ *
+ * Note: The incoming data stream must yield objects, or this will throw an error.
+ *
+ * @export
+ * @returns {Transform}
+ */
+export function getCsvTransformStream(
```
NickPhura, line 108: Parts of this function are generic, and could be re-used. But other parts, like the measurements part, seem to be specific to certain exports.

I wonder if this function should be generated by each of the individual export strategies. And then it could cater to exactly what data is needed by each one. Some of the generic parts could be util functions, and others could be inline for each export, as needed.

You could still return it, and it could still be called in the root export function. But then the root export function wouldn't need to know about measurements, or any other data specific to one export.

(Not sure how drastic of a change this would be, so this is possibly more of an enhancement/TODO for the next team, if you think its valid).

NickPhura, line 108: Parts of this function are generic, and could be re-used. But other parts, like the measurements part, seem to be specific to certain exports.

I wonder if this function should be generated by each of the individual export strategies. And then it could cater to exactly what data is needed by each one. Some of the generic parts could be util functions, and others could be inline for each export, as needed.

You could still return it, and it could still be called in the root export function. But then the root export function wouldn't need to know about measurements, or any other data specific to one export.

(Not sure how drastic of a change this would be, so this is possibly more of an enhancement/TODO for the next team, if you think its valid).

NickPhura:
I tried seeding some data locally, specifically just observations, and it seems to work well. I think I tried with 400 observations (1200 subcounts). Just using the existing seeds. It also exported like 18000 telemetry points. The telemetry took a few seconds to parse, but overall seemed pretty fast.

The frontend UI probably needs some loading bars, etc, to show its in progress. And maybe a message when its done to remind them to check their downloads. But someone else can add that later on easily enough.

api/src/services/export-services/export-utils.ts
oscar-ip: yes, this is a generic function that supports all the query based exports and has some exceptions that deal with specific exports. 

I thought about separating but the function would be the same for few exports and different for only 2.
I wasn't expecting it to grow further at the time I took the decision and I agree this could use some refactoring :-)

I will leave it as another TODO item for the next team as there is no much time to test the changes and I rather not risk breaking inadvertently something.

oscar-ip:
> I tried seeding some data locally, specifically just observations, and it seems to work well. I think I tried with 400 observations (1200 subcounts). Just using the existing seeds. It also exported like 18000 telemetry points. The telemetry took a few seconds to parse, but overall seemed pretty fast.
> 
> The frontend UI probably needs some loading bars, etc, to show its in progress. And maybe a message when its done to remind them to check their downloads. But someone else can add that later on easily enough.

Great, happy to hear that the performance is acceptable, I basically kept that in mind all along for queries and code. I timed all queries and also specific sections of code to ensure that I use the best performing solution.

I agree, the front end user experience could be improved, fortunately even if you export all survey data it is fairly quick and I assume it will be long before the system has significant amounts of data.  I agree again, next team can take on this task I will add that as a TODO to the ticket we have.

api/src/services/export-services/export-utils.ts
oscar-ip: I added this to the TODOs in ticket SIMSBIOHUB-680

NickPhura (approved):
🦖

PR #1509, opened 2025-02-27
New Environmental Variable - Visibility

LouisThedroux:
## Links to Jira Tickets

NA

## Description of Changes

New Quantitative measurement - Visibility 0-10,000 range in meters added to environment_quant table in new migration file.


MacDeluca (approved):
🩴

PR #1510, opened 2025-02-27
Observations Table - Environment Columns to display units associated with that variable/column

LouisThedroux:
## Links to Jira Tickets
NA

## Description of Changes

Updates the environment column definitions to include units in the column header names for environment measurements.
Also adds 's' in the context that the unit ends with 'er'

### DIFF
--- a/app/src/features/surveys/observations/observations-table/grid-column-definitions/GridColumnDefinitionsUtils.tsx
+++ b/app/src/features/surveys/observations/observations-table/grid-column-definitions/GridColumnDefinitionsUtils.tsx
@@ -104,7 +104,13 @@ export const getEnvironmentColumnDefinitions = (
   for (const environment of environments.quantitative_environments) {
     colDefs.push(
       ObservationQuantitativeEnvironmentColDef({
-        environment: environment,
+        environment: {
+          ...environment,
+          name: environment.unit
+            ? `${environment.name} (${environment.unit}${environment.unit.endsWith('er') ? 's' : ''})`
+            : environment.name,
+          description: environment.description
+        },
         hasError: hasError
       })
     );
--- a/app/src/features/surveys/observations/observations-table/grid-column-definitions/GridColumnDefinitionsUtils.tsx
+++ b/app/src/features/surveys/observations/observations-table/grid-column-definitions/GridColumnDefinitionsUtils.tsx
@@ -96,21 +96,35 @@ export const getMeasurementColumnDefinitions = (
   return colDefs;
 };
 
+const getQuantitativeEnvironmentHeaderName = (name: string, unit?: string): string => {
+  if (!unit) {
+    return name;
+  }
+
+  const unitSuffix = unit.endsWith('er') ? 's' : '';
+  return `${name} (${unit}${unitSuffix})`;
+};
+
+const formatQuantitativeEnvironment = (environment: EnvironmentQuantitativeTypeDefinition) => {
+  return {
+    ...environment,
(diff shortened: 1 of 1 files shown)


app/src/features/surveys/observations/observations-table/grid-column-definitions/GridColumnDefinitionsUtils.tsx
```diff
@@ -104,7 +104,13 @@ export const getEnvironmentColumnDefinitions = (
   for (const environment of environments.quantitative_environments) {
     colDefs.push(
       ObservationQuantitativeEnvironmentColDef({
-        environment: environment,
+        environment: {
+          ...environment,
+          name: environment.unit
+            ? `${environment.name} (${environment.unit}${environment.unit.endsWith('er') ? 's' : ''})`
```
MacDeluca: suggestion: We should place this logic inside its own function to use if statements instead of nested ternarys

ie: getQuantitativeEnvironmentHeaderName

MacQSL: There is also a `pluralize` util function that might be of use

NickPhura: I'd also move this modification code outside of the `colDefs.push(...)`

```
  for (const environment of environments.quantitative_environments) {
   const formattedEnvironment = <do stuff to modify the name, call getQuantitativeEnvironmentHeaderName, etc>

    colDefs.push(
      ObservationQuantitativeEnvironmentColDef({
        environment: formattedEnvironment,
        hasError: hasError
      })
    );
  }
  ```

**Review by NickPhura — APPROVED** [review]
Updates look great 👍

File: app/src/features/surveys/observations/observations-table/grid-column-definitions/GridColumnDefinitionsUtils.tsx [hunk 2]
```diff
@@ -96,15 +96,35 @@ export const getMeasurementColumnDefinitions = (
   return colDefs;
 };
 
+const getQuantitativeEnvironmentHeaderName = (name: string, unit?: string): string => {
+  if (!unit) {
+    return name;
+  }
+
+  const unitSuffix = unit.endsWith('er') ? 's' : '';
```
mauberti-bc, line 104: We have a ```pluralize``` function that does a similar thing. We should put this in utils as ```ensurePlural```:

```ensurePlural(word, endConditions)```, which might look like ```ensurePlural(unit, ['er'])```, so if the unit ends with any of the conditions, it will add an s. And if we have inconsistent spelling, we could do ```['er', 're']```

File: app/src/features/surveys/observations/observations-table/grid-column-definitions/GridColumnDefinitionsUtils.tsx [hunk 3]
```diff
@@ -96,15 +96,35 @@ export const getMeasurementColumnDefinitions = (
...
+  return `${name} (${unit}${unitSuffix})`;
+};
+
+const formatQuantitativeEnvironment = (environment: EnvironmentQuantitativeTypeDefinition) => {
+  return {
+    ...environment,
+    name: getQuantitativeEnvironmentHeaderName(environment.name, environment.unit ?? undefined),
+    description: environment.description
```
mauberti-bc, line 112: Not a big deal you shouldn't need to have ```description: environment.description``` because it will be included in the ```...environment```

PR #1517, opened 2025-03-04
FEATURE: Habitat Feature Table Improvements

MacDeluca:
## Description of Changes

- Updated columns for Habitat Features Table
- Created menu and controls for toggling column visibility
- Updated `useUnsavedChangesDialog` hook and provided additional callback to bypass the dialog
- Updated instances where we were passing `SKIP_CONFIRMATION_DIALOG` as state to location changes (now explicity handled with the callback using a ref)
- Editing the sample site from the `Habitat Features` page / `Observations` page should route back to the original page on save / cancel

## Testing Notes
- App should function the same, forms should route to the correct places
- Columns visibility should be toggleable

### DIFF
--- a/app/src/features/surveys/habitat-features/SurveyHabitatFeatureRouter.tsx
+++ b/app/src/features/surveys/habitat-features/SurveyHabitatFeatureRouter.tsx
@@ -26,6 +26,12 @@ export const HabitatFeatureRouter = () => {
         to="/admin/projects/:id/surveys/:survey_id/sampling/:survey_sample_site_id/edit"
       />
 
+      <Redirect
+        exact
+        from="/admin/projects/:id/surveys/:survey_id/habitat-features/sampling"
+        to="/admin/projects/:id/surveys/:survey_id/sampling"
+      />
+
       <RouteWithTitle
         exact
         path="/admin/projects/:id/surveys/:survey_id/habitat-features/details"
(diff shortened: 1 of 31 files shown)


app/src/features/surveys/habitat-features/SurveyHabitatFeatureRouter.tsx
```diff
@@ -26,6 +26,12 @@ export const HabitatFeatureRouter = () => {
         to="/admin/projects/:id/surveys/:survey_id/sampling/:survey_sample_site_id/edit"
       />
 
+      <Redirect
+        exact
+        from="/admin/projects/:id/surveys/:survey_id/habitat-features/sampling"
+        to="/admin/projects/:id/surveys/:survey_id/sampling"
+      />
+
```
NickPhura, line 34: I take it this is needed to make `history.goBack()` work? Something to do with the sampling path being a child of the habitat path?

MacDeluca: It was just appending the `sampling` route to the end of the path when you click that button from the `Habitat Feature` table. So this just redirects the request to the correct page. 

I didn't extensively look at this part, so if there is a better way to do it I am open

surveys/:survey_id/habitat-features/sampling -> nothing
surveys/:survey_id/sampling -> Manage Sampling page

MacDeluca: I think also, that the history.goBack doesn't care about the path structure, it just points to the previous page in the history stack.

NickPhura: Oh I see, it should be possible to navigate to the sampling page without redirects here.

But I don't actually know for sure. I'd have to play with it, I dont remember all of the details around the routes, and their nuances. 

I was thinking it might help with back button behaviour, but then realized we dont actually have an in-app back button, so it wouldnt help anyways.

NickPhura: For now, it doesnt really matter, we'll eventually go to react-router 6 and all of this will have to be scrutinized/updated anyways.

NickPhura (approved):
🌲

PR #1518, opened 2025-03-04
Add Environment Variables - Various improvements. 

LouisThedroux:
## Links to Jira Tickets
NA

## Description of Changes

Regarding the Add Environment Variables section of the Configure Columns dialogue box on the Manage Observations page.

- Modified environment search to display all options on initial click/focus
- Added alphabetical sorting to environment options list
- Added the env.units to the drop down so the user can see what unit it is measured in without clicking/adding and then having to remove if its not what they want
- Added unit pluralization for measurements ending in 'er' (e.g., "meter" → "meters")

### DIFF
--- a/app/src/features/surveys/observations/observations-table/configure-columns/components/environment/search/EnvironmentsSearchAutocomplete.tsx
+++ b/app/src/features/surveys/observations/observations-table/configure-columns/components/environment/search/EnvironmentsSearchAutocomplete.tsx
@@ -2,6 +2,7 @@ import { mdiMagnify } from '@mdi/js';
 import Icon from '@mdi/react';
 import Autocomplete from '@mui/material/Autocomplete';
 import Box from '@mui/material/Box';
+import Chip from '@mui/material/Chip';
 import ListItem from '@mui/material/ListItem';
 import Stack from '@mui/material/Stack';
 import TextField from '@mui/material/TextField';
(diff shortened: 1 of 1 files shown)


LouisThedroux:
Let wait for @mauberti-bc to have a look at this first. Maybe he is not happy with the idea of it and no point reviewing code thats not going to be used

app/src/features/surveys/observations/observations-table/configure-columns/components/environment/search/EnvironmentsSearchAutocomplete.tsx
```diff
@@ -69,6 +70,20 @@ export const EnvironmentsSearchAutocomplete = (props: IEnvironmentsSearchAutocom
...
+    const response = await getOptions('');
+    const allOptions = [...response.qualitative_environments, ...response.quantitative_environments];
+
+    const sortedOptions = allOptions.sort((a, b) => a.name.localeCompare(b.name));
+    setOptions(sortedOptions);
+  };
+
+  // This function is a repeat of a function in another file for pluralising. It should be edited to an imported function. GridColumnDefinitionsUtils.tsx when PR goes through
+  const getFormattedUnit = (unit: string) => {
+    const unitSuffix = unit.endsWith('er') ? 's' : '';
```
NickPhura, line 83: I think this is the second time we've seen this logic. I would suggest making it a standalone function.

I would add a new function next to the `pluralize` function in `.../utils/Utils.tsx` since it tackles a similar kind of issue.

mauberti-bc:
Closing because this is outdated after recent changes

PR #1519, opened 2025-03-05
FEATURE: Habitat feature create page

MacDeluca:
## Description of Changes

- Adds frontend route for creating habitat feature
- New form components for create
- 

## Testing Notes

- User should be able to create a habitat feature, see a success snackbar and view the new record in the table


app/src/features/surveys/habitat-features/create/CreateHabitatFeaturePage.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+      skipUnsavedChangesDialog();
+      history.goBack();
+
+      dialogContext.setSnackbar({
+        open: true,
+        snackbarMessage: 'Successfully created habitat feature'
+      });
+
+      await habitatFeatureContext.refreshData();
```
NickPhura: Do we need to await this refresh? 

Also, does the history.goBack() need to be balled after? I assume it gets queued up so it works fine as is?

MacDeluca: Ok I swapped the order of operations for this section, and dropped the await for the refresh.

NickPhura (approved):
👍

PR #1522, opened 2025-03-06
Habitat feature Branch: Some Api Tests + Import Happy Path

NickPhura:
## Description of Changes

Added some API unit tests.

Import habitat features happy path is working.

Bugfix: Fixed the create and update openapi schema, and related types, to remove `survey_id`


api/src/utils/csv-utils/row-validators/taxon-row-validator.ts
```diff
@@ -27,6 +27,10 @@ export const getTaxonRowValidator = <StaticHeaderType extends Uppercase<string>
     const taxonHeader = utils.getWorksheetHeader(taxonStaticHeader, params.row);
 
     if (taxonIdentifierCell === undefined) {
+      if (utils.isCellOptional(taxonStaticHeader)) {
+        return [];
+      }
+
```
NickPhura: @MacQSL Let me know if this addition, and the related utils method, is okay or not. Or if there is a better way to handle this case, etc.

I added this check to handle the species being optional in the habitat csv file. 
I haven't added anything to handle multiple species values.

MacDeluca: From our discord discussion:

The optional property for a standard column of the CSVConfig just indicates that the column can be omitted from the CSV. Different than allowing undefined / empty values but the header must be defined still.

I think all cases currently where we allow undefined values we also allow omitting the header (optional). For future proofing, we can provide a property to this row-validator to allow undefined values, seperate from the column optional logic.

api/src/utils/csv-utils/csv-config-utils.ts
```diff
@@ -236,4 +236,14 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string> = Upperca
...
     return uniqueDictionary[dictionaryKey] === 1 || uniqueDictionary[dictionaryKey] === undefined;
   }
+
+  /**
+   * Check if a cell is optional.
+   *
+   * @param {StaticHeaderType} header - The header name
+   * @return {boolean} - Whether the cell is optional
+   */
+  isCellOptional(header: StaticHeaderType): boolean {
```
MacDeluca: see previous comment

api/src/utils/csv-utils/row-validators/taxon-row-validator.ts
NickPhura: Updated.

api/src/utils/csv-utils/csv-config-utils.ts
NickPhura: Updated too :)

MacDeluca (approved):
🩴

PR #1527, opened 2025-03-12
Delete and Deactivate System Users

mauberti-bc:
## Links to Jira Tickets

- N/A

## Description of Changes

- Bug fix: Prevent users with a record_end_date from requesting access
- Update role guards to show the fallback for deactivated users, regardless of the user's roles
- Update middleware authorization check to fail for deactivated users

## Testing Notes

- You can deactivate (=block) yourself in the active users table to test a blocked user's perspective


MacDeluca (approved):
🦖 Code looks great!
Tested locally and it works as expected.

PR #1529, opened 2025-03-12
Habitat Features: Import Taxon Array

NickPhura:
### Changes
WIP: Add new functions to support validating csv cells which contain an array of values (ie: `value1; value2; value3`)

### TODO
- Test it against sample CSVs
- Add/update unit tests

### DIFF
--- a/api/src/utils/csv-utils/csv-config-utils.ts
+++ b/api/src/utils/csv-utils/csv-config-utils.ts
@@ -213,16 +213,14 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string> = Upperca
    * @return {(string[] | undefined)} - The array-cell values or undefined if none found
    */
   getArrayCellValue(header: StaticHeaderType, row: CSVRow, options: { delimiter: string }): string[] | undefined {
-    let cellValue: CSVCell;
-
     // Static header or dynamic header exact match
-    if ((header as Uppercase<string>) in row) {
-      cellValue = row[header];
-    }
+    let cellValue = this.getCellValue(header, row);
 
-    // Attempt to find the array-cell values from the header aliases
-    for (const alias of this.config.staticHeadersConfig[header]?.aliases ?? []) {
-      cellValue = row[alias];
+    if (!cellValue) {
+      // Attempt to find the array-cell values from the header aliases
+      for (const alias of this.config.staticHeadersConfig[header]?.aliases ?? []) {
+        cellValue = row[alias];
+      }
     }
 
     if (!cellValue) {
@@ -253,8 +251,8 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string> = Upperca
    * @param {StaticHeaderType} header - The header name
    * @return {((string | undefined)[])} - The array-cell values
    */
-  getArrayCellValues(header: StaticHeaderType): (string | undefined)[] {
-    return this.worksheetRows.flatMap((row) => this.getArrayCellValue(header, row, { delimiter: ';' }));
+  getArrayCellValues(header: StaticHeaderType, options: { delimiter: string }): (string | undefined)[] {
+    return this.worksheetRows.flatMap((row) => this.getArrayCellValue(header, row, options));
   }
 
   /**
@@ -273,8 +271,8 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string> = Upperca
    * @param {StaticHeaderType} header - The header name
    * @returns {((string | undefined)[])} - The unique array-cell values
    */
-  getUniqueArrayCellValues(header: StaticHeaderType): (string | undefined)[] {
-    return [...new Set(this.getArrayCellValues(header))];
+  getUniqueArrayCellValues(header: StaticHeaderType, options: { delimiter: string }): (string | undefined)[] {
+    return [...new Set(this.getArrayCellValues(header, options))];
   }
 
   /**
--- a/api/src/utils/csv-utils/csv-config-utils.ts
+++ b/api/src/utils/csv-utils/csv-config-utils.ts
@@ -188,7 +188,7 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string> = Upperca
    *
    * @param {StaticHeaderType} header - The header name
    * @param {CSVRow} row - The CSV row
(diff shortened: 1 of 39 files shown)


api/src/utils/csv-utils/csv-config-utils.ts
```diff
@@ -204,26 +204,79 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string> = Upperca
...
+   * @param {StaticHeaderType} header - The header name
+   * @param {CSVRow} row - The CSV row
+   * @param {{ delimiter: string }} options - The options
+   * @return {(string[] | undefined)} - The array-cell values or undefined if none found
+   */
+  getArrayCellValue(header: StaticHeaderType, row: CSVRow, options: { delimiter: string }): string[] | undefined {
+    let cellValue: CSVCell;
+
+    // Static header or dynamic header exact match
+    if ((header as Uppercase<string>) in row) {
```
MacDeluca: nit: I think this section can be replaced with `this.getCellValue`

api/src/utils/csv-utils/csv-header-configs.ts
```diff
@@ -72,10 +71,10 @@ export const validateZodCell = (cell: unknown, schema: z.ZodSchema, solution?: s
  *  1. The cell must be a positive number
  *  2. The cell is optional if the optional flag is set
  *
- * @param {CSVOptionalCell} [options] Optional cell config override
+ * @param {CSVCellValidatorOptions} [options] Optional cell config override
```
MacDeluca, line 58: praise: This is a good change

api/src/utils/csv-utils/csv-config-utils.ts
```diff
@@ -204,26 +204,79 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string> = Upperca
...
   }
 
+  /**
+   * Get all the array-cell values from a static header.
+   *
+   * @param {StaticHeaderType} header - The header name
+   * @return {((string | undefined)[])} - The array-cell values
+   */
+  getArrayCellValues(header: StaticHeaderType): (string | undefined)[] {
+    return this.worksheetRows.flatMap((row) => this.getArrayCellValue(header, row, { delimiter: ';' }));
```
MacDeluca: question: Should the delimiter be hardcoded?

api/src/utils/csv-utils/csv-header-configs.ts
```diff
@@ -369,3 +368,122 @@ export const getLookupIdCellValidator = (
...
+ *
+ * Rules:
+ *  1. The cell must be a string, with optional values separated by a delimiter.
+ *  2. The individual cell values (after splitting by delimiter) must satisfy the cell validator.
+ *
+ * @param {csvCellValidator} csvCellValidator The cell validator to apply to each element in the array
+ * @param {CSVArrayCellValidatorOptions} options The array-cell validator options
+ * @returns {*} {CSVCellValidator} The validate cell callback
+ */
+export const getArrayCellValidator: CSVArrayCellValidator = (
```
MacDeluca: thought: Maybe this function should only accept two params: (csvCellValidator: CSVCellValidator, options: CSVArrayCellValidatorOptions). This would allow array validators to accept "array" cells of other types ie: numbers, booleans etc. That way the caller would have full control of the cell validation.

MacDeluca: And in the case of `getTaxonCellValidator` we have all the information needed to generate the CSVErrors. This would allow us to drop the `getNonEmptyStringCellValidator` from the internal logic of this function.

api/src/utils/csv-utils/csv-header-configs.ts
```diff
@@ -369,3 +368,122 @@ export const getLookupIdCellValidator = (
...
+ * Rules:
+ *  1. The cell must be a valid ITIS TSN or scientific name
+ *  2. The cell must be a valid species from the provided taxons
+ *
+ * @template StaticHeaderType
+ * @param {TaxonMap} taxonMap The list of taxons
+ * @param {CSVCellValidatorOptions} [options] cell validator options
+ * @returns {*} {CSVCellValidator} The validate cell callback
+ */
+export const getTaxonCellValidator = (taxonMap: TaxonMap, options?: CSVCellValidatorOptions): CSVCellValidator => {
```
MacDeluca, line 375: question: Can the `getTaxonRowValidator` call this function now?

api/src/utils/csv-utils/csv-config-utils.ts
```diff
@@ -204,26 +204,77 @@ export class CSVConfigUtils<StaticHeaderType extends Uppercase<string> = Upperca
...
+   * @param {StaticHeaderType} header - The header name
+   * @param {CSVRow} row - The CSV row
+   * @param {{ delimiter: string }} options - The options
+   * @return {(string[] | undefined)} - The array-cell values or undefined if none found
+   */
+  getArrayCellValue(header: StaticHeaderType, row: CSVRow, options: { delimiter: string }): string[] | undefined {
+    // Static header or dynamic header exact match
+    let cellValue = this.getCellValue(header, row);
+
+    if (!cellValue) {
```
MacDeluca: note: This block from 219 -> 224 is not need I don't think. `getCellValue` already handles the aliases

api/src/utils/csv-utils/csv-config-utils.ts
NickPhura: fixed

api/src/utils/csv-utils/csv-config-utils.ts
NickPhura: fixed

api/src/utils/csv-utils/csv-header-configs.ts
NickPhura: fixed

api/src/utils/csv-utils/csv-header-configs.ts
NickPhura: fixed

api/src/utils/csv-utils/csv-config-utils.ts
NickPhura: fixed

MacDeluca (approved):
🌡️

PR #1530, opened 2025-03-13
BugFix: Change `deployment.frequency` Column Type To Decimal

NickPhura:
## Links to Jira Tickets

## Description of Changes

Migration to change `deployment.frequency` column type to decimal.

Before it was set to integer, which caused frequency values to be quietly truncated.

## Testing Notes

Create and/or edit a deployment with a frequency that has decimals (ex: `150.123`).
- See that in the database it is stored as `150.123`.
- See that in the UI it is displayed as `150.123`


MacDeluca (approved):
👍🏼

PR #1541, opened 2025-03-31
TechDebt: Fix API Taxon Response When BioHub Service Unavailable

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Add better fallbacks for taxon data when the biohub service is unavailable. Just happened to come across this particular instance because I had a bad biohub url in my env, and the survey failed to load as a result.

Fix issue with the observation popup SQL (ambiguous survey_id).

## Testing Notes

If you use a bad biohub url in your env, the survey page should not fail to load (previously it would, because the api response validation failed due to missing common names and scientific name)

Clicking an observation on the survey map should load the popup (previously there was a sql error due to ambiguous survey_id)


MacDeluca (dismissed):

MacDeluca (approved):
🩴

PR #1545, opened 2025-04-17
Survey Collections API Backend

LouisThedroux:
## Description of Changes
Feature branch for API and front end collections work. To be merged into collections before dev

Initially done endpoint index.ts and collection-service.ts

### DIFF
--- a/api/src/paths/collections/index.ts
+++ b/api/src/paths/collections/index.ts
@@ -1,40 +1,36 @@
+import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
-import { Request, Response } from 'express';
-import { getAPIUserDBConnection } from '../../database/db';
+import { authorizeRequestHandler } from '../../request-handlers/security/authorization';
+import { getDBConnection } from '../../database/db';
 import { CollectionService } from '../../services/collection-service';
 import { getLogger } from '../../utils/logger';
+import { SYSTEM_ROLE } from '../../constants/roles';
+import { getSystemUserFromRequest } from '../../utils/request';
+import { userHasValidRole } from '../../request-handlers/security/authorization';
 
 const defaultLog = getLogger('paths/collections');
-
-const collectionService = new CollectionService();
 
 /**
  * GET /api/collection
  * Fetch all collections.
  */
-export const GET: Operation = async (req: Request, res: Response) => {
-  const connection = getAPIUserDBConnection();
-
-  try {
-    await connection.open();
-
-    const collections = await collectionService.getAllCollections(connection);
-
-    await connection.commit();
-
-    res.status(200).json(collections);
-  } catch (error) {
-    defaultLog.error({ label: 'GET /api/collection', message: 'error', error });
-    connection.rollback();
-    res.status(500).json({ error: 'Failed to fetch collections' });
-  } finally {
-    connection.release();
-  }
-};
+export const GET: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          discriminator: 'SystemUser'
+        }
+      ]
+    };
+  }),
+  findCollections()
+];
 
 GET.apiDoc = {
   description: 'Fetch all collections.',
   tags: ['collection'],
+  security: [{ Bearer: [] }],
   responses: {
     200: {
       description: 'List of collections.',
@@ -54,8 +50,20 @@ GET.apiDoc = {
         }
       }
     },
+    400: {
+      $ref: '#/components/responses/400'
+    },
+    401: {
+      $ref: '#/components/responses/401'
+    },
+    403: {
+      $ref: '#/components/responses/403'
+    },
     500: {
-      description: 'Internal server error.'
+      $ref: '#/components/responses/500'
+    },
+    default: {
+      $ref: '#/components/responses/default'
     }
   }
 };
@@ -64,43 +72,23 @@ GET.apiDoc = {
  * POST /api/collection
  * Create a new collection.
  */
-export const POST: Operation = async (req: Request, res: Response) => {
-  const connection = getAPIUserDBConnection();
-
-  try {
-    await connection.open();
-
-    const newCollection = await collectionService.createCollection(connection, req.body);
-
-    await connection.commit();
-
-    res.status(201).json(newCollection);
-  } catch (error) {
-    defaultLog.error({ label: 'POST /api/collection', message: 'error', error });
-    connection.rollback();
-    res.status(500).json({ error: 'Failed to create collection' });
-  } finally {
-    connection.release();
-  }
-};
+export const POST: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          discriminator: 'SystemUser'
+        }
+      ]
+    };
+  }),
+  createCollection()
+];
 
 POST.apiDoc = {
   description: 'Create a new collection.',
   tags: ['collection'],
-  requestBody: {
-    content: {
-      'application/json': {
-        schema: {
-          type: 'object',
-          required: ['name', 'objectives'],
-          properties: {
-            name: { type: 'string' },
-            objectives: { type: 'string' }
-          }
-        }
-      }
-    }
-  },
+  security: [{ Bearer: [] }],
   responses: {
     201: {
       description: 'Collection created successfully.',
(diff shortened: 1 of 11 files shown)


api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
+import { Operation } from 'express-openapi';
+import { Request, Response } from 'express';
+import { getAPIUserDBConnection } from '../../database/db';
+import { CollectionService } from '../../services/collection-service';
+import { getLogger } from '../../utils/logger';
+
+const defaultLog = getLogger('paths/collections');
+
+const collectionService = new CollectionService();
```
mauberti-bc: The service should be instantiated within the function (currently called Operation) below, to ensure that the instance is removed from memory when the function returns

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+
+/**
+ * GET /api/collection
+ * Fetch all collections.
+ */
+export const GET: Operation = async (req: Request, res: Response) => {
+  const connection = getAPIUserDBConnection();
```
mauberti-bc: This establishes a connection as the API service account instead of a user. Consequently, any database interactions/auditing that happen through this connection will reference the API service account (ie. created_by, updated_by columns will point to the API service account instead of the logged in user triggering the request).

 We should always use ```getDBConnection(req.user)```, passing in the user object of the request, except for requests that don't require authentication (eg. the standards page)

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+
+const defaultLog = getLogger('paths/collections');
+
+const collectionService = new CollectionService();
+
+/**
+ * GET /api/collection
+ * Fetch all collections.
+ */
+export const GET: Operation = async (req: Request, res: Response) => {
```
mauberti-bc: You're bypassing the middleware/auth checks with this, and we only want authorized users to be able to get collections. All the existing API functions follow this correct structure:

```
export const GET: Operation = [
  authorizeRequestHandler(() => {
    return {
      and: [
        {
          discriminator: 'SystemUser'
        }
      ]
    };
  }),
  getUsers()
];

export function getUsers(): RequestHandler { return ... }
```

By pointing the GET binding to ```[authorizeRequestHandler(), function()]``` instead of just ```function()```, you're essentially screening the request to make sure the user has permissions to call ```function```. Your current code always executes function(), which is dangerous.

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    const collections = await collectionService.getAllCollections(connection);
+
+    await connection.commit();
+
+    res.status(200).json(collections);
+  } catch (error) {
+    defaultLog.error({ label: 'GET /api/collection', message: 'error', error });
+    connection.rollback();
+    res.status(500).json({ error: 'Failed to fetch collections' });
```
mauberti-bc: You don't need to explicitly send a response in the catch statement. Instead, just throw an error (eg. ```catch (error) { throw error}```) and the middleware will send the correct response for the given error.

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+  } finally {
+    connection.release();
+  }
+};
+
+GET.apiDoc = {
+  description: 'Fetch all collections.',
+  tags: ['collection'],
+  responses: {
```
mauberti-bc, line 59: In addition to adding
```
export const GET: Operation = [
  authorizeRequestHandler(() => {
    return {
      and: [
        {
          discriminator: 'SystemUser'
        }
      ]
    };
  }),
  findCollections()
];
```

 you'll need to add ```security: [{Bearer: []}]``` to the GET.apiDoc object. This will ensure that the bearer token gets passed to the auth checks in the middleware. You can see this in any index.ts file under /paths

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+      description: 'Internal server error.'
+    }
+  }
+};
+
+/**
+ * POST /api/collection
+ * Create a new collection.
+ */
+export const POST: Operation = async (req: Request, res: Response) => {
```
mauberti-bc: Exact same comments from above apply here. Need to add the middleware.

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    const newCollection = await collectionService.createCollection(connection, req.body);
+
+    await connection.commit();
+
+    res.status(201).json(newCollection);
+  } catch (error) {
+    defaultLog.error({ label: 'POST /api/collection', message: 'error', error });
+    connection.rollback();
+    res.status(500).json({ error: 'Failed to create collection' });
```
mauberti-bc: Throw an error instead of returning a 500

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+                collection_id: { type: 'integer' },
+                name: { type: 'string' },
+                objectives: { type: 'string' }
+              }
+            }
+          }
+        }
+      }
+    },
+    500: {
```
mauberti-bc, line 105: You should copy-paste all the status codes from every other endpoint:
```
400: {
      $ref: '#/components/responses/400'
    },
    401: {
      $ref: '#/components/responses/401'
    },
    403: {
      $ref: '#/components/responses/403'
    },
    500: {
      $ref: '#/components/responses/500'
    },
    default: {
      $ref: '#/components/responses/default'
    }
```

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+ * POST /api/collection
+ * Create a new collection.
+ */
+export const POST: Operation = async (req: Request, res: Response) => {
+  const connection = getAPIUserDBConnection();
+
+  try {
+    await connection.open();
+
+    const newCollection = await collectionService.createCollection(connection, req.body);
```
mauberti-bc: You shouldn't be passing the connection to a service's method. Instead, you should pass the connection when you establish an instance of the service. 
```
const collectionService= new CollectionService(connection)
```
Notice that all the existing services extend DBService, which handles the connection. When you pass connection to CollectionService, the instance will hang onto that connection for methods to reference - so when you call createCollection, it will get the connection that the class was instantiated with.

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+      description: 'Internal server error.'
+    }
+  }
+};
+
+/**
+ * GET /api/collection/{collection_id}
+ * Fetch a collection by ID.
+ */
+export const GET_BY_ID: Operation = async (req: Request, res: Response) => {
```
mauberti-bc: We should follow our current API structure and REST pattern by creating this folder structure:

```plaintext
- paths/
  └─ collections/
     ├─ index.ts
     ├─ index.test.ts
     └─ {collectionId}/
        ├─ index.ts
        └─ index.test.ts
```

And the endpoint for getting a specific collection should be structured EXACTLY like the code for getting all collections. The only difference should be a ```parameters``` property in the GET.apiDoc for getting a specific collection, indicating that a collectionId must be given in the API request

api/src/paths/collections/index.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    const updatedCollection = await collectionService.updateCollection(
+      connection,
+      Number(req.params.collection_id),
+      req.body
+    );
+
+    await connection.commit();
+
+    if (!updatedCollection) {
```
mauberti-bc: We shouldn't do this check here. Instead, we should do the check in the repository function where the SQL gets run. If we throw an error in the repository function, that error will propagate up such that the request responds with a 500 error.

api/src/services/collection-service.ts
```diff
@@ -0,0 +1,148 @@
+import { IDBConnection } from '../database/db';
+import { SQL, SQLStatement } from 'sql-template-strings';
+
+export class CollectionService {
```
mauberti-bc: This should extend the DBService, like how the ProjectService and all other services do.

api/src/services/collection-service.ts
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * Fetch all collections.
+   *
+   * @param {IDBConnection} connection
+   * @return {*}  {Promise<any[]>}
+   * @memberof CollectionService
+   */
+  async getAllCollections(connection: IDBConnection): Promise<any[]> {
```
mauberti-bc: All this new code should exactly match the patterns and structure of our API.

Our API essentially has 3 levels:
- Endpoint
- Service
- Repository

The endpoint code is in the paths/ directory and handles the request itself. The service methods contain business rules and logic. The repository methods interact with the database. 

* I like to think of the endpoint as the gatekeeper - if the request is allowed to proceed, the gatekeeper will call the service. 
* I think of the service as a decision tree of what to do based on the data: usually, the service doesn't need to check anything and just relays the inputs to the repository. 
** But, for example, if a database record is being deleted, the service will ensure any dependent database records are deleted first to avoid foreign key constraint errors. 
* The repository is "dumb": it doesn't have logical checks, it just calls SQL and validates that the output from the database matches expectations.

api/src/services/collection-service.ts
```diff
@@ -0,0 +1,148 @@
...
+      INSERT INTO collection (
+        name,
+        objectives,
+        create_date,
+        create_user
+      ) VALUES (
+        ${collectionData.name},
+        ${collectionData.objectives},
+        now(),
+        1 -- Replace with actual user ID
```
mauberti-bc: 👀

api/src/services/collection-service.ts
```diff
@@ -0,0 +1,148 @@
...
+        update_date = now(),
+        update_user = 1 -- Replace with actual user ID
+      WHERE
+        collection_id = ${collectionId}
+        AND record_end_date IS NULL;
+    `;
+
+    const response = await connection.sql(sql);
+
+    return response.rowCount !== null && response.rowCount > 0;
```
mauberti-bc: Here is where you would ```throw``` an error if the response doesn't match expectations, instead of in the endpoint/service code.  You'll want to include ```RETURNING collection_id``` or ```RETURNING *``` in the SQL, so that you have a response.rows to validate. 

It's also good to check ```response.rowCount === 1```  if you know you're updating/deleting a single collection (something is wrong if the row count is > 1).

Here's an example
```
const sqlStatement = SQL`
      UPDATE alert
      SET
        name = ${alert.name},
        message = ${alert.message},
        alert_type_id = ${alert.alert_type_id},
        severity = ${alert.severity},
        data = ${JSON.stringify(alert.data)}::json,
        record_end_date = ${alert.record_end_date}
      WHERE
        alert_id = ${alert.alert_id}
      RETURNING alert_id
      ;
      `;

    const response = await this.connection.sql(sqlStatement);

    if (response.rowCount !== 1) {
      throw new ApiExecuteSQLError('Failed to update alert', [
        'AlertRepository->updateAlert',
        'rowCount was !== 1, expected rowCount === 1'
      ]);
    }

    return response.rows[0].alert_id;
  }
```

app/src/hooks/api/useCollectionApi.ts
```diff
@@ -0,0 +1,70 @@
+import axios from 'axios';
```
mauberti-bc, line 1: This is just vanilla axios, meaning that you'll need to add headers, contentType, etc. to every request. Instead, you should use our ```useAxios``` hook in useAxios.ts. This hook formats every request, refreshes auth tokens, and formats errors. 

When creating a new api hook, you should following existing hooks EXACTLY. Almost everything looks the way it does for a reason, but we should assume that _everything_ looks that way for a reason.

database/src/migrations/20250401101427_collections.ts
mauberti-bc, line 1: Because this code was added from a different PR, you shouldn't change it here. Instead, go to the original branch and change it there. 

If this file gets changed in the base branch for whatever reason, the 2 branches will now have conflicts.

mauberti-bc:
Looks like you touched most of what goes into our API, which is awesome! Hopefully it's starting to make some sense. My general comment is that new code should _*exactly*_ match existing patterns: create an endpoint in the paths/ directory, create a new service, create a new repository. 

It should feel very natural to just copy-paste code from a similar feature (eg. get/create collections is analogous to get/create projects) and do some strategic find-all-replace-ing.

I'd suggest starting by copy-pasting all the code for getting a list of projects that the user has access to. You can update the seed file to include some sample collections so the endpoint actually returns real data, letting you validate the OpenAPI schema.

For getting a list of projects with filter parameters, we're started using the convention of ```findCollections``` instead of ```getCollections```, implying that there might be none and we're just searching with some filter parameters.

api/src/paths/collections/index.ts
LouisThedroux: I have created a Get Operation with findCollections within it to fetch all collections which has authorizeRequestHandler to ensure permissions. However, would there ever be a time where fetching all collections is appropriate? Wouldn't we want it to be refined to collections associated to that specific user. Does this functionality sit at the paths/collections/index.ts file or would it be in something like paths/collections/{userId}/index.ts ?

mauberti-bc:
Closing because I copied the relevant code into the collections branch, so this is accounted for 👍 Looks good after your changes!

PR #1549, opened 2025-04-30
Updating technique form and adding methods and attributes

AMEIJER1:
## Description of Changes

- Removes detection distance from the Technique form (frontend only)
- Adds PIT Tag Reader as a new method


LouisThedroux:
LGTM, I just added a comment to the TechniqueDetailsForm file, Its not used anywhere else in the app so I think it could be delete , but we may want to keep it for future use or reference or something. But at the same time doing this a lot would cause a lot of unecessary code just sitting in the app.

mauberti-bc:
> LGTM, I just added a comment to the TechniqueDetailsForm file, Its not used anywhere else in the app so I think it could be delete , but we may want to keep it for future use or reference or something. But at the same time doing this a lot would cause a lot of unecessary code just sitting in the app.

Best to just delete unused code. If we ever delete a complex feature, we can go find it in the git history and copy relevant code. The deleted code is very simple though

mauberti-bc (approved):
Looks great! I removed 2 of the new method attributes because they didn't seem common enough to warrant cluttering the list.

PR #1550, opened 2025-05-01
Bulk import functionality for telemetry devices

AMEIJER1:
## Description of Changes

- Added the ability to import a csv of telemetry devices


AMEIJER1:
Currently duplicate records are just notified about, and the user has to come in and remove those duplicates. do we want the system to filter them out?

I also think I copied over the .test files from telemetry import when I created all the device import files and then I never modified them. So i will either have to make the tests work properly or I need to delete these totally 100% duplicate files.

api/src/paths/project/{projectId}/survey/{surveyId}/devices/import.ts
```diff
@@ -0,0 +1,142 @@
...
+import { csvFileSchema } from '../../../../../../openapi/schemas/file';
+import { authorizeRequestHandler } from '../../../../../../request-handlers/security/authorization';
+import { ImportDeviceService } from '../../../../../../services/import-services/devices/import-device-service';
+import { CSV_ERROR_MESSAGE } from '../../../../../../utils/csv-utils/csv-config-validation.interface';
+import { getLogger } from '../../../../../../utils/logger';
+import { parseMulterFile } from '../../../../../../utils/media/media-utils';
+import { getFileFromRequest } from '../../../../../../utils/request';
+import { constructXLSXWorkbook, getDefaultWorksheet } from '../../../../../../utils/xlsx-utils/worksheet-utils';
+
+const defaultLog = getLogger('/api/project/{projectId}/survey/{surveyId}/devices/upload');
```
mauberti-bc: This should match the file path (import instead of upload)

mauberti-bc:
Looks good! I'll test it more extensively on Monday but the code looks correct - nice work! The test files should basically be identical to the other csv import tests, they'll just call different functions and use different mock data. I can get the tests working if they're currently failing.

We should give users as few surprises as possible, so silently removing duplicates might be bad. Showing an error is probably better

api/src/paths/project/{projectId}/survey/{surveyId}/devices/import.ts
AMEIJER1: Fixed! In the telemetry import.ts I believe it uses upload - should I correct that in that import.ts file as well??

api/src/paths/project/{projectId}/survey/{surveyId}/devices/import.ts
```diff
@@ -0,0 +1,142 @@
...
+
+      if (errors.length) {
+        throw new HTTP422CSVValidationError(CSV_ERROR_MESSAGE, errors);
+      }
+
+      await connection.commit();
+
+      return res.status(200).send();
+    } catch (error) {
+      if (error instanceof HTTP422CSVValidationError === false) {
```
mauberti-bc: This isn't adding a log message if it's a 422 error, but we want to log all errors regardless of the status code. So just remove this conditional check

mauberti-bc (dismissed):
Looks great!

mauberti-bc (approved):

PR #1551, opened 2025-05-06
Hide Midnight Times in Animal Event Date Displays 

LouisThedroux:
## Description of Changes

This PR updates all animal event date/time displays (captures, releases, mortalities) to hide the time portion when it is exactly midnight (00:00 or 12:00 AM). The logic is now centralized using a shared utility function: shouldShowTime.

This affects:

Accordion headers for captures and mortalities
Details sections for captures, releases, and mortalities
Dropdowns for selecting capture/mortality events
Survey spatial animal mortality popups

A technical debt comment has been added to note that the current logic assumes 00:00 means "no time provided," which may not always be true if a user intentionally sets midnight. This should be revisited in the future to distinguish between a true null and a valid midnight time.


app/src/features/surveys/animals/profile/captures/components/AnimalCaptureCardContainer.tsx
```diff
@@ -160,10 +161,14 @@ export const AnimalCaptureCardContainer = (props: IAnimalCaptureCardContainer) =
                   }}>
                   <Stack gap={0.5} display="flex">
                     <Typography fontWeight={700}>
-                      {getFormattedDate(
-                        DATE_FORMAT.MediumDateTimeFormat,
-                        combineDateTime(capture.capture_date, capture.capture_time)
-                      )}
+                      {(() => {
```
mauberti-bc: We generally avoid this style of defining and immediately executing a function. You can define and use the function above, then just reference the memoized output here. Right now, the function gets defined and executes with every re-render, and it's a bit messier to read.

app/src/features/surveys/animals/profile/captures/components/AnimalCaptureCardContainer.tsx
```diff
@@ -160,10 +161,14 @@ export const AnimalCaptureCardContainer = (props: IAnimalCaptureCardContainer) =
...
+                        const dateTime = combineDateTime(capture.capture_date, capture.capture_time);
+                        const dateStr = getFormattedDate(DATE_FORMAT.MediumDateFormat, dateTime);
+                        const timeStr = dayjs(dateTime).format('HH:mm:ss');
+                        return shouldShowTime(timeStr)
+                          ? `${dateStr} ${getFormattedDate(DATE_FORMAT.TimeFormat, dateTime)}`
+                          : dateStr;
+                      })()}
```
mauberti-bc: We should simplify all of this into fewer utils, like a single displayDateTime(date, time).

The combineDateTime function is almost what we want: 

```
export const combineDateTime = (date: string, time?: string | null) => {
  if (date && time) {
    return dayjs(`${date} ${time}`).format(TIMESTAMP_FORMAT);
  }

  return dayjs(`${date}`).format(TIMESTAMP_FORMAT);
};
```

but the last line should return your new time format: 
```
return dayjs(`${date}`).format(DATE_FORMAT.TimeFormat);
```

Because the current combineDateTime is used to return a valid timestamp ('YYYY-MM-DDTHH:mm:ss'), you could just copy the logic to the new displayDateTime utility and change the last return statement as above

mauberti-bc (approved):

PR #1553, opened 2025-06-02
Collections delete

LouisThedroux:
## Links to Jira Tickets

Ability to delete collections.


api/src/paths/collection/{collectionId}/index.ts
```diff
@@ -307,3 +307,85 @@ export function createSubcollection(): RequestHandler {
...
+    return {
+      or: [
+        {
+          validSystemRoles: [SYSTEM_ROLE.SYSTEM_ADMIN, SYSTEM_ROLE.DATA_ADMINISTRATOR],
+          discriminator: 'SystemRole'
+        } ,
+              {
+                discriminator: 'CollectionRole',
+                collectionId: Number(req.params.collectionId),
+                validCollectionRoles: [COLLECTION_ROLE.ADMIN, COLLECTION_ROLE.MEMBER]
```
mauberti-bc: Collection members shouldn't be able to delete collections.

api/src/paths/collection/{collectionId}/index.ts
```diff
@@ -307,3 +307,85 @@ export function createSubcollection(): RequestHandler {
...
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+
+      const collectionId = Number(req.params.collectionId);
+
+      const systemUserId = connection.systemUserId();
+      const collectionService = new CollectionService(connection);
+      await collectionService.deleteCollection(collectionId, systemUserId);
```
mauberti-bc: systemUserId doesn't need to be passed because it doesn't do anything. The authorization check on collectionRole means that if the code reaches this point, we've already confirmed that the user making the request is an Admin of the given collection (or is a system/data admin).

api/src/paths/collection/{collectionId}/index.ts
LouisThedroux: of course!

app/src/features/collection/details/tags/SubcollectionContainer.tsx
```diff
@@ -211,9 +217,53 @@ export const SubcollectionContainer = (props: ICollectionsTagContainerProps) =>
...
 
+  // Add state for action menu
+  const [actionMenuAnchorEl, setActionMenuAnchorEl] = useState<{
+    anchorEl: MenuProps['anchorEl'];
+    collectionId: number;
+  } | null>(null);
+
+  const handleCloseActionMenu = () => setActionMenuAnchorEl(null);
+
+  // Add your edit and delete handlers here
```
mauberti-bc, line 246: We can use AI to generate code, but we should be reviewing every line. Removing comments like this that are obviously from AI is a good indicator that we have reviewed the code.

mauberti-bc:
Pulled into main collections branch

PR #1554, opened 2025-06-04
Collections users drew

LouisThedroux:
---
## Description
The purpose of the branch was to enable adding bulk members to collections.

This pull request was originally created as a quick fix for Annika to address critical survey page rendering issues in the `collections-users` branch. The primary intention was to resolve immediate problems, not to serve as a long-term solution or main development branch. However, as fixes accumulated, this branch was adopted as the main workstream for related survey and member management improvements. The branch will now be merged into `collections`, superseding its original purpose.

## Summary of Original Changes (June 4, 2025)

- Created an invite survey members page that calls forms to handle selecting surveys and members

- **Added POST Endpoint for Adding Members**  
  Introduced a dedicated endpoint to add survey members in bulk to more than one survey

- **Validation Update – Members and surveys Required**  
  Enforced “members” and "surveys" as required fields to resolve validation errors

- **Standalone Member Addition Logic**  **??**
  Refactored member addition so it no longer relies solely on a fetch method within the API call. This now targets the correct port (`7100` for app, not `6100`), improving reliability and avoiding cross-port issues.

## Next Steps
- Snackbar messages for when members cannot be added (if member is already in survey, api post request fails but no messages pop up). 
- There is no functionality to delete members from a survey. Not sure if the endpoint even exists right now and may need some degree of cascading logic to tidy things up (maybe not actually) 
- Need to add index.test for paths\survey\members\index.ts
- delete branch collections-users

---


AMEIJER1 (approved):

mauberti-bc:
Should this be merged into the collections branch or collections-users?

LouisThedroux:
@mauberti-bc 

> Should this be merged into the collections branch or collections-users?

collection-users. This was a branch I made to help Annika on the collection-users branch, which would then be pulled into collections.

api/src/paths/survey/{surveyId}/members/index.ts
```diff
@@ -132,3 +132,105 @@ export function getSurveyMembers(): RequestHandler {
...
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  },
+  responses: {
+    200: {
```
mauberti-bc, line 175: This implies that there is no response body for status 200, but your return (res.status(200).json(...) is indeed sending a response body.

api/src/paths/survey/{surveyId}/members/index.ts
```diff
@@ -132,3 +132,105 @@ export function getSurveyMembers(): RequestHandler {
...
+    }
+
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+      const surveyMemberService = new SurveyMemberService(connection);
+      await surveyMemberService.postSurveyMembers(surveyId, members);
+      await connection.commit();
+      return res.status(200).json({ message: 'Members added successfully' });
```
mauberti-bc: Replace .json(...) with .send()

AMEIJER1:
> Should this be merged into the collections branch or collections-users?

collections!

LouisThedroux:
Do we need the functionality where we can change the role of a user? It really doesn't add that much but in order to do it as of right now, user must remove someone, then re-add them under a new role. Which works fine, but feels like a bad standard to not have an edit button?

I am currently getting a broken/forbidden error when trying to view collection > Data. I need to look into it further

app/src/features/surveys/invite/InviteSurveyMembersPage.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export interface ITrimmedPayload {
+  selectedSurveys: number[];
+  selectedMembers: ITrimmedSurveyMember[];
+}
+
+const manageUsersYupSchema = yup.object().shape({
+  selectedSurveys: yup.array(yup.number()).min(1, 'You must select at least one survey.'),
+  selectedMembers: yup
+    .array(yup.object().shape({ system_user_id: yup.number(), collection_role_name: yup.string() }))
```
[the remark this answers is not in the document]
LouisThedroux: @AMEIJER1 Does this make sense? It does to me at face value. It works locally when changed, but I know the roles between collections, surveys and members has gotten a bit messy recently. Naming conventions are all over the place. Maybe we should have a clarifying session to make sure we are all on the same page

app/src/features/surveys/invite/form/member/InviteSurveyMembersMemberForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+    if (matchedRole && Object.values(SURVEY_ROLE).includes(matchedRole as SURVEY_ROLE)) {
+      const updated = [...values.selectedMembers];
+      updated[index] = {
+        ...updated[index],
+        survey_role_name: matchedRole as SURVEY_ROLE
+      };
+      setFieldValue('selectedMembers', updated);
+      clearErrors();
+    } else {
+      console.warn(`Invalid role selected: ${roleLabel}`);
```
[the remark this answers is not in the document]
LouisThedroux: @AMEIJER1 This feels like it makes sense. However, if the roles are selected from a predefined array of roles in a drop down, is this code ever necessary? When will the user be able to add someone with the wrong role if the roles are defined by us? Does that makes sense? We could implement these changes but I almost feel like it might be a little bit redundant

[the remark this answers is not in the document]
LouisThedroux: I guess its an extra layer of safety, having some error handling might be a good idea and having none is probably bad practice

api/src/paths/survey/members/index.ts
```diff
@@ -0,0 +1,127 @@
...
+import { paginationRequestQueryParamSchema } from '../../../openapi/schemas/pagination';
+import { CreateSurveyMemberSchema } from '../../../openapi/schemas/survey';
+import { authorizeRequestHandler } from '../../../request-handlers/security/authorization';
+import { SurveyMemberService } from '../../../services/survey-member-service';
+import { getLogger } from '../../../utils/logger';
+
+const defaultLog = getLogger('paths/survey/members/index');
+
+export const POST: Operation = [
+  authorizeRequestHandler((req) => {
```
LouisThedroux: _req

api/src/services/import-services/deployments/deployment-header-configs.ts
LouisThedroux, line 1: Empty file?

api/src/services/import-services/deployments/import-deployment-service.ts
LouisThedroux, line 1: empty file again?

app/src/hooks/api/useSurveyMemberApi.ts
```diff
@@ -20,6 +21,17 @@ const useSurveyMemberApi = (axios: AxiosInstance) => {
     return data;
   };
 
+  /**
+   * Add users to multiple surveys
+   *
+   * @param {IManageUsersFormValues} values
+   * @return {*}  {Promise<void>}
+   */
+  const addBulkSurveysMembers = async (values: ITrimmedPayload): Promise<void> => {
```
LouisThedroux: Should this be Promise<any> instead of void? As it is expecting something in return?

api/src/services/survey-member-service.ts
```diff
@@ -94,6 +94,30 @@ export class SurveyMemberService extends DBService {
...
 
+  /**
+   * Adds multiple survey participants to any number of surveys (bulk permissions).
+   *
+   * @param {number[]} surveyIds
+   * @param {IInsertSurveyMember[]} members
+   * @return {*}  {Promise<void[]>}
+   * @memberof SurveyMemberService
+   */
+  async insertMemberstoSurveys(surveyIds: number[], members: IPostSurveyMember[]): Promise<void> {
```
LouisThedroux: insertMembersToSurveys for proper camel casing

app/src/features/surveys/invite/constants/roles
LouisThedroux, line 1: Empty?

api/src/repositories/survey-member-repository.ts
```diff
@@ -337,27 +337,43 @@ export class SurveyMemberRepository extends BaseRepository {
...
       `;
+    }
 
     const response = await this.connection.sql(sqlStatement);
 
-    if (!response.rowCount) {
+    if (!response || !response.rowCount) {
       throw new ApiExecuteSQLError('Failed to insert survey team member', [
         'SurveyRepository->postSurveyMember',
         'rows was null or undefined, expected rows != null'
```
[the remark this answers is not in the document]
AMEIJER1: I think we are fine not to change this because there would not be a situation where there are no affected rows. The form is handling inserting survey members. if no survey members are inserted then that is a problem. I hope i'm understanding that right?

app/src/features/surveys/invite/form/member/InviteSurveyMembersMemberForm.tsx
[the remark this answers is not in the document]
AMEIJER1: removing console makes sense!! thanks for catching that

api/src/services/import-services/deployments/deployment-header-configs.ts
AMEIJER1: I must have switched branches from the bulk import deployments one I was working on and accidentally added those file names into this repository. Can delete safely.

api/src/paths/survey/members/index.ts
```diff
@@ -0,0 +1,127 @@
...
+          validSystemRoles: [SYSTEM_ROLE.PROJECT_CREATOR],
+          discriminator: 'SystemRole'
+        },
+        {
+          validSystemRoles: [SYSTEM_ROLE.DATA_ADMINISTRATOR],
+          discriminator: 'SystemRole'
+        },
+        {
+          validSystemRoles: [SYSTEM_ROLE.SYSTEM_ADMIN],
+          discriminator: 'SystemRole'
```
mauberti-bc, line 20: The ```or``` array in authorizeRequestHandler should not have duplicate discriminators, so you should have an array with 1 element and the validSystemRoles array in that object should have 3 elements:
```
{
 validSystemRoles: [SYSTEM_ROLE.PROJECT_CREATOR, SYSTEM_ROLE.DATA_ADMINISTRATOR, SYSTEM_ROLE.SYSTEM_ADMIN],
discriminator: 'SystemRole'
}
```

For each item in the ```or``` array, the middleware has to await a database query, so the current code will be relatively slow.

api/src/paths/survey/members/index.ts
```diff
@@ -0,0 +1,127 @@
...
+export function addMembersToSurveys(): RequestHandler {
+  return async (req, res) => {
+    defaultLog.debug({ label: 'addMembersToSurveys' });
+
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+
+      const surveyIds = req.body.selectedSurveys;
```
mauberti-bc: Data on the request body (```req.body.selectedSurveys``` in this case) has type ```any```, which we should avoid as much as possible. When accessing data from the request body, you can use ```req.body.selectedSurveys as type```, replacing type with your real type. Since selectedSurveys is an array of numbers, you can do ```req.body.selectedSurveys as number[]```, like you did below for selectedMembers.

api/src/paths/survey/{surveyId}/members/index.ts
```diff
@@ -119,11 +119,21 @@ export function getSurveyMembers(): RequestHandler {
...
     return {
-      and: [
+      or: [
+        {
+          validSurveyRoles: [SURVEY_ROLE.ADMIN],
+          surveyId: Number(req.params.surveyId),
+          discriminator: 'SurveyRole'
+        },
+        {
+          validSystemRoles: [SYSTEM_ROLE.DATA_ADMINISTRATOR],
```
mauberti-bc: Same comment as above: put all SystemRole validation in a single object.

api/src/repositories/survey-member-repository.ts
```diff
@@ -338,27 +338,43 @@ export class SurveyMemberRepository extends BaseRepository {
   async insertSurveyMember(surveyId: number, systemUserId: number, surveyMemberRole: number | string): Promise<void> {
```
mauberti-bc: surveyMemberRole should not be ```number | string```, because you guaranteed that it's a string in the openapi spec request validation.

api/src/repositories/survey-member-repository.ts
```diff
@@ -338,27 +338,43 @@ export class SurveyMemberRepository extends BaseRepository {
...
     let sqlStatement;
 
-    sqlStatement = SQL`
+    // If surveyMemberRole is a string (role name), look up the ID in the insert (case-insensitive)
+    if (isNaN(Number(surveyMemberRole))) {
```
mauberti-bc: You can remove this whole if check once changing surveyMemberRole to be type string.

api/src/repositories/survey-member-repository.ts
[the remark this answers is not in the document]
mauberti-bc: I agree, we can ignore copilot here. 

The check is designed to catch issues related to inserting rows, and because it uses the ```throw``` keyword, the database operations will be rolled back upon failure (which we want). There shouldn't be a situation where no rows are inserted.

api/src/repositories/survey-member-repository.ts
```diff
@@ -469,4 +485,67 @@ export class SurveyMemberRepository extends BaseRepository {
...
+   *
+   * @param {{
+   *   survey_id: number;
+   *   system_user_id: number;
+   *   survey_role_name: string;
+   * }[]} members
+   * @return {*} Promise<void>
+   * @memberof SurveyMemberRepository
+   */
+  async insertMembersBatch(
```
mauberti-bc: You should try to match the descriptive naming style that we use elsewhere, like insertMultipleSurveyMembers instead of insertMembersBatch

api/src/repositories/survey-member-repository.ts
```diff
@@ -469,4 +485,67 @@ export class SurveyMemberRepository extends BaseRepository {
...
+    members: {
+      survey_id: number;
+      system_user_id: number;
+      survey_role_name: string;
+    }[]
+  ): Promise<void> {
+    if (!members.length) {
```
mauberti-bc, line 490: You should rewrite this function to be more explicit. It's short and clever but hard to read = hard to maintain. I find that AI loves to create very confusing SQL unless you give an example of the preferred style.

This is untested, but something like this:

```
const surveyIds = members.map(m => m.survey_id);
  const systemUserIds = members.map(m => m.system_user_id);
  const roleNames = members.map(m => m.survey_role_name);

  const sql = SQL`
    WITH input_data AS (
      SELECT
        unnest(${surveyIds}::int[]) AS survey_id,
        unnest(${systemUserIds}::int[]) AS system_user_id,
        unnest(${roleNames}::text[]) AS survey_role_name
    )
    INSERT INTO survey_member (
      survey_id,
      system_user_id,
      survey_role_id
    )
    SELECT
      i.survey_id,
      i.system_user_id,
      sr.survey_role_id
    FROM input_data i
    JOIN survey_role sr
      ON LOWER(sr.name) = LOWER(i.survey_role_name)
    WHERE sr.record_end_date IS NULL
    RETURNING *;
  `;
```

api/src/paths/survey/members/index.ts
mauberti-bc: You can also just exclude all params if none are used

mauberti-bc:
Also make sure to check the sonarcloud issues when reviewing PRs
image

api/src/paths/collection/{collectionId}/member/index.ts
```diff
@@ -256,3 +256,78 @@ export function addMembersToCollection(): RequestHandler {
...
+  security: [{ Bearer: [] }],
+  parameters: [
+    {
+      in: 'path',
+      name: 'collectionId',
+      schema: { type: 'integer', minimum: 1 },
+      required: true
+    },
+    {
+      in: 'query',
```
mauberti-bc, line 287: This doesn't appear in the file name so this will currently fail

mauberti-bc: This should be in the path, not query:

collection/{collectionId}/member/{collectionMemberId}/index

PR #1563, opened 2025-07-09
SIMSBIOHUB-715: Collection links - Functionality for user to add external resources to their collections

LouisThedroux:
## Links to Jira Tickets

[SIMSBIOHUB-715](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-715)

---

## Description of Changes

Implements collection links management functionality, allowing users to create, view, update, and delete external resource links associated with a collections. Allowing users to have a central repo of all appropriate external links related to their collection/project in one easy accessible place.  

### Backend
- Introduced a PUT API endpoint at `/collection/{collectionId}/links/{linkId}` to handle both updates and soft deletion of collection links (using a `record_end_date` for soft delete).
- Developed `CollectionLinkRepository` and enhanced `CollectionLinkService` to support full CRUD operations, including soft delete logic.

### Frontend
- Added `CollectionLinkDialog` for creating and editing links, with appropriate loading states.
- Built `CollectionLinkForm` using Yup for validating name, description, and URL fields.
- Updated API hooks for collections to support link update and soft delete.

### Features
- Create, update, and soft delete (via PUT) collection links
- Form validation with error handling
- Loading states and user feedback
- Authorization enforced for sensitive operations

### TODO/Tech Debt
- Add comprehensive unit tests for all components
- Remove any seed files before merging
- Consider multi-delete functionality for links
- Fix rendering issues with delete operations
- Evaluate displaying links directly on collection details page

---

## Testing Notes

- Test creating, updating, and soft deleting collection links via the UI.
- Confirm validation, loading states, and user feedback are working as expected.
- Ensure soft deletes are reflected in both UI and database (`record_end_date` is set).
- Verify only authorized users can manage links where appropriate.
- Ensure seed files are removed before merge.
- Report any link rendering or feedback issues.

**Known Issue/Consideration:**  
Annika is experiencing 500 response errors from this code, which cannot be reproduced in my environment. The cause is currently unknown and requires further investigation.


api/src/models/collection-link.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ApiPaginationResults } from '../zod-schema/pagination';
+
+export const CollectionLink = z.object({
+  collection_links_id: z.number(),
+  name: z.string(),
+  description: z.string().nullable(),
+  url: z.string(),
+  collection_id: z.number(),
+  record_end_date: z.string().nullable().optional(),
+  create_date: z.string(),
```
mauberti-bc: These audit columns shouldn't be in zod schema/sent to the client. We started adding separate zod schemas in the database-models/ folder to enable zod validation for queries that simply return the record: ```returning *```, in which case those schemas _do_ have the create_date, create_user, etc. But this CollectionLink zod object should match the response of the GET collection-link request, and that response won't have any audit info.

api/src/models/collection-link.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface IPostCollectionLinkRequest {
+  name: string;
+  description?: string;
+  url: string;
+}
+
+export interface IPutCollectionLinkRequest {
+  name: string;
+  description?: string;
```
mauberti-bc: Making description nullable (```description: string | null```) in the update payload might help because when a user edits a collection link, the form will be populated with the name, description, and url. If description is null in the database record, description will be added to the form's initial values as null. As is, unless you removed the description property from your update request body if it's null, you'll likely get an error in the PUT request saying description was expected to be a string because it is being received as null.

If you make description nullable in this interface, you'll also need to make it nullable in the PUT request's openapi schema.

api/src/openapi/schemas/collection-link.ts
```diff
@@ -0,0 +1,103 @@
+import { OpenAPIV3 } from 'openapi-types';
+
+export const CollectionLinkSchema: OpenAPIV3.SchemaObject = {
+  title: 'Collection Link',
+  type: 'object',
+  additionalProperties: false,
+  required: ['id', 'name', 'url', 'collection_id', 'create_date', 'create_user'],
+  properties: {
+    id: {
```
mauberti-bc: We shouldn't use "id". We should always refer to the "[table name]_id" (ie. collection_link_id) - the primary key of the table should be called this and it should always be referenced like this.

api/src/paths/collection/{collectionId}/links/{linkId}/index.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    try {
+      await connection.open();
+
+      const collectionId = Number(req.params.collectionId);
+      const linkId = Number(req.params.linkId);
+      const collectionLinkService = new CollectionLinkService(connection);
+
+      // Check if this is an end operation (has record_end_date field)
+      if ('record_end_date' in req.body) {
```
mauberti-bc: Logic like this (if X, do Y) should live in the service layer, not the endpoint.

app/src/features/collection/details/links/CollectionLinkContainer.tsx
```diff
@@ -0,0 +1,344 @@
...
+      'with pagination:',
+      paginationSort
+    );
+    collectionLinksDataLoader.refresh(paginationSort);
+  }, [paginationSort]);
+
+  const rows = collectionLinksDataLoader.data?.links ?? [];
+
+  // Debug logging
+  console.log('CollectionLinkContainer: Data loader state:', {
```
mauberti-bc: Remove all console.logs once you're done debugging

api/src/repositories/collection-link-repository.ts
```diff
@@ -0,0 +1,257 @@
...
+   * @memberof CollectionLinkRepository
+   */
+  async getCollectionLinks(collectionId: number, pagination?: ApiPaginationOptions): Promise<CollectionLink[]> {
+    let sqlStatement;
+
+    if (pagination) {
+      const limit = pagination.limit;
+      const offset = (pagination.page - 1) * pagination.limit;
+
+      sqlStatement = SQL`
```
mauberti-bc, line 31: 1. You could simply this code to not repeat yourself, like this:
```
const sqlStatement = SQL`
    SELECT 
      cl.id as collection_links_id,
      cl.name,
      cl.description,
      cl.url,
      cl.collection_id,
      cl.record_end_date,
      cl.create_date,
      cl.create_user
    FROM 
      collection_links cl
    WHERE 
      cl.collection_id = ${collectionId}
      AND cl.record_end_date IS NULL
    ORDER BY cl.create_date DESC
  `;

  if (pagination) {
    sqlStatement.append(SQL` LIMIT ${pagination.limit} OFFSET ${(pagination.page - 1) * pagination.limit}`);
  }

  const response = await this.connection.sql(sqlStatement, CollectionLink);
```
2. The pagination options include the order (desc or asc), so you shouldn't hard code the order by: 
```
ORDER BY ${pagination.sort} ${pagination.order}
```
3. Better yet, use knex to match how we conditionally apply pagination else. Here's an example from survey-repository.ts:
```
async findSurveys(
    isUserAdmin: boolean,
    systemUserId: number | null,
    filterFields: ISurveyAdvancedFilters,
    pagination?: ApiPaginationOptions
  ): Promise<FindSurveysResponse[]> {
    const query = this._makeFindSurveysQuery(isUserAdmin, systemUserId, filterFields);

    // Pagination
    if (pagination) {
      query.limit(pagination.limit).offset((pagination.page - 1) * pagination.limit);

      if (pagination.sort && pagination.order) {
        query.orderBy(pagination.sort, pagination.order);
      }
    }

    const response = await this.connection.knex(query, FindSurveysResponse);

    return response.rows;
  }
```

api/src/repositories/collection-link-repository.ts
```diff
@@ -0,0 +1,257 @@
...
+        description,
+        url,
+        collection_id,
+        create_user
+      ) VALUES (
+        ${linkData.name},
+        ${linkData.description || null},
+        ${linkData.url},
+        ${collectionId},
+        ${this.connection.systemUserId()}
```
mauberti-bc: You shouldn't set the create_user like this. There are triggers that automatically update the audit columns (created_at, created_by, etc.) based on which user is connected to the database.

app/src/features/collection/details/links/CollectionLinkContainer.tsx
```diff
@@ -0,0 +1,344 @@
...
+    page: paginationModel.page + 1 // API pagination pages begin at 1, but MUI DataGrid pagination begins at 0.
+  };
+
+  const collectionLinksDataLoader = useDataLoader((pagination: ApiPaginationRequestOptions) =>
+    biohubApi.collection.getCollectionLinks(collectionId, pagination)
+  );
+
+  // Fetch collection links when pagination or sort changes
+  useDeepCompareEffect(() => {
+    console.log(
```
mauberti-bc: 👀

app/src/features/collection/details/links/CollectionLinkContainer.tsx
```diff
@@ -0,0 +1,344 @@
...
+ *
+ * @return {*}
+ */
+export const CollectionLinkContainer = (props: ICollectionLinkContainerProps) => {
+  const { collectionId, showSearch = false } = props;
+
+  const biohubApi = useBiohubApi();
+
+  const { searchParams, setSearchParams } = useSearchParams<StringValues<CollectionLinkDataTableURLParams>>();
+  const [linkDialogIsOpen, setLinkDialogIsOpen] = useState(false);
```
mauberti-bc: Do you need both these useStates? Or could you just infer that the dialog is open when editingLink !== null?

app/src/features/collection/details/links/CollectionLinkContainer.tsx
```diff
@@ -0,0 +1,344 @@
...
+const initialPaginationParams: Required<ApiPaginationRequestOptions> = {
+  page: 0,
+  limit: 10,
+  sort: 'id',
+  order: 'desc'
+};
+
+/**
+ * Displays a list of collection links.
+ *
```
mauberti-bc: Make sure that all your functions have complete jsdocs (description, params, return at the very least)

app/src/features/collection/details/links/CollectionLinkContainer.tsx
```diff
@@ -0,0 +1,344 @@
...
+    }
+  ];
+
+  return (
+    <>
+      <Toolbar style={{ display: 'flex', justifyContent: 'space-between' }}>
+        <Typography variant="h4" component="h2">
+          External Resources &zwnj;
+          <Typography component="span" color="textSecondary" lineHeight="inherit" fontSize="inherit" fontWeight={400}>
+            ({Number(collectionLinksDataLoader.data?.pagination?.total ?? 0).toLocaleString()})
```
mauberti-bc: data.pagination.total should never be a string, so you should make sure the database returns it as a number and that it stays a number throughout the request.

app/src/features/collection/details/links/CollectionLinkContainer.tsx
```diff
@@ -0,0 +1,344 @@
...
+          hasNoDataFallback={
+            <NoDataOverlay
+              minHeight="400px"
+              title="Create External Resources"
+              subtitle={`There are no external resources. When you create one, it will appear here.`}
+              icon={mdiLink}
+            />
+          }
+          hasNoDataFallbackDelay={100}>
+          <StyledDataGrid
```
mauberti-bc: You should put the data grid in its own component, as a child of CollectionLinkContainer. Think of the table as just being a shell, then the parent component passes down the content to fill the shell. WIth that design, the parent (CollectionLinkContainer) makes the API request and passes dataLoader.data.links to CollectionLinkTable. 

Another benefit of that design is that once you're in CollectionLinkTable, you know that the data is defined because the parent will check that dataLoader.data.links is defined before rendering CollectionLinkTable.

app/src/features/collection/details/links/dialog/CollectionLinkDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+interface ICollectionLinkDialogProps {
+  collectionId: number;
+  link?: ICollectionLink | null;
+  onSubmit: () => void;
+  onClose: () => void;
+  open: boolean;
+}
+
+/**
+ * Dialog for creating or editing collection links
```
mauberti-bc: Complete the jsdoc

app/src/features/collection/details/links/dialog/CollectionLinkDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+          name: formData.name,
+          description: formData.description,
+          url: formData.url
+        };
+        await biohubApi.collection.createCollectionLink(collectionId, createData);
+      }
+
+      onSubmit();
+    } catch (error) {
+      console.error('Error saving collection link:', error);
```
mauberti-bc: Remove console.errors. I'd suggest using the dialogContext to show a snackbar upon failure. You can see an example of that in CreateAlert.tsx

app/src/features/collection/details/links/dialog/CollectionLinkDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  const handleClose = () => {
+    if (!isLoading) {
```
mauberti-bc: Is this !isLoading check necessary? Patterns like this open the door for issues because if the API request hangs or some other bug causes isLoading to stay falsey, you won't be able to close the dialog without refreshing the browser.

app/src/features/collection/details/links/dialog/CollectionLinkDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+  ICollectionLink,
+  ICreateCollectionLinkRequest,
+  IUpdateCollectionLinkRequest
+} from 'interfaces/useCollectionApi.interface';
+import { ReactNode, useState } from 'react';
+import CollectionLinkForm, { ICollectionLinkFormData } from './form/CollectionLinkForm';
+
+interface ICollectionLinkDialogProps {
+  collectionId: number;
+  link?: ICollectionLink | null;
```
mauberti-bc: Whenever you allow 3 different types like this: undefined, null, or the interface, it's good to pause and ask why. In general, allowed types should be as specific as possible. If the parent is trying to pass an undefined link, it should be the parent's responsibility to fall back to null. So I would change this to:
```
link: ICollectionLInk | null
```

app/src/features/collection/details/links/dialog/CollectionLinkDialog.tsx
mauberti-bc, line 1: I'd suggest creating separate create and edit components. Check out CreateAlert.tsx and EditAlert.tsx and follow that structure exactly (everything should be the same, just different data being sent and collected). 

Separate components is cleaner because
- Validation is going to be different between create and edit (eg. edit alert requires collection_link_id to be defined while the create dialog doesn't)
- Initial values are going to be different
- The edit dialog will include an API request while the create dialog wont
- You avoid redundant if statements, tightening the code

app/src/features/collection/details/links/dialog/CollectionLinkDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+  };
+
+  const handleClose = () => {
+    if (!isLoading) {
+      onClose();
+    }
+  };
+
+  return (
+    <Dialog fullWidth maxWidth="sm" open={open} onClose={handleClose} aria-labelledby="collection-link-dialog-title">
```
mauberti-bc: We have a bunch of existing dialog components that integrate with Formik. You should use EditDialog.tsx here - look for examples of where it's used.

app/src/features/collection/details/links/dialog/form/CollectionLinkForm.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { Formik, FormikProps } from 'formik';
+import { ReactNode } from 'react';
+import * as yup from 'yup';
+
+//TECH DEBT:
+//TO-DO:
+// - ENSURE ALL ASPECTS OF THIS FORM AND OTHER FILES ARE COVERED BY UNIT TESTS
+// - REMOVE SEED FILES BEFORE MERGING ANYWHERE
+// - DO WE NEED THE ABILITY TO MULTI-DELETE LINKS?
+// - DELETING IS A BIT BROKEN AND DOESNT RENDER PROPERLY? this might just be the docker status issue
```
mauberti-bc: If the page loads, it's very unlikely to do with Docker.

app/src/features/collection/details/links/dialog/form/CollectionLinkForm.tsx
```diff
@@ -0,0 +1,104 @@
...
+export interface ICollectionLinkFormData {
+  name: string;
+  description: string;
+  url: string;
+}
+
+export interface ICollectionLinkFormProps {
+  initialValues: ICollectionLinkFormData;
+  onSubmit: (values: ICollectionLinkFormData) => void;
+  renderForm: (formikProps: FormikProps<ICollectionLinkFormData> & { children: ReactNode }) => ReactNode;
```
mauberti-bc: You can specify children using React's native PropsWithChildren, like this:
```
const myFunction = (props: PropsWithChildren<myCustomInterface>) => {}
```

You should be able to avoid writing {children: ReactNode} if the children are indeed just nested DOM elements of the component you're passing props to

app/src/features/collection/details/links/dialog/form/CollectionLinkForm.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <Formik
+      initialValues={initialValues}
+      validationSchema={CollectionLinkFormValidationSchema}
+      validateOnBlur={true}
+      validateOnChange={false}
+      onSubmit={onSubmit}>
+      {(formikProps) => {
+        const { values, touched, errors, handleChange, handleBlur } = formikProps;
+
+        const formContent = (
```
mauberti-bc: We have custom form components that handle the Formik state being updated from input and select components. CustomTextField, CustomAutocomplete, etc. But if you use EditDialog.tsx, you won't even use <Formik/> because the dialog will handle that internally.

app/src/interfaces/useCollectionApi.interface.ts
```diff
@@ -82,3 +82,35 @@ export interface IPostCollectionMember {
...
+
+export interface ICollectionLink {
+  collection_links_id: number;
+  name: string;
+  description: string | null;
+  url: string;
+  collection_id: number;
+  record_end_date: string | null;
+  create_date: string;
+  create_user: number;
```
mauberti-bc: We shouldn't pass audit columns to the frontend unless we display them, which we shouldn't be doing in this case. So you can remove create_date, create_user from all the types and make sure you're not selecting them in the database query.

database/src/migrations/20250703101530_collection_links.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Table for user-specific survey filters
+ *
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set search_path=biohub;
+
```
mauberti-bc: Take a look at the 20240927000000_alert_table.ts as an example of a complete migration. 
* Use singular words (ie. collection_link instead of collection_links)
* Postgres automatically indexes primary keys, but you'll need to manually create an index on all foreign keys
* Add foreign key constraints to all foreign keys (ie. collection_id)
* Add audit triggers to the table so that the create_date and create_user are automatically set
* Add descriptive comments to each column
* Prevent 2 collection links from having the same name (ie. unique constraints on (name, collection_id))

All of those bullet points are addressed in the alert_table migration, so use that migration as a template.

*NOTE*: We no longer create views for every table, so you don't need to create a view in the biohub_dapi_v1 schema like the alert migration does.

database/src/seeds/05_collection_links.ts
```diff
@@ -0,0 +1,66 @@
...
+      description: 'Best hardware store on the planet',
+      url: 'https://www.canadiantire.ca',
+      collection_id: collectionId,
+      create_date: knex.fn.now(),
+      create_user: 8
+    },
+    {
+      name: 'Costco',
+      description: 'Cheap hotdogs',
+      url: 'https://www.costco.ca',
```
mauberti-bc: The seed links should all be BC gov links, like https://www2.gov.bc.ca/gov/content/home.

database/src/seeds/05_collection_links.ts
```diff
@@ -0,0 +1,66 @@
...
+      create_user: 8
+    }
+  ]);
+
+  // Get user IDs for Annika and Macgregor
+  const [annikaUser] = await knex('system_user')
+    .select('system_user_id')
+    .where('user_identifier', 'ameijer');
+    
+  const [macgregorUser] = await knex('system_user')
```
mauberti-bc: The seed data shouldn't mention any name

api/src/openapi/schemas/collection-link.ts
```diff
@@ -0,0 +1,117 @@
+import { OpenAPIV3 } from 'openapi-types';
+
+export const CollectionLinkSchema: OpenAPIV3.SchemaObject = {
+  title: 'Collection Link',
+  type: 'object',
+  additionalProperties: false,
+  required: ['collection_link_id', 'name', 'url', 'collection_id', 'create_date', 'create_user'],
```
mauberti-bc: Don't send audit columns like create_date, create_user to the frontend unless that information is used on the frontend. These are just for keeping track of where data came from.

api/src/paths/collection/{collectionId}/link/{linkId}/index.ts
```diff
@@ -0,0 +1,139 @@
...
+      name: 'collectionId',
+      required: true,
+      schema: {
+        type: 'integer',
+        minimum: 1
+      }
+    },
+    {
+      in: 'path',
+      name: 'linkId',
```
mauberti-bc, line 52: Use the exact name of the parameter as it appears in the database, so ```collectionLinkId``` instead of linkId. And rename the file to match.

api/src/paths/collection/{collectionId}/link/{linkId}/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  ],
+  requestBody: {
+    description: 'Collection link update or end request object.',
+    required: true,
+    content: {
+      'application/json': {
+        schema: {
+          oneOf: [
+            UpdateCollectionLinkSchema,
+            EndCollectionLinkSchema
```
mauberti-bc: This style of multiple schemas might get messy and isn't something we've done before. Instead, the frontend should just send the request to match the UpdateCollectionLinkSchema, and that update schema should have everything optional - making it possible to just include record_end_date.

api/src/paths/collection/{collectionId}/link/{linkId}/index.ts
```diff
@@ -0,0 +1,139 @@
...
+
+      const collectionId = Number(req.params.collectionId);
+      const linkId = Number(req.params.linkId);
+      const collectionLinkService = new CollectionLinkService(connection);
+
+      // Check if this is an end operation (has record_end_date field)
+      if ('record_end_date' in req.body) {
+        defaultLog.debug({ label: 'endCollectionLink' });
+        const data = req.body as IEndCollectionLinkRequest;
+        const endedLink = await collectionLinkService.endCollectionLink(collectionId, linkId, data.record_end_date);
```
mauberti-bc: This code can be simplified to just calling updateCollectionLink with only record_end_date.

api/src/paths/collection/{collectionId}/link/{linkId}/index.ts
```diff
@@ -0,0 +1,139 @@
...
+    try {
+      await connection.open();
+
+      const collectionId = Number(req.params.collectionId);
+      const linkId = Number(req.params.linkId);
+      const collectionLinkService = new CollectionLinkService(connection);
+
+      // Check if this is an end operation (has record_end_date field)
+      if ('record_end_date' in req.body) {
+        defaultLog.debug({ label: 'endCollectionLink' });
```
mauberti-bc: We don't have a consistent strategy for where to place these loggers, but we should be very intentional when adding them to avoid massive logs. If this request fails, the logger on line 132 will print out the error. I think we can remove this debug logger.

api/src/services/collection-link-service.ts
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * End a collection link by setting record_end_date.
+   *
+   * @param {number} collectionId
+   * @param {number} linkId
+   * @param {string} recordEndDate
+   * @return {*}  {Promise<CollectionLink>}
+   * @memberof CollectionLinkService
+   */
+  async endCollectionLink(collectionId: number, linkId: number, recordEndDate: string): Promise<CollectionLink> {
```
mauberti-bc: Consider removing this method and just using update.

app/src/features/collection/details/link/CollectionLinkContainer.tsx
```diff
@@ -0,0 +1,361 @@
...
+    setEditingLink(link);
+    setLinkDialogIsOpen(true);
+  };
+
+  const handleDeleteLink = async (linkId: number) => {
+    try {
+      await biohubApi.collection.endCollectionLink(collectionId, linkId);
+      collectionLinksDataLoader.refresh(paginationSort);
+    } catch (error) {
+      console.error('Error ending collection link:', error);
```
mauberti-bc: Avoid console.error statements in the frontend. Instead, print part of the error in a snackbar (lots of examples if you search the code)

LouisThedroux:
As of today. All comments have been addressed, sonaqube issues should be 0. the only thing that is left to do is test coverage.

PR #1568, opened 2025-07-24
Adding sample site period / visit comment capability and ARU sampling method

AMEIJER1:
## Description of Changes

- adding ARU as a sampling method 
- Adding comments to sampling_period_visit 
- changing create sampling period page to allow for submission of comments 
- configured bulk upload capability to include comments

## Testing Notes

-  handlesubmit on my createsampleperiodpage is using a for loop to submit one period at a time. should we update the page and index file so that it pushes them all up at one instead of iterating?

### DIFF
--- a/database/src/migrations/20250721000001_sampling_period_comment.ts
+++ b/database/src/migrations/20250721000001_sampling_period_comment.ts
@@ -24,7 +24,7 @@ export async function up(knex: Knex): Promise<void> {
     --------------------------------------------------------------------------------------------------
 
     ALTER TABLE survey_sample_period
-    ADD COLUMN comment TEXT;
+    ADD COLUMN comment VARCHAR(1000);
   `);
 }
 
--- a/database/src/migrations/20250721000001_sampling_period_comment.ts
+++ b/database/src/migrations/20250721000001_sampling_period_comment.ts
@@ -1,0 +1,33 @@
+import { Knex } from 'knex';
+
+/**
+ * - Adds ARUs as a sampling method
+ * - Adds a column to the survey_sample_period table to store comments
+ *
+ * @export
+ * @param {Knex} knex
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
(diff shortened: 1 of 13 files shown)


database/src/migrations/20250721000001_sampling_period_comment.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    INSERT INTO method_lookup (name, description, record_effective_date)
+    VALUES ('Acoustic Recording Unit', 'Detecting species using an Acoustic Recording Unit (ARU)', now());
+
+    --------------------------------------------------------------------------------------------------
+    --------------------- Insert column for comment in survey_sample_period --------------------------
+    --------------------------------------------------------------------------------------------------
+
+    ALTER TABLE survey_sample_period
+    ADD COLUMN comment TEXT;
```
mauberti-bc: It's good to limit the length of comment columns, so instead of TEXT you can do something like VARCHAR(1000).

LouisThedroux:
@AMEIJER1 

So now we have a comment column on display in the sampling period table on the manage sampling information page. However, I dont see anyway of adding a comment to an existing or new sampling period. Is this something I am missing, or does it need to be added? I guess it can still be added via importing a csv with a comment column? Do we need a way to add manual comments when directly adding or editing comments in the UI?

database/src/migrations/20250721000001_sampling_period_comment.ts
LouisThedroux: This is done (Y)

AMEIJER1:
> @AMEIJER1
> 
> So now we have a comment column on display in the sampling period table on the manage sampling information page. However, I dont see anyway of adding a comment to an existing or new sampling period. Is this something I am missing, or does it need to be added? I guess it can still be added via importing a csv with a comment column? Do we need a way to add manual comments when directly adding or editing comments in the UI?

Hi!! Thanks for reviewing it - I just built web and was able to edit the existing sampling period seed data in the UI by editing the sampling period through that vertical ellipses on the sampling period table beside each record. It will pull up the create sampling period form in which they can add a comment. @LouisThedroux

PR #1570, opened 2025-08-06
Adding inactive or active tabs to devices container

AMEIJER1:
## Description of Changes

- modifying devices container to create helper functions determining whether devices are currently engaged in deployments
- removing the "deployed on" animal column from the devices table 

## Testing Notes

- no data overlay is formatted the same in devicescontainer as it is in deployments container, but deployments container has the grey that stretches to the bottom of the paper on the survey tabs page, whereas devices does not stretch to the bottom on the survey view of the devices table. cannot figure out how to extend this. Also, modifying height seems to not be the solution since on the telemetry/manage page we need to have some min heights for those tables so that the tables aren't totally collapsed if they have no data.


.vscode/tasks.json
mauberti-bc, line 1: You should add .vscode to your .gitignore

.gitignore
LouisThedroux, line 1: Still not sure why this is in here, but I dont know what to do or if its a problem

LouisThedroux (approved):
I have gone over this and made a bunch of changes. It is working as it should now with some tweaks to styling etc.

LouisThedroux:
Code duplication failed. Will go over it again and try to refine where possible.

LouisThedroux (approved):
Gone over and changed a few things but is working and looks better now.

PR #1573, opened 2025-08-27
Simsbiohub 728 - Husky pre-commit hooks 

LouisThedroux:
## Links to Jira Ticket:
[SIMSBIOHUB-728 
](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-728)

## Description of Changes
This PR introduces Husky Git hooks to automatically enforce code quality and run tests before code is committed or pushed.

Pre-commit: Lints and formats all staged files automatically.

Pre-push: Runs tests for api and app packages, blocking pushes if tests fail.

Pre-commit hooks (lint-fix, format-fix) run only on relevant files to save time.

Pre-push hook (tests) runs on everything (not db as it has no tests)

## Notes
No changes to existing package scripts; uses current lint-fix, format-fix, and test scripts.

Everyone should run npm install after pulling this PR to activate Husky hooks (prepare script is configured).

Also need a second opinion on  whether lint-staged package is justified as a dependency, I think it is but there could be a way to add this functionality directly without the extra dependency?

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -8,7 +8,8 @@
 
     <meta name="color-scheme" content="light dark" />
     <meta name="viewport" content="viewport-fit=cover, width=device-width, initial-scale=1.0, minimum-scale=1.0" />
-    <meta name=" format-detection" content="telephone=no" />
+    <meta name=" format-detection" 
+    content="telephone=no" />
     <meta name="msapplication-tap-highlight" content="no" />
 
     <link rel="manifest" href="/manifest.json" />
--- a/app/index.html
+++ b/app/index.html
@@ -8,8 +8,7 @@
 
     <meta name="color-scheme" content="light dark" />
     <meta name="viewport" content="viewport-fit=cover, width=device-width, initial-scale=1.0, minimum-scale=1.0" />
-    <meta name=" format-detection" 
-    content="telephone=no" />
+    <meta name=" format-detection" content="telephone=no" />
     <meta name="msapplication-tap-highlight" content="no" />
 
     <link rel="manifest" href="/manifest.json" />
--- a/app/src/features/support/SupportPage.tsx
+++ b/app/src/features/support/SupportPage.tsx
@@ -13,8 +13,6 @@ import {
   SupportPageViews
 } from './views/SupportPageView';
 
-//THIS IS JUST A TESTER FOR THE HUSKY PRE COMMIT HOOK AND PRE PUSH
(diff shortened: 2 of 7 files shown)


app/index.html
```diff
@@ -8,7 +8,8 @@
 
     <meta name="color-scheme" content="light dark" />
     <meta name="viewport" content="viewport-fit=cover, width=device-width, initial-scale=1.0, minimum-scale=1.0" />
-    <meta name=" format-detection" content="telephone=no" />
+    <meta name=" format-detection"
```
mauberti-bc: nit: revert this change

app/src/features/support/SupportPage.tsx
```diff
@@ -13,6 +13,8 @@ import {
   SupportPageViews
 } from './views/SupportPageView';
 
+//THIS IS JUST A TESTER FOR THE HUSKY PRE COMMIT HOOK AND PRE PUSH
```
mauberti-bc: Remove comments like this before marking the PR ready for review. If there is some incomplete feature or code that should be merged, it's useful to annotate the code with "TODO:" so that we can find it later. SonarQube also has some features that highlight TODOs so we don't lose visibility.

LouisThedroux: Yes, this should not have been left it and is a mistake on my part.

package.json
```diff
@@ -0,0 +1,30 @@
...
+    "api/**/*.{js,ts,tsx}": [
+      "npm run lint-fix --prefix api",
+      "npm run format-fix --prefix api"
+    ],
+    "database/**/*.{js,ts,tsx}": [
+      "npm run lint-fix --prefix database",
+      "npm run format-fix --prefix database"
+    ]
+  }
+}
```
dylanrogowsky-oxd, line 30: In the `Makefile`, I would also have `npm install` run in the root to get husky and these dependencies installed when someone runs `make install`

dylanrogowsky-oxd (changes requested):
I've tested this locally on my Mac and it does run the tests and lint the staged files I have modified.

dylanrogowsky-oxd (approved):
Retested:

- Ran `make install` and confirmed it installed Husky in the root. This should make onboarding easier. 
- Confirmed the formatting works on pre-commit after adjusting a component's indentation

mauberti-bc (approved):
👍

PR #1582, opened 2025-10-24
Publish features to BioHub

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-682: Publish Survey Metadata to BioHub](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-682)
- [SIMSBIOHUB-687: Publish Sampling Information to BioHub](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-687)
- [SIMSBIOHUB-688: Publish Observations to BioHub](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-688)
- [SIMSBIOHUB-689: Publish Animals to BioHub](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-689)
- [SIMSBIOHUB-690: Publish Telemetry to BioHub](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-690)
- [SIMSBIOHUB-691: Publish Habitat Features to BioHub](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-691)
- [SIMSBIOHUB-692: Publish Documents to BioHub](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-692)

## Description of Changes

- Publish Survey Metadata to BioHub
- Publish Sampling Information to BioHub
- Publish Observations to BioHub
- Publish Animals to BioHub
- Publish Telemetry to BioHub
- Publish Habitat Features to BioHub
- Publish Documents to BioHub

## Testing Notes

- [Publishing schema](https://apps.nrs.gov.bc.ca/int/confluence/display/TASHIS/Publishing+to+BioHub)
- Update the BACKBONE_INTAKE_PATH env variable to "/api/submission/intake"


mauberti-bc:
Overall, looks great!

#### A few things I noticed:

1. **Malformed date strings** after combining the separate date and time fields:
   Malformed Date Strings

2. **Telemetry features are not children of deployments**. Perhaps this is expected based on the seed data?
   Telemetry Features

3. **Empty strings should be turned into `null`**. We probably allow empty strings in the database, but BioHub should disallow these.
   Empty Strings

---

#### Thoughts after looking at the payload:

1. **We need a convention for code values**, related to [Let Contributor Systems Publish Codeset](https://github.com/bcgov/biohubbc-platform/pull/270). 

   - Example: SIMS has an environmental variable called "Snow Cover". A submission might have `Snow Cover = 28%`. However, a different system might also have a "Snow Cover" variable. Both concepts have the same name but different definitions and interpretations. BioHub needs the code ID to know that System A's snow cover represents `% ground cover` while System B's snow cover is actually `snow depth`.
   
2. **Distinguishing different definitions for "Snow Cover" within a single system**:
   - Example: SIMS' snow cover changed from `% cover of a 1 hectare square` to `% cover of a 10km circle` (a different code ID, but the same name).

   Snow Cover Definitions

3. **All features should have an optional external UID field** to identify the original resource.

api/src/models/biohub-create.ts
```diff
@@ -64,10 +107,503 @@ export class PostSurveyObservationToBiohubObject implements BioHubSubmissionFeat
...
+
+  constructor(captureRecord: ICritterDetailed['captures'][0]) {
+    defaultLog.debug({ label: 'PostSurveyCaptureToBiohubObject', message: 'params', captureRecord });
+
+    // Combine date and time into a single timestamp
+    const timestamp = captureRecord.capture_time
+      ? `${captureRecord.capture_date}T${captureRecord.capture_time}`
+      : captureRecord.capture_date;
+
+    this.id = captureRecord.capture_id;
```
mauberti-bc: Is this ID intended as a temporary ID to distinguish features in a submission, or to encode the original resource's ID? 

Once we flatten the payload, I think each feature will need at least 2 IDs: 
- An external UID, which would the PK of the record in SIMS or some other identifier from the original system
- A temporary ID that is unique within the submission, for letting features reference one another.

The current code works but we should decide how to set the singular ID field that currently exists for this PR.

Note that this feature will get a stable URN in BioHub (https://github.com/bcgov/biohubbc-platform/pull/261), so what we're missing is a way to find the original SIMS record. Using just the integer PK seems brittle, but the alternative of doing "<table_name>:<primary_key>" or similar seems like it exposes too much information?

Here's how ID shows up in the current payload (there are some features that don't prefix the feature type name like in the capture code above): 
image

api/src/models/biohub-create.ts
```diff
@@ -1,10 +1,37 @@
...
+  ICritterDetailed
+} from '../services/critterbase-service';
+import { ITaxonomyWithEcologicalUnits } from '../services/platform-service';
+import { SampleTechniqueRecord } from '../services/sample-technique-service';
 import { getLogger } from '../utils/logger';
 import { GetSurveyData, GetSurveyPurposeAndMethodologyData } from './survey-view';
 
+// Extended interface for sampling sites with geometry data
+interface SampleSiteRecordWithGeojson extends SampleSiteRecordExtendedNonSpatial {
+  geojson: any;
```
mauberti-bc: We can avoid ```any``` by using Feature: ```import { Feature } from 'geojson';```. You can find examples by searching "geojson: Feature"

api/src/models/biohub-create.ts
```diff
@@ -64,7 +108,493 @@ export class PostSurveyObservationToBiohubObject implements BioHubSubmissionFeat
...
+    const timestamp = captureRecord.capture_time
+      ? `${captureRecord.capture_date}T${captureRecord.capture_time}`
+      : captureRecord.capture_date;
+
+    this.id = crypto.randomUUID();
+    this.type = BiohubFeatureType.CAPTURE;
+    this.properties = {
+      ...(captureRecord.capture_comment && { comment: captureRecord.capture_comment }),
+      timestamp: timestamp,
+      geometry: captureRecord.capture_location
```
mauberti-bc, line 397: Geometry should be null if capture_location doesn't exist, to avoid misleading biohub data

api/src/models/biohub-create.ts
```diff
@@ -64,7 +108,493 @@ export class PostSurveyObservationToBiohubObject implements BioHubSubmissionFeat
...
+ * @class PostSurveyMeasurementToBiohubObject
+ * @implements {BioHubSubmissionFeature}
+ */
+export class PostSurveyMeasurementToBiohubObject implements BioHubSubmissionFeature {
+  id: string;
+  type: string;
+  properties: Record<string, any>;
+  child_features: BioHubSubmissionFeature[];
+
+  constructor(measurementRecord: any) {
```
mauberti-bc: We've been pretty strict about avoiding ```any``` types - can you make this the union of possible types instead of any?

api/src/services/sample-technique-service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export interface SampleTechniqueRecord {
+  method_technique_id: number;
+  method_name: string;
+  description: string | null;
+  method_lookup_name: string;
+  attractants: string | null;
+  distance_threshold: number | null;
+  response_metric: string;
+  attrib_data: Array<{ ah: string; av: string }>;
```
mauberti-bc: nit: use descriptive properties instead of acronyms/abbreviations. What is ah and av? 

Assuming it can be: ```attribute_data: Array<{ header: string, value: string }>```?

api/src/models/test-ecological-units.js
mauberti-bc, line 1: Should this be in the root of the api folder?

dylanrogowsky-oxd: Moved

api/src/services/sample-technique-service.ts
dylanrogowsky-oxd: It was based on existing nomenclature: eg [export-utils.test.ts#L111](https://github.com/bcgov/biohubbc/blob/d2a86ef8f8e46346caf5dcdea147b45f51c76077/api/src/services/export-services/export-utils.test.ts#L111)

database/src/seeds/04_telemetry.ts
```diff
@@ -336,7 +336,7 @@ const getAtsTelemetry = (deviceSerial: number, numRecords: number) => {
       longitude: longitude,
       cepradius_km: faker.number.int({ min: 0, max: 10 }),
       temperature: String(faker.number.float({ min: -20, max: 40 })), // TODO: Invesitgate why temperature is a string?
-      hdop: faker.string.alpha({ length: { min: 0, max: 10 } }),
+      hdop: faker.number.float({ min: 0, max: 10 }),
```
dylanrogowsky-oxd, line 339: This fixes incorrect seed data

api/src/services/platform-service.ts
```diff
@@ -429,7 +568,7 @@ export class PlatformService extends DBService {
 
     const backboneArtifactIntakeUrl = new URL(getBackboneArtifactIntakePath(), getBackboneInternalApiHost()).href;
 
-    const { data } = await axios.post<{ artifact_uuid: string }>(backboneArtifactIntakeUrl, formData.getBuffer(), {
+    const { data } = await axios.post<{ artifact_uuid: string }>(backboneArtifactIntakeUrl, formData, {
```
dylanrogowsky-oxd, line 571: This seemed to be required for documents/files to work correctly

mauberti-bc (approved):
Testing against the counterpart BioHub PR (https://github.com/bcgov/biohubbc-platform/pull/280) and the submission works as expected 👍

PR #1598, opened 2026-02-04
SIMSBIOHUB-752: Include Sampling Data in Habitat Feature Table

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-752

## Description of Changes

- Adds sampling period data to the habitat feature API responses (list and single feature endpoints).
- Displays start and end date of the sampling period for each habitat feature observation in the Habitat Feature table (Sample Period column now shows a formatted date range when available, with fallback to raw datetime).
- Reuses existing sampling information cache and `getDateTimeLabel` so the table matches the format used elsewhere (e.g. observations).

## Testing Notes

- **Where:** Survey → Sampling → Habitat Features table (or the habitat features view that uses the same table).
- **Pre-reqs:** A survey with at least one habitat feature that has a sampling period (start/end date and optionally time).
- **Verify:**
  - Load the habitat feature table and confirm the **Sample Period** column shows a readable date range (e.g. start–end) for rows with a sampling period.
  - Rows without a sampling period or with null dates still show the raw `survey_sample_period_start_datetime` when present, or an empty cell.
  - Single habitat feature fetch (e.g. edit flow) returns `supplementaryData.sampling_periods` and the UI behaves correctly when opening that context.

### DIFF
--- a/api/src/openapi/schemas/survey-habitat-feature.ts
+++ b/api/src/openapi/schemas/survey-habitat-feature.ts
@@ -286,12 +286,108 @@ export const SurveyHabitatFeaturesWithTaxonsSchema: OpenAPIV3.SchemaObject = {
 export const SurveyHabitatFeaturesSupplementaryDataSchema: OpenAPIV3.SchemaObject = {
   type: 'object',
   additionalProperties: false,
-  required: ['count', 'habitatFeatureQuantitativeDefinitions', 'habitatFeatureQualitativeDefinitions'],
+  required: ['count', 'habitatFeatureQuantitativeDefinitions', 'habitatFeatureQualitativeDefinitions', 'sampling_data'],
   properties: {
     count: {
       description: 'The total count of survey habitat features for the survey.',
       type: 'integer',
       minimum: 0
+    },
+    sampling_data: {
+      type: 'array',
+      items: {
+        type: 'object',
+        required: [
+          'survey_sample_period_id',
+          'survey_id',
+          'survey_sample_site_id',
+          'method_technique_id',
+          'start_date',
+          'start_time',
+          'end_date',
+          'end_time',
+          'method_technique',
+          'survey_sample_site'
+        ],
+        additionalProperties: false,
+        properties: {
+          survey_sample_period_id: {
+            type: 'integer',
+            minimum: 1
+          },
+          survey_id: {
+            type: 'integer',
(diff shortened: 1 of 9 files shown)


dylanrogowsky-oxd:
## Review: Prioritized issues

### High (should fix before or soon after merge)

1. **Unstable `useSamplingInformationCache` reference** — The hook returns a new object every render. In `habitatFeatureTableContext.tsx`, both the `useEffect` (that calls `initCachedSamplingInformationRef`) and the columns `useMemo` depend on `samplingInformationCache`, so the effect runs and columns recompute on every render when `sampling_periods` has length. **Fix:** Memoize the hook’s return value in `useSamplingInformationCache` (e.g. `useMemo` with stable deps) so the reference is stable. Same pattern exists in the observations table; fixing it in the hook helps both.

### Medium (recommended)

2. **Single-feature endpoint not fully parallel** — In `survey-habitat-feature-service.ts`, `getSurveyHabitatFeature` does a `Promise.all` for feature + definitions, then awaits `getSamplePeriodsForSurvey(surveyId)` separately, adding an extra round trip. **Fix:** Add `this.samplePeriodService.getSamplePeriodsForSurvey(surveyId)` to the existing `Promise.all` so all three requests run in parallel.

3. **API naming consistency** — In `SurveyHabitatFeaturesSupplementaryDataSchema`, the new key is `sampling_periods` (snake_case) while the other keys are camelCase (`habitatFeatureQuantitativeDefinitions`, `habitatFeatureQualitativeDefinitions`). **Fix:** If the API convention for this response is camelCase, rename to `samplingPeriods` and update app + tests; otherwise leave as-is and document the convention.

### Low (nice to have)

4. **Effect dependency on `sampling_periods`** — The `useEffect` depends on `habitatFeatureDataLoader.data?.supplementaryData.sampling_periods`. If the loader reuses the same array reference on refetch, the effect might not re-run when the list of periods changes. **Suggestion:** Confirm that refetched data gets a new array reference so cache re-init runs when sampling periods change.

---
*Summary: Functionality and types look correct; tests are updated. The high-priority item is stabilizing the cache hook to avoid unnecessary effect and column recalculations.*

dylanrogowsky-oxd (approved):

PR #1602, opened 2026-02-06
SIMSBIOHUB-867: Analytics Tab Infinite Load Error

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-867 – BugFix: Analytics Tab Infinite Load Error](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-867)

## Description of Changes

- **Analytics tab:** On the survey details Analytics tab, the observations analytics API was being called many times per second and only stopped when switching back to the Counts tab. The effect that triggers the fetch had `analyticsDataLoader` in its dependency array. The loader object from `useDataLoader` gets a new reference whenever its state updates (e.g. after each request), so the effect kept re-running and calling `refresh()` in a loop. The fix is to hold the loader’s `refresh` in a ref and have the effect depend only on the real inputs (`surveyId` and the group-by options).
- **Same pattern elsewhere:** The same “loader in effect deps” pattern existed in a bunch of other places (codes context, survey spatial components, devices/deployments containers, create survey page, etc.). Those are updated to use the same ref approach so we don’t get similar request loops or unnecessary refetches.

## Testing Notes

- **Analytics tab (primary):** Go to survey details (e.g. `/admin/projects/1/surveys/1/details`), open the **Analytics** tab and leave it open. In the Network tab you should see a single request to `/api/analytics/observations` (or one per change to group-by options), not a continuous stream of requests.
- **Other areas to spot-check** (same ref-based fix applied; confirm data still loads and no request storms):
  - **Survey spatial:** Observations, Animals, Telemetry, and Habitat Features map/table views under survey details.
  - **Survey sampling:** Techniques list/cards, Sampling Sites map, Sampling Periods table.
  - **Telemetry:** Devices and Deployments tables under a survey (manage devices/deployments).
  - **Survey create/edit:** Create Survey page (codes + project data), Funding section, and any forms that use codes (e.g. habitat feature type dropdowns).
  - **Summary / list filters:** Habitat Features list filter form (codes for habitat feature types).
  - **Animals:** Mortality details (cause-of-death codes), ecological units card (species standards).
  - **Standards:** Species standards ecological unit card.

### DIFF
--- a/app/src/contexts/codesContext.tsx
+++ b/app/src/contexts/codesContext.tsx
@@ -1,7 +1,7 @@
 import { useBiohubApi } from 'hooks/useBioHubApi';
 import useDataLoader, { DataLoader } from 'hooks/useDataLoader';
 import { IGetAllCodeSetsResponse } from 'interfaces/useCodesApi.interface';
-import { createContext, PropsWithChildren, useEffect, useMemo } from 'react';
+import { createContext, PropsWithChildren, useEffect, useMemo, useRef } from 'react';
 
 /**
  * Context object that stores information about codes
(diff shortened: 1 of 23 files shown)


mauberti-bc (approved):
Looks good

PR #1608, opened 2026-02-11
SIMSBIOHUB-870: Align payload property names with BioHub

dylanrogowsky-oxd:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-870

## Description of Changes

- **observation_environmental_condition (quantitative):** Use `environmental_condition_value` for quantitative conditions instead of `environmental_quantitative_value`, so both qualitative and quantitative use the same property name and match what the companion system expects.
- **telemetry_device:** Renamed payload property from `model` to `device_model` to match the companion audit.
- **report (report attachments):** Send `artifact_key` (from report attachment key), `name` (from title), and `year` (from year_published) instead of `artifact_id`, `title`, and `year_published` so the payload matches the companion audit.
- Updated the tests that were asserting on the old property names so the suite passes.

## Testing Notes

- `npm test` in the api package — all tests pass. The only changes in this PR are the payload property names above and the corresponding test expectations; no behaviour change beyond what we send to the API.


mauberti-bc (approved):
👍

PR #1609, opened 2026-02-24
SIMSBIOHUB-850: Implement Crunchy DB

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-850](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-850)

## Description of Changes

- **Crunchy Postgres for DEV/TEST/PROD**: New Helm chart `infrastructure/crunchy-db` deploys Crunchy Postgres (PGO) as a separate release. Static environments (dev/test/prod) disable the in-repo database subchart and use Crunchy; PR previews keep the legacy single-pod Postgres.
- **One-time migration job**: Optional migration Job (pg_dump from legacy DB → pg_restore into Crunchy) with manual app-user secret (`<release>-pguser-biohub-api`). Post-restore step sets `search_path` and ACLs for `biohub_api` so the API can resolve and run biohub schema functions.
- **API and database-setup**: Default `useCrunchy: true`; connection details (host, secret names) come from values. Secret keys (`user`/`password`/`dbname`) default in templates so env values only set host and secret names. PRs explicitly set `useCrunchy: false` and use the legacy `dbHost` secret.
- **Simplified values**: Removed redundant `migrationHost` when same as `host` (template defaults); removed duplicate `adminUserKey`/`appUserKey`/etc. from dev/test/prod.
- **CI**: Removed `buildAndPushDatabase` job and its dependency from the deploy job for static envs; Chart description and `biohubbc-db-setup` dependency updated.

## Testing Notes

- **PR previews**: Ensure `useCrunchy: false` for api and db-setup so PRs still use the legacy database secret and keys.
- **DEV/TEST/PROD (after Crunchy is deployed)**: Create the app user secret before running migration (see `docs/LOCAL-DEV-TEST.md`). Run migration once with `migration.enabled=true`, then set it false. Verify API and db-setup use Crunchy primary and `pguser-biohub-api` secret.

### DIFF
--- a/database/src/knexfile.ts
+++ b/database/src/knexfile.ts
@@ -1,9 +1,7 @@
-import * as fs from 'fs';
+import * as fs from 'node:fs';
 
 // Use SSL with CA verification when PG_SSL_CA_PATH is set (Crunchy); otherwise no SSL (legacy/PR).
-const sslConfig = process.env.PG_SSL_CA_PATH
-  ? { ca: fs.readFileSync(process.env.PG_SSL_CA_PATH) }
-  : false;
+const sslConfig = process.env.PG_SSL_CA_PATH ? { ca: fs.readFileSync(process.env.PG_SSL_CA_PATH) } : false;
 
 export default {
   development: {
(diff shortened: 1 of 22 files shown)


database/src/knexfile.ts
```diff
@@ -1,3 +1,6 @@
+// When PGSSLMODE=require (e.g. Crunchy Postgres with self-signed cert), use SSL and accept self-signed for this connection only.
+const sslConfig = process.env.PGSSLMODE === 'require' ? { rejectUnauthorized: false } : false;
```
mauberti-bc: Can we have tigher security here? I'm still wrapping my head around this but the node API must receive the crunchy CA to allow `{ rejectUnauthorized: true }`

PR #1613, opened 2026-03-05
SIMSBIOHUB-876: Show Publish History

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-875: CRUD Operations on Submissions](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-875)
- [SIMSBIOHUB-876: Show Publish History in SIMS](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-876)

## Description of Changes
- **Publish flow:** First-time publish uses `{BACKBONE_SUBMISSION_UPLOAD_PATH}/upload/archive`; re-publish (when `survey_metadata_publish` exists) uses `{path}/{submission_uuid}/upload`. One env var only: `BACKBONE_SUBMISSION_UPLOAD_PATH` (e.g. `/api/submission`). Stored `submission_uuid` is the BioHub submission resource UUID, not the upload session id.

- **Submission history API:** New proxy endpoints: `GET /api/submission/:submissionId/history` and `DELETE /api/submission/:submissionId/upload/:submissionUploadId`. Auth via resolve-to-survey and same permissions as publish. History returns rows with `submissionUploadId`, `status`, `createDate`, and optional `id` (for View in BioHub). Backend normalizes BioHub’s array or `{ submissionId, history }` shape.

- **Shared middleware:** Resolve and auth for submission routes moved to `api/src/paths/submission/resolveSubmissionToSurvey.ts`; history and delete paths import from it.

- **UI:** Removed Status line from survey header and publish dialog. Added History button (outlined + icon) next to Publish, visible only when published; opens modal with table (Date, Status, actions). Submitted rows: ellipsis → Delete (confirm; row shows “deleted”, modal stays open). Approved rows: ellipsis → View in BioHub (uses row `id` and config base URL). Modal has OK to close.

- **Tests:** PlatformService (getSubmissionHistory, deleteSubmissionUpload), history-publish repo/service (getSurveyMetadataPublishRecordBySubmissionUuid), and GET/DELETE submission path handlers.

## Testing Notes

- First-time publish: completes and stores a valid UUID in `survey_metadata_publish.submission_uuid`; initiate URL uses `/upload/archive`.
- Re-publish: initiate goes to `/{submission_uuid}/upload`; same `submission_uuid` retained.
- History modal: loads table; View in BioHub opens correct URL; Delete updates row to “deleted” and keeps modal open; OK closes.
- Without publish permission, History is not available and history/delete endpoints are blocked.
- Set `BACKBONE_SUBMISSION_UPLOAD_PATH` (e.g. `/api/submission`).


api/src/paths/submission/resolveSubmissionToSurvey.ts
```diff
@@ -0,0 +1,57 @@
+import { Request, RequestHandler } from 'express';
+import { PROJECT_PERMISSION, SYSTEM_ROLE } from '../../constants/roles';
+import { getDBConnection } from '../../database/db';
+import { authorizeRequestHandler } from '../../request-handlers/security/authorization';
+import { HistoryPublishService } from '../../services/history-publish-service';
+import { getLogger } from '../../utils/logger';
+
+const defaultLog = getLogger('paths/submission/resolveSubmissionToSurvey');
```
mauberti-bc: This pattern feels out of sync with the codebase.
* Only include `index.ts` files in paths
* Request-modifying middleware (eg. inject property into the request) should be in `middleware/`, but in this case, we can just have a repository method for retrieving the submission ID

api/src/paths/submission/resolveSubmissionToSurvey.ts
```diff
@@ -0,0 +1,57 @@
...
+  } finally {
+    connection.release();
+  }
+};
+
+/**
+ * Returns the auth config for submission routes (coordinator/collaborator or system admin).
+ * Use with authorizeRequestHandler(getSubmissionAuthRequestHandler).
+ */
+export function getSubmissionAuthRequestHandler(req: Request) {
```
mauberti-bc: Just inline this inside the endpoint's `index.ts` file, like how the rest of the endpoints add middleware. The pattern of adding non-index helpers to `paths/` feels like a divergence.

api/src/repositories/history-publish-repository.ts
```diff
@@ -229,6 +229,29 @@ export class HistoryPublishRepository extends BaseRepository {
...
+   * Gets a record from `survey_metadata_publish` for a given submission_uuid.
+   *
+   * @param {string} submissionUuid
+   * @return {*}  {Promise<(SurveyMetadataPublish | null)>}
+   * @memberof HistoryPublishRepository
+   */
+  async getSurveyMetadataPublishRecordBySubmissionUuid(submissionUuid: string): Promise<SurveyMetadataPublish | null> {
+    const sqlStatement = SQL`
+      select
+        *
```
mauberti-bc: Avoid `select *`. There should be a zod schema (named SurveyMetadataPublish) for the expected output. We should select only the fields to match the schema, then pass the schema to `this.connection.sql(sql, <zod schema>)` for validation.

api/src/repositories/history-publish-repository.ts
```diff
@@ -229,6 +229,29 @@ export class HistoryPublishRepository extends BaseRepository {
...
   }
 
+  /**
+   * Gets a record from `survey_metadata_publish` for a given submission_uuid.
+   *
+   * @param {string} submissionUuid
+   * @return {*}  {Promise<(SurveyMetadataPublish | null)>}
+   * @memberof HistoryPublishRepository
+   */
+  async getSurveyMetadataPublishRecordBySubmissionUuid(submissionUuid: string): Promise<SurveyMetadataPublish | null> {
```
mauberti-bc: If a repository GET method is allowed to return null, it should be named `find*` instead of `get*`. If we don't expect null, we should throw an error if the rowCount is < 1.

In this case, the record won't exist if the survey has not been published, so just rename this to `findSurveyMetadataPublishRecordBySubmissionUuid`

api/src/repositories/history-publish-repository.ts
```diff
@@ -229,6 +229,29 @@ export class HistoryPublishRepository extends BaseRepository {
...
+      from
+        survey_metadata_publish
+      where
+        submission_uuid = ${submissionUuid}
+      limit 1;
+    `;
+
+    const response = await this.connection.sql<SurveyMetadataPublish>(sqlStatement);
+
+    return (response.rows.length && response.rows[0]) || null;
```
mauberti-bc: Simplify to just `response.rows[0] ?? null`?

api/src/repositories/history-publish-repository.ts
```diff
@@ -229,6 +229,29 @@ export class HistoryPublishRepository extends BaseRepository {
...
+      select
+        *
+      from
+        survey_metadata_publish
+      where
+        submission_uuid = ${submissionUuid}
+      limit 1;
+    `;
+
+    const response = await this.connection.sql<SurveyMetadataPublish>(sqlStatement);
```
mauberti-bc: We should avoid typescript hints like `<SurveyMetadataPublish>`. Instead, pass the zod schema.

api/src/services/platform-service.ts
```diff
@@ -251,18 +255,21 @@ export class PlatformService extends DBService {
...
+        data.submissionComment,
+        existingSubmissionUuid
+      );
 
     defaultLog.info({
       label: 'submitSurveyToBioHub',
       message: 'Initiated multipart upload',
       uploadId,
       submissionId,
+      submissionUuid: submissionUuid ?? '(none)',
```
mauberti-bc: I don't think we want `'(none)', and submissionUuid should always be defined

api/src/services/platform-service.ts
```diff
@@ -298,13 +305,14 @@ export class PlatformService extends DBService {
       }
     }
 
-    // Insert publish history record
+    // Insert or update publish history record (re-publish uses existing submission_uuid).
+    const submissionUuidToStore = existingSubmissionUuid ?? submissionUuid ?? String(submissionId);
```
mauberti-bc: We shouldn't never fallback to the submission integer ID if we're expecting the uuid. Also, existingSubmissionUuid should equal submissionUuid. Since BioHub always returns the submissionUuid from `this._initiateSubmissionUpload`, we should use only that value:
```
const submissionUuidToStore = submissionUuid
```

in which case just remove `submissionUuidToStore` entirely and pass the submissionUuid downstream

api/src/services/platform-service.ts
```diff
@@ -828,28 +836,36 @@ export class PlatformService extends DBService {
...
+    submissionComment: string,
+    existingSubmissionUuid: string | null = null
   ): Promise<{
     uploadId: string;
     s3UploadId: string;
     key: string;
     presignedUrls: Array<{ partNumber: number; url: string }>;
     partCount: number;
     submissionId: number;
+    submissionUuid?: string;
```
mauberti-bc: submissionUuid should always exist if submissionId exists, so why potentially undefined?

api/src/services/platform-service.ts
```diff
@@ -828,28 +836,36 @@ export class PlatformService extends DBService {
...
+   * @return {*}  {Promise<{uploadId: string, s3UploadId: string, key: string, presignedUrls: Array<{partNumber: number, url: string}>, partCount: number, submissionId: number, submissionUuid?: string}>}
    * @memberof PlatformService
    */
   async _initiateSubmissionUpload(
     token: string,
     tarFileSize: number,
     surveyDataPackage: PostSurveySubmissionToBioHubObject,
-    submissionComment: string
+    submissionComment: string,
+    existingSubmissionUuid: string | null = null
```
mauberti-bc: Don't set a default if it's required?

api/src/services/platform-service.ts
```diff
@@ -1105,4 +1140,87 @@ export class PlatformService extends DBService {
...
   }
+
+  /**
+   * Get submission upload history from BioHub.
+   *
+   * @param {string} submissionId - Submission UUID
+   * @return {*}  {Promise<Array<{ submissionUploadId: string; status: string; createDate: string; id?: number }>>}
+   * @memberof PlatformService
+   */
+  async getSubmissionHistory(
```
mauberti-bc: This whole method feels messy and misaligned with codebase standards.

api/src/services/platform-service.ts
```diff
@@ -1105,4 +1140,87 @@ export class PlatformService extends DBService {
...
+            history?: Array<{ submissionUploadId: string; status: string; createDate: string; id?: number }>;
+          }
+      >(url, {
+        headers: {
+          authorization: `Bearer ${token}`,
+          'Content-Type': 'application/json'
+        }
+      });
+      const data = response.data;
+      if (Array.isArray(data)) {
```
mauberti-bc, line 1211: We should know exactly what the response shape is, and we should have an interface for it.

Create types to match responses defined in https://github.com/bcgov/biohubbc-platform/pull/351/changes

api/src/services/platform-service.ts
```diff
@@ -1105,4 +1140,87 @@ export class PlatformService extends DBService {
...
+   *
+   * @param {string} submissionId - Submission UUID
+   * @param {string} submissionUploadId - Submission upload UUID
+   * @return {*}  {Promise<void>}
+   * @memberof PlatformService
+   */
+  async deleteSubmissionUpload(submissionId: string, submissionUploadId: string): Promise<void> {
+    const keycloakService = new KeycloakService();
+    const token = await keycloakService.getKeycloakServiceToken();
+    const url = new URL(`/submission/${submissionId}/upload/${submissionUploadId}`, getBackboneInternalApiHost()).href;
```
mauberti-bc: Probably worth trying to address the `Server-side request forgery` warnings

api/src/services/platform-service.ts
```diff
@@ -1105,4 +1140,87 @@ export class PlatformService extends DBService {
...
+  async getSubmissionHistory(
+    submissionId: string
+  ): Promise<Array<{ submissionUploadId: string; status: string; createDate: string; id?: number }>> {
+    const keycloakService = new KeycloakService();
+    const token = await keycloakService.getKeycloakServiceToken();
+    const url = new URL(`/submission/${submissionId}/history`, getBackboneInternalApiHost()).href;
+
+    try {
+      const response = await axios.get<
+        | Array<{ submissionUploadId: string; status: string; createDate: string; id?: number }>
```
mauberti-bc: Avoid using `id` because it's unclear what the id represents. Is it necessary here, since we really just want `submissionUploadId`, `status`, and `date`?

mauberti-bc (changes requested):

api/src/paths/submission/resolveSubmissionToSurvey.ts
dylanrogowsky-oxd: Implemented. I removed the standalone `resolveSubmissionToSurvey.ts` helper and moved resolve/auth logic into the submission endpoint files so paths/ no longer has a non-route helper file.

api/src/paths/submission/resolveSubmissionToSurvey.ts
dylanrogowsky-oxd: Done. I inlined `resolveSubmissionToSurvey` + auth handler composition directly in:

- `api/src/paths/submission/{submissionId}/history.ts`
- `api/src/paths/submission/{submissionId}/upload/{submissionUploadId}.ts`

api/src/repositories/history-publish-repository.ts
dylanrogowsky-oxd: Addressed. I replaced `select *` with explicit columns for survey metadata publish lookups and switched query validation to zod schema usage in `connection.sql(sql, SurveyMetadataPublish)` in `api/src/repositories/history-publish-repository.ts`.

api/src/repositories/history-publish-repository.ts
dylanrogowsky-oxd: Done. Renamed `getSurveyMetadataPublishRecordBySubmissionUuid` to `findSurveyMetadataPublishRecordBySubmissionUuid`

api/src/repositories/history-publish-repository.ts
dylanrogowsky-oxd: Return logic now uses `response.rows[0] ?? null`

api/src/services/platform-service.ts
dylanrogowsky-oxd: Fixed. Removed `'(none)'` fallback logging and now log/use the actual `submissionUuid` value only.

api/src/services/platform-service.ts
dylanrogowsky-oxd: Addressed. Removed integer fallback and `submissionUuidToStore`. Persisted UUID now comes directly from `_initiateSubmissionUpload` and is passed through consistently.

api/src/services/platform-service.ts
dylanrogowsky-oxd: Implemented. Replaced ambiguous history `id` with `submissionId`

api/src/services/platform-service.ts
dylanrogowsky-oxd: Done. Added explicit history response interfaces and normalized array/wrapped response shapes in `PlatformService` according to the platform contract patterns from PR 351.

api/src/services/platform-service.ts
dylanrogowsky-oxd: Addressed. Added UUID validation for `submissionId` and `submissionUploadId` before building downstream URLs, while keeping host/protocol fixed via `getBackboneInternalApiHost()`.

api/src/services/platform-service.ts
dylanrogowsky-oxd: Updated. Removed the default value from `_initiateSubmissionUpload(..., existingSubmissionUuid)`

api/src/services/platform-service.ts
dylanrogowsky-oxd: Implemented. `_initiateSubmissionUpload` now treats `submissionUuid` as required in the returned shape

api/src/repositories/history-publish-repository.ts
dylanrogowsky-oxd: Implemented. Query now uses schema-validated form: `this.connection.sql(sqlStatement, SurveyMetadataPublish)` instead of generic-based typing

api/src/services/platform-service.ts
dylanrogowsky-oxd: Refactored `getSubmissionHistory()` to simplify/clean it up

api/src/paths/submission/{submissionId}/upload/{submissionUploadId}.ts
```diff
@@ -0,0 +1,95 @@
+import { Request, RequestHandler } from 'express';
+import { Operation } from 'express-openapi';
+import { PROJECT_PERMISSION, SYSTEM_ROLE } from '../../../../constants/roles';
+import { getDBConnection } from '../../../../database/db';
+import { authorizeRequestHandler } from '../../../../request-handlers/security/authorization';
+import { HistoryPublishService } from '../../../../services/history-publish-service';
+import { PlatformService } from '../../../../services/platform-service';
+import { getLogger } from '../../../../utils/logger';
+
+const defaultLog = getLogger('paths/submission/{submissionId}/upload/{submissionUploadId}');
```
mauberti-bc: It might make more sense to nest this endpoint under the survey. Then existing middleware for authorizing survey access applies.

project/:projectId/survey/:surveyId/submission/:submissionId

api/src/paths/submission/{submissionId}/history.ts
```diff
@@ -0,0 +1,116 @@
+import { Request, RequestHandler } from 'express';
+import { Operation } from 'express-openapi';
+import { PROJECT_PERMISSION, SYSTEM_ROLE } from '../../../constants/roles';
+import { getDBConnection } from '../../../database/db';
+import { authorizeRequestHandler } from '../../../request-handlers/security/authorization';
+import { HistoryPublishService } from '../../../services/history-publish-service';
+import { PlatformService } from '../../../services/platform-service';
+import { getLogger } from '../../../utils/logger';
+
+const defaultLog = getLogger('paths/submission/{submissionId}/history');
```
mauberti-bc: Move to paths/project/:projectId/survey/:surveyId/submission/:submissionId/history

api/src/paths/submission/{submissionId}/history.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export const resolveSubmissionToSurvey: RequestHandler = async (req, res, next) => {
```
mauberti-bc: We shouldn't need any middleware changes in this PR. The `PlatformService`'s method for getting history for a given submission ID should validate that the submission ID belongs to given Survey, then makes the request to BioHub for history and returns the result.

api/src/paths/submission/{submissionId}/history.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const connection = getDBConnection(req.keycloak_token);
+  try {
+    await connection.open();
+    const historyPublishService = new HistoryPublishService(connection);
+    const record = await historyPublishService.findSurveyMetadataPublishRecordBySubmissionUuid(submissionId);
+    if (!record) {
+      return res.status(404).json({ message: 'Submission not found' });
+    }
+    req.survey_id_for_submission = record.survey_id;
```
mauberti-bc: We shouldn't be adding anything to the `req`

api/src/paths/submission/{submissionId}/history.ts
```diff
@@ -0,0 +1,116 @@
...
+    connection.release();
+  }
+};
+
+export function getSubmissionAuthRequestHandler(req: Request) {
+  return {
+    or: [
+      {
+        validProjectPermissions: [PROJECT_PERMISSION.COORDINATOR, PROJECT_PERMISSION.COLLABORATOR],
+        surveyId: req.survey_id_for_submission!,
```
mauberti-bc: Once we move the path to `project/:projectId/survey/:surveyId/submission/:submissionId/history`, we can pass the `surveyId` from the path for validation.

api/src/services/platform-service.ts
```diff
@@ -71,6 +72,25 @@ const getBackboneTaxonTsnPath = () => getEnvironmentVariable('BIOHUB_TAXON_TSN_P
 const getBackboneTaxonPath = () => getEnvironmentVariable('BIOHUB_TAXON_PATH');
 const getBackboneSubmissionUploadPath = () => getEnvironmentVariable('BACKBONE_SUBMISSION_UPLOAD_PATH');
 const getBackboneUploadCompletePath = () => getEnvironmentVariable('BACKBONE_UPLOAD_COMPLETE_PATH');
+const uuidSchema = z.string().uuid();
+
+interface IBioHubSubmissionHistoryRow {
```
mauberti-bc: Move interface/type definitions to either `platform-service.interface.ts`

api/src/services/platform-service.ts
```diff
@@ -1105,4 +1165,100 @@ export class PlatformService extends DBService {
...
+   *
+   * @param {string} submissionId - Submission UUID
+   * @param {string} submissionUploadId - Submission upload UUID
+   * @return {*}  {Promise<void>}
+   * @memberof PlatformService
+   */
+  async deleteSubmissionUpload(submissionId: string, submissionUploadId: string): Promise<void> {
+    const keycloakService = new KeycloakService();
+    const token = await keycloakService.getKeycloakServiceToken();
+    const validatedSubmissionId = this._requireUuid(submissionId, 'submissionId');
```
mauberti-bc: We don't need a `_requireUuid` method because the openapi schemas will validate it anyway. The SIMS endpoint will require that `submissionId` in the path is a UUID, and BioHub's endpoint will do the same.

app/src/components/publish/SurveyPublishHistoryDialog.tsx
```diff
@@ -0,0 +1,257 @@
...
+interface ISurveyPublishHistoryDialogProps {
+  open: boolean;
+  onClose: () => void;
+  submissionId: string | undefined;
+}
+
+/**
+ * Dialog showing publish history for a survey: table with Date, Status, and delete action for submitted uploads.
+ */
+function renderDeleteErrorSnackbar(message: string) {
```
mauberti-bc: nit: Move this into the React functional component. Each .tsx file should just define and export a single React FC.

app/src/components/publish/SurveyPublishHistoryDialog.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+  const loadHistory = useCallback(async () => {
+    if (!submissionId) {
+      setHistory([]);
+      return;
+    }
+    setLoading(true);
+    setError(null);
+    try {
+      const data = await biohubApi.publish.getSubmissionHistory(submissionId);
```
mauberti-bc: Follow the `useDataLoader` pattern that we use for fetching data.
- Define the data loader
- Call `dataLoader.load()` in a useEffect
- Access the data via `dataLoader.data`

app/src/components/publish/SurveyPublishHistoryDialog.tsx
```diff
@@ -0,0 +1,257 @@
...
+            </Typography>
+          )}
+          {submissionId && !loading && !error && history.length === 0 && (
+            <Typography variant="body2" color="textSecondary">
+              No history yet.
+            </Typography>
+          )}
+          {submissionId && !loading && !error && history.length > 0 && (
+            <TableContainer>
+              <Table size="small">
```
mauberti-bc: Never use MUI's Table, always use the `StyledDataGrid`

app/src/components/publish/SurveyPublishHistoryDialog.tsx
```diff
@@ -0,0 +1,257 @@
...
+    }
+    try {
+      await biohubApi.publish.deleteSubmissionUpload(submissionId, toDelete.submissionUploadId);
+      closeDeleteDialog();
+      setHistory((prev) =>
+        prev.map((r) => (r.submissionUploadId === toDelete.submissionUploadId ? { ...r, status: 'deleted' } : r))
+      );
+      dialogContext.setSnackbar({ snackbarMessage: 'Upload request cancelled.', open: true });
+    } catch (err: unknown) {
+      const message = err instanceof Error ? err.message : 'Failed to cancel upload request';
```
mauberti-bc: Just `const message = (error as APIError).message`

app/src/components/publish/SurveyPublishHistoryDialog.tsx
```diff
@@ -0,0 +1,257 @@
...
+        fullWidth>
+        <DialogTitle id="survey-publish-history-dialog-title">Publish History</DialogTitle>
+        <DialogContent>
+          {!submissionId && (
+            <Typography variant="body2" color="textSecondary">
+              No submission found for this survey.
+            </Typography>
+          )}
+          {submissionId && loading && (
+            <Typography variant="body2" color="textSecondary">
```
mauberti-bc: Use `LoadingGuard` with the `isLoading` and `isLoadingFallback` props when conditionally showing a loading state.

app/src/components/publish/SurveyPublishHistoryDialog.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+      <Menu
+        open={Boolean(actionMenuAnchor)}
+        onClose={() => setActionMenuAnchor(null)}
+        anchorEl={actionMenuAnchor?.anchor}
+        anchorOrigin={{ vertical: 'top', horizontal: 'right' }}
+        transformOrigin={{ vertical: 'top', horizontal: 'right' }}>
+        {actionMenuAnchor?.row.status === 'approved' && (
+          <MenuItem
+            onClick={() => {
```
mauberti-bc: Create a function called `openBioHubSubmission` for clarity.

mauberti-bc (changes requested):

api/src/paths/submission/{submissionId}/history.ts
dylanrogowsky-oxd: Moved to `project/{projectId}/survey/{surveyId}/submission/{submissionId}/history.ts`

api/src/paths/submission/{submissionId}/history.ts
dylanrogowsky-oxd: Removed and now uses `PlatformService`

api/src/paths/submission/{submissionId}/history.ts
dylanrogowsky-oxd: Removed

api/src/paths/submission/{submissionId}/history.ts
dylanrogowsky-oxd: Updated: `const surveyId = Number(req.params.surveyId);`

app/src/interfaces/usePublishApi.interface.ts
```diff
@@ -4,3 +4,10 @@ export interface IProjectSubmitForm {
   reports: IGetProjectReportAttachment[];
   attachments: IGetProjectAttachment[];
 }
+
+export interface ISubmissionHistoryRow {
+  submissionUploadId: string;
+  status: string;
+  createDate: string;
+  submissionId?: number;
```
mauberti-bc: `submissionId` should be required?

mauberti-bc (changes requested):

app/src/interfaces/usePublishApi.interface.ts
dylanrogowsky-oxd: Updated

mauberti-bc (approved):
👍

PR #1627, opened 2026-03-25
SIMSBIOHUB-934: Update publishing to use part-specific size for each multipart url

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-934

## Description of Changes

- Uses the `partSizeBytes` per presigned URL, instead of the top-level `partSizeBytes` provided by BioHub
- https://github.com/bcgov/biohubbc-platform/pull/393: BioHub finds an optimal part size depending on size of tarball, ranging from 5Mb to whatever is required to stay below 10,000 parts.
- Updates publishing flow to stream the tarball

## Testing Notes

- Publishing should behave like before

### DIFF
--- a/api/src/services/platform-service.interface.ts
+++ b/api/src/services/platform-service.interface.ts
@@ -16,6 +16,35 @@ export interface ISubmissionHistoryRow {
   submissionId?: number;
 }
 
+export interface UploadPart {
+  partNumber: number;
+  url: string;
+  partSizeBytes: number;
+}
+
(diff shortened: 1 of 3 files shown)


mauberti-bc:
Note: the code smell about changing `sort` to `toSorted` requires a newer TypeScript version

dylanrogowsky-oxd (approved):

PR #1630, opened 2026-04-29
SIMSBIOHUB-970: Publish Part of a Survey

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-970](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-970)

## Description of Changes

- Added support for publishing selected parts of a survey instead of always publishing everything.
- Added an endpoint to return publishable feature types for a survey, and wired the publish dialog to load and submit selected feature types.
- Enforced feature dependency rules (parent/child) in UI and backend normalization so payloads are valid and consistent.
- Updated BioHub submission packaging flow to include only selected feature types, and added coverage for feature filtering and multipart upload behavior.

## Testing Notes

- In the survey view, open the Publish dialog and verify only available feature types are shown for the current survey.
- Verify feature dependency behavior in the dialog:
  - selecting a child auto-selects required parent(s)
  - unselecting a parent removes dependent children
- Publish with different feature subsets (for example: only observations, only sampling-related data, telemetry-related data) and confirm submission succeeds.
- Confirm publish history/submission behavior still works for republish and no regressions in survey header actions.
- Spot-check that attachment publishing includes only intended publishable attachment types.

### DIFF
--- a/api/src/paths/project/{projectId}/survey/{surveyId}/publish/features.ts
+++ b/api/src/paths/project/{projectId}/survey/{surveyId}/publish/features.ts
@@ -2,8 +2,10 @@ import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
 import { PROJECT_PERMISSION, SYSTEM_ROLE } from '../../../../../../constants/roles';
 import { getDBConnection } from '../../../../../../database/db';
+import { HTTP400 } from '../../../../../../errors/http-error';
 import { authorizeRequestHandler } from '../../../../../../request-handlers/security/authorization';
 import { PlatformService } from '../../../../../../services/platform-service';
+import { SurveyService } from '../../../../../../services/survey-service';
 import { getLogger } from '../../../../../../utils/logger';
 
 const defaultLog = getLogger('paths/project/{projectId}/survey/{surveyId}/publish/features');
@@ -65,10 +67,18 @@ GET.apiDoc = {
 export function getSurveyPublishableFeatures(): RequestHandler {
   return async (req, res) => {
     const connection = getDBConnection(req.keycloak_token);
+    const projectId = Number(req.params.projectId);
     const surveyId = Number(req.params.surveyId);
 
     try {
       await connection.open();
+
+      const surveyService = new SurveyService(connection);
+      const surveyData = await surveyService.getSurveyData(surveyId);
+
+      if (surveyData.project_id !== projectId) {
+        throw new HTTP400('Invalid project or survey identifier.');
+      }
 
       const platformService = new PlatformService(connection);
       const response = await platformService.getSurveyPublishableFeatures(surveyId);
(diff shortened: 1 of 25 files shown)


api/src/services/platform-service.ts
```diff
@@ -99,6 +109,246 @@ export class PlatformService extends DBService {
...
+  /**
+   * Get publishable survey feature types that exist for the survey.
+   *
+   * If a child feature exists, its parent feature type is automatically included.
+   *
+   * @param {number} surveyId
+   * @return {Promise<{ featureTypes: BiohubFeatureType[] }>}
+   * @memberof PlatformService
+   */
+  async getSurveyPublishableFeatures(surveyId: number): Promise<{ featureTypes: BiohubFeatureType[] }> {
```
mauberti-bc, line 151: Telemetry deployments lose animal linkage when only telemetry is selected
Partial publishing can include telemetry deployment without animal features. But _generateSurveyDataPackage only loads surveyAnimals when ANIMAL is selected, and PostTelemetryDeploymentToBiohubObject uses animalRecords to populate animal_identifier. So publishing telemetry without animals will have deployments with animal_identifier: null.

api/src/paths/project/{projectId}/survey/{surveyId}/publish/features.ts
```diff
@@ -0,0 +1,84 @@
...
+        }
+      }
+    },
+    401: { $ref: '#/components/responses/401' },
+    403: { $ref: '#/components/responses/403' },
+    500: { $ref: '#/components/responses/500' }
+  }
+};
+
+export function getSurveyPublishableFeatures(): RequestHandler {
```
mauberti-bc, line 67: Since this endpoint now projectId for authorization and surveyId for the publish target, ensure that we validate that the survey belongs to the project. Otherwise a request can authorize against one project and publish a survey from another.

mauberti-bc (changes requested):

api/src/services/platform-service.ts
dylanrogowsky-oxd: - Updated `_generateSurveyDataPackage` in `api/src/services/platform-service.ts` to load survey animals when either `ANIMAL` or `TELEMETRY_DEPLOYMENT` is selected.
- This ensures telemetry deployment features still have `animalRecords` available to resolve `animal_identifier`, even when `ANIMAL` is not part of the selected publish subset.
- Added a test in `api/src/services/platform-service.test.ts` that verifies telemetry-deployment-only selection still triggers critter loading.

api/src/paths/project/{projectId}/survey/{surveyId}/publish/features.ts
dylanrogowsky-oxd: - Added a survey/project ownership check in `api/src/paths/project/{projectId}/survey/{surveyId}/publish/features.ts`:
  - Load survey via `SurveyService.getSurveyData(surveyId)`.
  - Return `HTTP400` when `survey.project_id` does not match route `projectId`.
- Added the same ownership check in `api/src/paths/publish/survey.ts`:
  - Validate `survey.project_id` against request body `projectId` before calling `submitSurveyToBioHub`.
  - Return `HTTP400` on mismatch.
- Added endpoint tests for both mismatch scenarios:
  - `api/src/paths/project/{projectId}/survey/{surveyId}/publish/features.test.ts`
  - `api/src/paths/publish/survey.test.ts`

mauberti-bc (approved):
👍

